PR #7, opened 2023-04-24
[#PAB-207] enum validation

cyrusdobbs:
https://trello.com/c/ZCEVybXJ/207-implement-enum-validation-in-spreadsheet-validator

### Change description
Adds the ability to define and validate enums.

### How to test

### DIFF
--- a/schemas/towns_fund.json
+++ b/schemas/towns_fund.json
@@ -120,6 +120,15 @@
         "parent_table": "Project_Dim",
         "parent_pk": "Project_ID"
       }
+    },
+    "enums": {
+      "Status": [
+        "1. Not yet started",
+        "2. Ongoing - on track",
+        "3. Ongoing - delayed",
+        "4. Completed",
+        "5. Other"
+      ]
     }
   },
   "Procurement": {
@@ -137,6 +146,21 @@
         "parent_table": "Project_Dim",
         "parent_pk": "Project_ID"
       }
+    },
+    "enums": {
+      "Status": [
+        "1. Not yet started",
+        "2. Ongoing - on track",
+        "3. Ongoing - delayed",
+        "4. Completed",
+        "5. Other"
+      ],
+      "Procurement_Status": [
+        "1. Publication of ITT",
+        "2. Evaluation of Tenders",
+        "3. Awarding of Construction Contract",
+        "4. Signing of Construction Contract"
+      ]
     }
   },
   "Project_Progress": {
@@ -171,6 +195,16 @@
         "parent_table": "Project_Dim",
         "parent_pk": "Project_ID"
       }
+    },
+    "enums": {
+      "Actual/Forecast": [
+        "Actual",
+        "Forecast"
+      ],
+      "PRA/Other": [
+        "PRA",
+        "Other"
+      ]
     }
   },
   "Capital_Data": {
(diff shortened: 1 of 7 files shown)


schemas/towns_fund.json
```diff
@@ -120,6 +120,15 @@
         "parent_table": "Project_Dim",
         "parent_pk": "Project_ID"
       }
+    },
+    "enums": {
+      "Status": [
```
nsnape, line 125: this is starting to feel a bit repetitious. Not sure if it's a problem, but for example, if the schema was written in python rather than json, the enum's could be more easily defined as something like "status": StatusEnum (rather than verbosely repeating all the options, then checking they match). What do you think?

nsnape: Since editing the schema will require inherent knowledge of the code anyway, even if we kept JSON, we could just switch this section to reference the enum name, then the schema validator can just check the name matches the actual Enum class?

nsnape: And the same in all the validation code. Can just check a field value is in an enum class, rather than in a recreated list?

cyrusdobbs: I think I agree to some extent. I think a consideration of your proposed approach is that we're then coupling the validation with the rest of the application code when at the moment the validation is completely separate

cyrusdobbs: >  then writing long verbose test assertions

what do you mean by this?

cyrusdobbs: We could alternatively define the enums separately in the JSON and then reference them by each sheet. i.e. define 'status' enum once and then reference it multiple times throughout the schema.

nsnape: Yeah I get your point re decoupling validation from the code. 

However, I think to some degree that this difference is academic to some point (at least at the moment) as the data itself will still ultimately have to synchronously pass both validation and then ingest into a DB. 

The idea of the validation is to capture errors that will prevent the data being inserted into our DB, and usefully represent them all to a user (rather than DB insert failure on the first incorrect instance). So if data fails validation, that data should not be able to be ingested into the DB without error. Likewise, in the case where the DB changes, and validation doesn't, ingest should also fail. So at the moment they are coupled. 

I think the dynamic / decoupled part is the idea that we may have further configurable data structures for different funds, but I think this is currently beyond our code - just a concept with no real requirements yet.

nsnape: > > then writing long verbose test assertions
> 
> what do you mean by this?

Actually, yeah they are not long, sorry (3 values is not long). But if we changed schema mechanism to use actual enum's rather than manually defined lists, we could create a test example enum, or even use a real one (from the code) for our tests, rather than writing a test that passes in a reconstructed enum equivalent like `"Non Existent Column": ["EnumValueA", "EnumValueB", "EnumValueC"]`. 

If the schema was updated to use an enum, the test could simply reference one too, which I think then also has the advantage of making it clearer what the test is testing.

nsnape: > r

but then we have 2 sources of truth for enum definition.

cyrusdobbs: using `json` provides no benefit with the current requirements and is much less useful than just using `python`

nsnape (approved):

PR #11, opened 2023-04-26
Instantiate data via ingest API call

nsnape:
https://trello.com/c/YOSvTOOJ
https://trello.com/c/ltVRaRu2

### Change description
Improved the "workflow" so that API ingest call actually loads data into fake DB. This data can then be queried via the usual get calls for project or package.

Previously all returned data was instantiated at import in a "hacky" indirect way. Removing/ uncoupling this unblocks more suitable end to end testing etc. 

We also now no longer store live data spreadsheets in the repo, these can be ingested via the UI from local sources.

### How to test

- run the flask debug server locally
- on swagger UI, ingest a spreadsheet via the ingest call:
- The Data Model project spreadsheet will probably fail and should return lots of useful validation messages. 
- The basic example spreadsheet in the test resources folder should load successfully
- Once ingested, run the get package or get project data calls with valid input ID's from the spreadsheet - you should get a payload back in the response containing all the data.


core/controllers/ingest.py
```diff
@@ -31,13 +32,9 @@ def ingest(body):
...
 
-    data = {}
-    for sheet_name, sheet in workbook.items():
-        sheet_as_dict = sheet.fillna("").to_dict(orient="split")
-        del sheet_as_dict["index"]
-        data[sheet_name] = sheet_as_dict
+    current_app.db = FakeDB.from_dataframe(workbook)
 
-    return data, 200
+    return "Success: Data ingested", 200
```
cyrusdobbs, line 37: Small point but we could align this with the example response in `openapi/components.yml`

tests/controller_tests/resources/response_assertion_data.py
```diff
@@ -52,57 +63,57 @@
...
+            "amount": 0,
             "end_date": "2020-04-30T00:00:00Z",
             "geography_indicator": "Town",
             "outcome": "Year on Year % Change in Monthly Footfall",
             "outcome_category": "Place",
             "project_id": "FHSFDCC001",
             "start_date": "2020-04-01T00:00:00Z",
             "state": "Actual",
-            "unit_of_measurement": "Year on Year % Change in Monthly Footfall",
+            "unit_of_measurement": "Year on Year % Change in Monthly " "Footfall",
```
cyrusdobbs, line 100: small point again but theres some ill formatting from copying from a test diff throughout this file

cyrusdobbs (changes requested):

core/controllers/ingest.py
nsnape: Sure thing

tests/controller_tests/resources/response_assertion_data.py
cyrusdobbs: what goes around comes around

cyrusdobbs (approved):

PR #12, opened 2023-04-26
Feature/104 implement sql alchemy

bobbi-hbrown:
https://trello.com/c/xo7Uvx3E/104-replace-fake-datastore-with-sqlite-sqlalchemy

### Change description
This PR adds an in-memory SQLite datastore alongside the mock DB so we can start building our models on top of it.

- [ X] Unit tests and other appropriate tests added or updated
- [ X] README and other documentation has been updated / added (if needed)
- [X ] Commit messages are meaningful and follow good commit message guidelines (e.g. "FS-XXXX: Add margin to nav items preventing overlapping of logo")

### How to test


core/db/db_models/contact.py
cyrusdobbs, line 1: I don't think we should split models into separate files like this. A file per model is just the same problem but at the file level - which IMO is worse and doesn't actually help organisation. I also don't see any obvious logical groups to put them in. I think illogical or unobvious groupings would be worse than one module full of models too.

Keen to hear what others think too though?

tests/conftest.py
```diff
@@ -3,6 +3,17 @@
...
 
+@pytest.fixture
+def app_ctx(flask_test_client):
+    """
+    Pushes an application context to a test.
+
+    :return: a test application context.
+    """
+    with flask_test_client.application.app_context():
+        yield
```
cyrusdobbs, line 14: I think we want to add this after `yield` to wipe the DB after each test.

```
db.session.remove()
db.drop_all()
```

nsnape (changes requested):

core/db/db_models/contact.py
```diff
@@ -0,0 +1,19 @@
+import uuid  # noqa
+
+from sqlalchemy import Column, String
+
+from core.db import db
+from core.db.types import GUID
+
+
+class Contact(db.Model):
```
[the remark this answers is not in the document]
bobbi-hbrown: Sounds good, we can start adding these now as we build out the models

nsnape (approved):

PR #16, opened 2023-04-28
[#PAB-226] Create Organisation Model 

nsnape:
https://trello.com/c/IiJFjd9v

### Change description
Added new SQLAlchemy model for Organisation table. 
Updated fk ref's in Contact to reference Organisation.
Basic test to sanity check the FK relations are set up as expected.


core/db/entities.py
```diff
@@ -1,19 +1,40 @@
...
 
 from core.db import db
 from core.db.types import GUID
 
 
+class Organisation(db.Model):
+    __tablename__ = "organisation_dim"
+
+    organisation_id = sqlalchemy.Column(GUID(), default=uuid.uuid4, primary_key=True)
+    organisation_name = sqlalchemy.Column(
```
gidsg, line 13: Can we just call these e.g. `id` and `name` as context is derivable from the table name.

core/db/entities.py
```diff
@@ -1,19 +1,40 @@
...
+        sqlalchemy.String, nullable=False, unique=True
+    )
+    geography = sqlalchemy.Column(sqlalchemy.String, nullable=True)
```
gidsg, line 16: Are we sure `geography` is a string? Can we confirm with Maeve?

core/db/entities.py
```diff
@@ -1,19 +1,40 @@
 import uuid  # noqa
 
-from sqlalchemy import Column, String
+import sqlalchemy
```
cyrusdobbs, line 3: small thing but I find long module names distracting when using imports like this, what do you think about  `import sqlalchemy as sqla` or `sa`?

core/db/entities.py
```diff
@@ -1,19 +1,40 @@
...
-    telephone = Column("telephone", String, nullable=True)
+    email_address = sqlalchemy.Column(sqlalchemy.String, nullable=False, unique=True)
+    contact_name = sqlalchemy.Column(sqlalchemy.String, nullable=True)
+    organisation_id = sqlalchemy.Column(
+        sqlalchemy.String,
+        sqlalchemy.ForeignKey("organisation_dim.organisation_id"),
+        nullable=False,
+    )
+    telephone = sqlalchemy.Column(sqlalchemy.String, nullable=True)
+    #
```
cyrusdobbs, line 37: can remove this

tests/model_tests/test_contact.py
```diff
@@ -1,22 +1,40 @@
 import uuid
 
+#
```
cyrusdobbs, line 3: and this

core/db/entities.py
```diff
@@ -1,19 +1,40 @@
...
+
+    contacts = sqlalchemy.orm.relationship("Contact", back_populates="organisation")
+
+
 class Contact(db.Model):
-    id = Column(
-        "id",
+    __tablename__ = "contact_dim"
+
+    contact_id = sqlalchemy.Column(
```
gidsg, line 24: As per previous comment, can we make these just `id`, `name` etc.

core/db/entities.py
gidsg, line 37: Looks like this line should be removed

core/db/entities.py
```diff
@@ -1,19 +1,40 @@
...
+    email_address = sqlalchemy.Column(sqlalchemy.String, nullable=False, unique=True)
+    contact_name = sqlalchemy.Column(sqlalchemy.String, nullable=True)
+    organisation_id = sqlalchemy.Column(
+        sqlalchemy.String,
+        sqlalchemy.ForeignKey("organisation_dim.organisation_id"),
+        nullable=False,
+    )
+    telephone = sqlalchemy.Column(sqlalchemy.String, nullable=True)
+    #
+    organisation = sqlalchemy.orm.relationship(
```
gidsg, line 38: Am a bit confused why we have both `organisation` and `organisation_id` here, do we know if we need both and how they map to underlying SQL?

tests/model_tests/test_contact.py
```diff
@@ -1,22 +1,40 @@
...
-from core.db.entities import Contact
+from core.db.entities import Contact, Organisation
 
 
-def test_contact(app_ctx):
+def test_contact_organisation(app_ctx):
+    """
+    Test basic relationship structure b/w Contact and Organisation.
+
+    Sanity check that fk relationships are looked up as expected.
```
gidsg, line 12: Can we replace this with more [inclusive naming](https://inclusivenaming.org/word-lists/tier-2/sanity-check/) e.g. confidence check

tests/model_tests/test_contact.py
```diff
@@ -1,22 +1,40 @@
...
+
+    organisation = Organisation(
+        organisation_name="Evil Corp",
+        geography="Earth",
+    )
+    db.session.add(organisation)
+    db.session.flush()  # get the db-generated UUID without committing to session.
+
+    read_org = Organisation.query.first()
+    assert read_org.organisation_name == "Evil Corp"
```
gidsg, line 23: Can we call this something more generic e.g. ACME corp?

core/db/entities.py
nsnape: I thought about the sqla option actually. Yeah can go for that

core/db/entities.py
nsnape: In my experience that can become confusing quickly when dealing with multiple tables with the same field name and deeply nested or complex queries. Verbose is good sometimes. 

Sometimes the context of the table name can be lost, take the example of simple test:
```
 read_org = Organisation.query.first()
 assert read_org.organisation_name == "Evil Corp"
```
This is obvious as they are right next to each other, but that var on the first line could be named anything, and that's all it takes to loose context. whereas organisation_name always means the same thing.

core/db/entities.py
gidsg: @nsnape seems like there is some uncertainty around this, see https://communities-govuk.slack.com/archives/C04UVF4U01X/p1681803446408639?thread_ts=1681734853.042499&cid=C04UVF4U01X can we add a TODO so we don't forget to review this?

core/db/entities.py
nsnape: whoops, not sure how that happened :/

core/db/entities.py
gidsg: I think there is value in e.g. using `id` consistently for the primary key of tables. So if we can accept that I may compromise on the other fields.
I would hope in real usage we would be returning e.g. Data classes rather than individual plucked columns so the context should also be described from that.

core/db/entities.py
cyrusdobbs: strike 2/3, one more and ChatGPT takes your job

core/db/entities.py
nsnape: so, organisation here is the name of the relationship to the whole table, not a single field of it. This is standard naming convention when referencing the parent entity from a child. All the SQL is handled under the hood by SQL Alchemy.

So now in ORM style queries, we can access the organisation from any contact via attribute lookups i.e. Contact.organisation.geography  -> it makes attributes available for the parent, directly from the child. Does that make sense, I'm happy to demo if not

tests/model_tests/test_contact.py
nsnape: they're everywhere :D

tests/model_tests/test_contact.py
nsnape: aha, yes that is a very good idea.

tests/model_tests/test_contact.py
nsnape: haha Evil corp is a generic term off Mr Robot. But yeah, I can make it even more generic

gidsg (approved):
👍

core/db/entities.py
gidsg: FYI @mgoodwin-version1

PR #19, opened 2023-05-02
[#PAB-226] create project progress sql model

cyrusdobbs:
https://trello.com/c/IiJFjd9v/226-create-additional-sqlalchemy-models

### Change description
Adds the ProjectProgress SQLAlchemy ORM model 

### How to test


core/db/entities.py
```diff
@@ -110,3 +110,51 @@ class Package(db.Model):
     m_and_e_contact = sqla.orm.relationship(
         "Contact", back_populates="m_and_e_packages", foreign_keys=[m_and_e_contact_id]
     )
+
+
+class ProjectProgress(db.Model):
+    """Stores Project Progress answers."""
+
+    # TODO: Would it not be better to model this as id, question_id, answer
```
nsnape, line 118: So we could do that, it would be a more "normalised" form of the data, but we would have to add additional constraints/logic to make sure it maps as we expect. For instance, we would need to track:

-  the question number (assume we have a max of 6 per package). This would require additional field (is this what you propose as question_id?). Then you would need extra unique-together constraints composed of package_id and question number, to make sure you don't duplicate question numbers for a package.
- make sure the questions are in order, ie package doesn't have only q1, q2, q4
- track up to 6 different fk relations to each package rather than 1. 

Although the current approach doesn't look as good wrt DB mapping theory, I think it is a much simpler way of representing the spreadsheet input directly. What do you think?

cyrusdobbs: Makes sense. I suppose at this point keeping it simple is best and we can change later if we want.

I assume we won't have a back populated variable covering ProjectProgress in Package then? We'll just have to query the ProjectPackage table using a PackageID

cyrusdobbs: Removed these comments

nsnape (approved):

PR #20, opened 2023-05-03
[#PAB-226] Create SQLAlchemy model for Project

nsnape:
https://trello.com/c/IiJFjd9v

### Change description
New model for Project entity

### DIFF
--- a/core/db/entities.py
+++ b/core/db/entities.py
@@ -81,45 +81,43 @@ class Package(db.Model):
         foreign_keys=[m_and_e_contact_id],
     )
 
+    projects = sqla.orm.relationship("Project", back_populates="package")
+
+
+class Project(db.Model):
+    """Stores Project Entities."""
+
+    __tablename__ = "project_dim"
+
+    id = sqla.Column(GUID(), default=uuid.uuid4, primary_key=True)  # this should be UUIDType once using Postgres
+    project_id = sqla.Column(sqla.String, nullable=False, unique=True)
+
+    # TODO: should this be unique?
+    project_name = sqla.Column(sqla.String, nullable=False)
+    package_id = sqla.Column(sqla.String, sqla.ForeignKey("package_dim.id"), nullable=True)
+
+    # TODO: Should we change this field to "Postcode" from "Address" to match example data.
+    #  Should we also have both as separate fields, or assume some front-end process combines them to be
+    #  stored here?
+    #  Also, should it be nullable?
+    address = sqla.Column(sqla.String, nullable=False)
+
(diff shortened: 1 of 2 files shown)


core/db/entities.py
```diff
@@ -62,42 +51,21 @@ class Package(db.Model):
...
-    )
-    m_and_e_contact_id = sqla.Column(
-        sqla.String,
-        sqla.ForeignKey("contact_dim.id"),
-        nullable=False,
-    )
+    name_contact_id = sqla.Column(sqla.String, sqla.ForeignKey("contact_dim.id"), nullable=False)
+    project_sro_contact_id = sqla.Column(sqla.String, sqla.ForeignKey("contact_dim.id"), nullable=False)
+    cfo_contact_id = sqla.Column(sqla.String, sqla.ForeignKey("contact_dim.id"), nullable=False)
+    m_and_e_contact_id = sqla.Column(sqla.String, sqla.ForeignKey("contact_dim.id"), nullable=False)
```
cyrusdobbs, line 68: could also do this tidying up for the ProjectProgress class for consistency? I just copied and pasted but I think it looks better when it's all on one line like this now that we have the longer line length

core/db/entities.py
```diff
@@ -108,9 +76,34 @@ class Package(db.Model):
...
+
+
+class Project(db.Model):
+    """Stores Project Entities."""
+
+    __tablename__ = "project_dim"
+
+    id = sqla.Column(GUID(), default=uuid.uuid4, primary_key=True)  # this should be UUIDType once using Postgres
+    project_id = sqla.Column(sqla.String, nullable=False, unique=True)
+    # TODO: should this be unique?
```
cyrusdobbs, line 94: Might be less messy if we try to put shorter comments on the same line as the variable?

cyrusdobbs: Theres a few others like this too

cyrusdobbs (changes requested):

core/db/entities.py
nsnape: yeah, I can do that in this PR

core/db/entities.py
nsnape: I don't mind where they go as the aim is to "do" them ie remove them ASAP. These read better on my screen TBH, so will leave as they are (I'll add a blank line above each to make it clearer). Also, the Git-diff looks cleaner when you remove them if they are not in-line

cyrusdobbs: makes sense

cyrusdobbs (approved):

PR #21, opened 2023-05-03
[#PAB-228] move test data to seed script in conftest

bobbi-hbrown:
https://trello.com/c/pHGBCS8S/228-create-test-seed-script-for-sql-db

### Change description

- [ X] Unit tests and other appropriate tests added or updated
- [X ] README and other documentation has been updated / added (if needed)
- [X ] Commit messages are meaningful and follow good commit message guidelines (e.g. "FS-XXXX: Add margin to nav items preventing overlapping of logo")

### How to test

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,7 +1,10 @@
 import pytest
 
+import tests.db_tests.resources.db_seed_data as db_seed
 from app import create_app
 from core.db import db
+from core.db.entities import Contact, Organisation
+from tests.util_test import seed_test_database
 
 
 @pytest.fixture
@@ -26,3 +29,22 @@ def flask_test_client():
     """
     with create_app().test_client() as test_client:
         yield test_client
+
+
+@pytest.fixture
+def seed_test_dataset(flask_test_client):
+    """
+    Pushes an application context to a test.
+
+    :return: a test application context.
+    """
+    with flask_test_client.application.app_context():
+        seed_test_database(Organisation, db_seed.ORGANISATION_DATA)
+        seed_test_database(Contact, db_seed.CONTACT_DATA)
+        # TODO: Add further models like this. In order -> package -> project -> the rest
+
+        db.session.commit()
+
+        yield
+        db.session.remove()
+        db.drop_all()
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -3,8 +3,19 @@ import pytest
 import tests.db_tests.resources.db_seed_data as db_seed
 from app import create_app
 from core.db import db
-from core.db.entities import Contact, Organisation
+from core.db.entities import Contact, Organisation, Package
 from tests.util_test import seed_test_database
+
+
+@pytest.fixture()
+def flask_test_client():
+    """
+    Creates the test client we will be using to test the responses
+    from our app, this is a test fixture.
+    :return: A flask test client.
+    """
+    with create_app().test_client() as test_client:
+        yield test_client
 
 
 @pytest.fixture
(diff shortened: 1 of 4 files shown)


tests/db_tests/resources/db_seed_data.py
```diff
@@ -17,4 +17,13 @@
...
 }
 
-# TODO add further test data for all models. Need to generate some new UUIDs for use in further table fk lookups.
+PACKAGE_DATA = {
+    "id": [uid["uuid7"], uid["uuid8"], uid["uuid9"]],
+    "package_name": ["Regeneration Project", "North Eastern Scheme", "Southern Access Project"],
+    "fund_type_id": ["HIJ", "LMN", "OPQR"],
+    "organisation_id": [str(uid["uuid1"]), str(uid["uuid2"]), str(uid["uuid3"])],
+    "name_contact_id": [str(uid["uuid4"]), str(uid["uuid5"]), str(uid["uuid6"])],
+    "project_sro_contact_id": [str(uid["uuid10"]), str(uid["uuid11"]), str(uid["uuid12"])],
```
nsnape, line 27: The last 3 fields in this table (project_sro_contact_id, cfo_contact_id, m_and_e_contact_id) the fk's need to reference a valid PK of Contact (see model definition of these fields for fk relations).

So according to the pre-defined CONTACT_DATA they will need to be in the range "uuid4" to "uuid6". These Contact rows can be referenced as many times as we want, so you can re-use the keys multiple times (since it's a many-to-one).

Ordinarily the data here would result in a FK ref not found type error, but it's turned off on SQLite as default (I'll fix this shortly)

tests/conftest.py
```diff
@@ -1,7 +1,10 @@
 import pytest
```
nsnape, line 1: Also, we seem to have got a CI pre-commit error. It's not from anything you have touched (it's somehow slipped through the CI), but it'll need fixing before this will build properly. Running pre-commit locally should sort it out

nsnape (changes requested):

tests/util_test.py
```diff
@@ -54,3 +55,13 @@ def test_postcode_to_itl1_post_code_area_not_exist_raises_error():
         postcode_to_itl1("ZZ1 2AB")
 
     assert exc_info.value.args[0] == 'Postcode Area "ZZ" is invalid and has no mapping.'
+
+
+def seed_test_database(model, model_data):
```
nsnape, line 60: This could do with a docstring really, sorry, it was me that committed the func without one. I would add now but don't want to conflict with your local branches. I can still do it if you'd prefer, let me know :)

nsnape (changes requested):

tests/conftest.py
```diff
@@ -26,3 +29,22 @@ def flask_test_client():
...
+def seed_test_dataset(flask_test_client):
+    """
+    Pushes an application context to a test.
+
+    :return: a test application context.
+    """
+    with flask_test_client.application.app_context():
+        seed_test_database(Organisation, db_seed.ORGANISATION_DATA)
+        seed_test_database(Contact, db_seed.CONTACT_DATA)
+        seed_test_database(Package, db_seed.PACKAGE_DATA)
```
nsnape, line 44: We also need to add seed data for the other 13 or so tables. If it's easier and you feel the ticket is getting too big we could get this tidied up, and merge in what we have, then add the remaining models in an additional feature branch?

Will need to rebase in / resolve conflicts for the fix for FK constraints that is in PR: https://github.com/communitiesuk/funding-service-design-post-award-data-store/pull/31 for this to work properly

nsnape (approved):

core/db/entities.py
```diff
@@ -33,6 +33,7 @@ class Contact(db.Model):
     organisation_id = sqla.Column(GUID(), sqla.ForeignKey("organisation_dim.id"), nullable=False)
     telephone = sqla.Column(sqla.String(), nullable=True)
 
+    # TODO: Do we actually need this orm.relationship boilerplate, or can we remove if just using standard queries?
```
cyrusdobbs, line 36: I minimise this boiler plate in #30

tests/conftest.py
cyrusdobbs, line 44: I propose that we could pass the `app_ctx` fixture to this function to reuse it and also rename this fixture to `seeded_app_ctx` and update the docstring

cyrusdobbs (changes requested):

tests/conftest.py
cyrusdobbs: i.e.

```
@pytest.fixture
def seeded_app_ctx(app_ctx):
    """
    Updated docstring
    """
    seed_test_database(Organisation, db_seed.ORGANISATION_DATA)
    seed_test_database(Contact, db_seed.CONTACT_DATA)
    seed_test_database(Package, db_seed.PACKAGE_DATA)
    db.session.commit()
    yield
```

core/db/entities.py
nsnape: Nice. removed the comment

tests/conftest.py
cyrusdobbs: Nice, the re-ordering makes sense. Yeah docstring looks all g

cyrusdobbs (approved):

PR #22, opened 2023-05-03
[#PAB-234] added postcode to ITL1 region mapping

cyrusdobbs:
https://trello.com/c/sdQoWmah/234-map-postcodes-to-itl-regions

### Change description
Added mapping for postcode to ITL1 region code

### How to test


core/const.py
```diff
@@ -70,3 +70,135 @@ class ProximityEnum(StrEnum):
     APPROACHING = "3. Approaching: next 6 months"
     CLOSE = "4. Close: next 3 months"
     IMMINENT = "5. Imminent: next month"
+
+
+# TLZ is given to any location outside the primary 12 ITL 1 regions as stated on the link below (previously NUTS & UKZ)
+# This introduces the problem that TLZ now maps to multiple locations (Channel Islands, Isle of Man, Non-geographic)
+# https://en.wikipedia.org/wiki/International_Territorial_Level
```
cyrusdobbs, line 77: @gidsg what do you think about this? Channel Islands and Isle of Man do not have their own ITL1 (previously NUTS1) codes. This means when we give them an 'other' code (UKZ), we can't then map it back to the actual location on the front-end. I imagine this is a problematic edge case? Or does this mean we just have 'Other'? Maybe a question for Maeve?

gidsg: Don't _think_ we'd fund projects in those regions so don't think this would concern us.

nsnape (commented):
looks reasonable for basic mapping function. If others are happy we might be able to merge this as a stand-alone util, and use it when we need it?

core/const.py
cyrusdobbs: ideal

core/util.py
```diff
@@ -0,0 +1,15 @@
...
+def postcode_to_itl1(postcode: str) -> str:
+    """
+    Maps a given UK postcode to its corresponding ITL1 code.
+
+    :param postcode: A string representing a UK postcode.
+    :return: A string representing the corresponding ITL1 code.
+    :raises IndexError: If the given postcode is invalid and has no mapping.
+    """
+    try:
+        return POSTCODE_TO_ITL1[postcode]
```
gidsg, line 13: We should be looking for a match at the start of the supplied full postcode. e.g. an input postcode like `W1A 1AA` should return `TLI` matching should also be case-insensitive.

cyrusdobbs: Good point 👍

gidsg: We probably should trim whitespace too to allow for malformed data

cyrusdobbs: This has been implemented now

tests/util_test.py
```diff
@@ -0,0 +1,41 @@
...
+def test_postcode_to_itl1_returns_itl1_leading_space():
+    itl1 = postcode_to_itl1(" dt1 8pd")
+    assert itl1 == "TLK"
+
+
+def test_postcode_to_itl1_returns_itl1_trailing_space():
+    itl1 = postcode_to_itl1("dt1 8pd ")
+    assert itl1 == "TLK"
+
+
```
gidsg, line 25: Can we test with a single letter example e.g. `N1 0AA` should return `TLI`?

core/util.py
```diff
@@ -0,0 +1,25 @@
...
+    postcode_area_matches = re.search("(^[A-z]{1,2})[0-9R][0-9A-z]?", postcode)
+
+    if not postcode_area_matches:
+        raise ValueError("Postcode is invalid.")
+
+    postcode_area = postcode_area_matches.groups()[0]
+    try:
+        return POSTCODE_AREA_TO_ITL1[postcode_area.upper()]
+    except KeyError:
+        raise KeyError(f'Postcode Area "{postcode_area}" is invalid and has no mapping.')
```
gidsg, line 25: Should we raise a custom exception e.g. `InvalidPostcodeError`?

core/util.py
```diff
@@ -0,0 +1,25 @@
...
+    :param postcode: A string representing a UK postcode.
+    :return: A string representing the corresponding ITL1 code.
+    :raises ValueError: If the given postcode is invalid.
+    :raises KeyError: If the postcode area (first 1-2 characters) of the given postcode is invalid and has no mapping.
+    """
+    postcode = postcode.strip()
+    postcode_area_matches = re.search("(^[A-z]{1,2})[0-9R][0-9A-z]?", postcode)
+
+    if not postcode_area_matches:
+        raise ValueError("Postcode is invalid.")
```
gidsg, line 19: As per below, would a custom exception here make more sense?

core/const.py
gidsg: @cyrusdobbs can we reference ONS in the comment rather than Wikipedia? e.g. https://geoportal.statistics.gov.uk/documents/ons::local-authority-district-april-2021-to-lau1-to-itl3-to-itl2-to-itl1-january-2021-lookup-in-united-kingdom-1/about

cyrusdobbs: The reason I linked the Wikipedia page is because it was the only place I saw 'TLZ' mentioned as a code for anywhere outside of the main 12 codes. I tried to search around ONS but couldn't find it anywhere

core/util.py
cyrusdobbs: I did think about this but came to the conclusion it wasn't really necessary because the original errors were still intuitive enough. I'll add it in anyway

tests/util_test.py
cyrusdobbs: yeah nice spot

core/util.py
gidsg: Makes sense not precious about this if you’d rather leave as-is

gidsg (approved):
LGTM 🗺️ ✉️ 👍

PR #25, opened 2023-05-03
[#PAB-226] create project delivery plan, procurement and risk register models

cyrusdobbs:
https://trello.com/c/IiJFjd9v/226-create-additional-sqlalchemy-models

### Change description
Added the final lot of models

### How to test

### DIFF
--- a/core/db/entities.py
+++ b/core/db/entities.py
@@ -419,3 +419,47 @@ class OutcomeDim(db.Model):
     outcome_category = sqla.Column(sqla.String(), nullable=False, unique=False)
 
     outcomes = sqla.orm.relationship("OutcomeData", back_populates="outcome_dim")
+
+
+class RiskRegister(db.Model):
+    """Stores Risk Register data for projects."""
+
+    __tablename__ = "risk_register"
+
+    id = sqla.Column(GUID(), default=uuid.uuid4, primary_key=True)  # this should be UUIDType once using Postgres
+
+    project_id = sqla.Column(sqla.String(), sqla.ForeignKey("project_dim.id"), nullable=False)
+    risk_name = sqla.Column(sqla.String(), nullable=False)
+
+    # TODO: Should this be an enum or is just free text?
+    risk_category = sqla.Column(sqla.String(), nullable=False)
+
+    # TODO: Are any of these nullable?
+    short_desc = sqla.Column(sqla.String(), nullable=False)
+    full_desc = sqla.Column(sqla.String(), nullable=False)
+    consequences = sqla.Column(sqla.String(), nullable=False)
+
+    pre_mitigated_impact = sqla.Column(sqla.Enum(const.ImpactEnum, name="risk_register_pre_mitigated_impact"))
+    pre_mitigated_likelihood = sqla.Column(
+        sqla.Enum(const.LikelihoodEnum, name="risk_register_pre_mitigated_likelihood")
+    )
+    mitigations = sqla.Column(sqla.String(), nullable=False)
+    post_mitigated_impact = sqla.Column(sqla.Enum(const.ImpactEnum, name="risk_register_post_mitigated_impact"))
+    post_mitigated_likelihood = sqla.Column(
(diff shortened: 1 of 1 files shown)


core/db/entities.py
```diff
@@ -109,6 +111,41 @@ class Project(db.Model):
...
+
+    id = sqla.Column(GUID(), default=uuid.uuid4, primary_key=True)  # this should be UUIDType once using Postgres
+    milestone = sqla.Column(sqla.String(), nullable=False)
+    project_id = sqla.Column(sqla.String(), sqla.ForeignKey("project_dim.id"), nullable=False)
+    start_date = sqla.Column(sqla.DateTime(), nullable=False)
+    end_date = sqla.Column(sqla.DateTime(), nullable=False)
+    status = sqla.Column(sqla.Enum(StatusEnum, name="project_delivery_plan_status"), nullable=False)
+
+    # TODO: Should this be nullable? Is Status alone enough in some circumstances?
+    comments = sqla.Column(sqla.String(), nullable=False)
```
nsnape, line 127: we should probably add a unique constraint to prevent duplicate data being entered. I would suggest an index composed of milestone and project_id. Then we can comment with the intention that data model people can check it?

core/db/entities.py
```diff
@@ -109,6 +111,41 @@ class Project(db.Model):
...
+
+
+class Procurement(db.Model):
+    """Stores Procurement data for projects."""
+
+    __tablename__ = "procurement"
+
+    id = sqla.Column(GUID(), default=uuid.uuid4, primary_key=True)  # this should be UUIDType once using Postgres
+
+    # TODO: Assuming contract is the PK and is unique, only other option I see is that contract + project is unique,
```
nsnape, line 137: id is the surrogate PK, and not really relevant, as, by definition it is always unique.

We should add a unique index consisting of contract and project id. Then remove the unique constraint on construction_contract as this would prevent different projects that happen to have the same name from entering them (as per that example data). But unique together in an index means each project cannot have duplicate contract names. Obvs data model people can review though

core/db/entities.py
```diff
@@ -227,3 +264,44 @@ class OutcomeDim(db.Model):
...
+    post_mitigated_impact = sqla.Column(sqla.Enum(ImpactEnum, name="risk_register_post_mitigated_impact"))
+    post_mitigated_likelihood = sqla.Column(sqla.Enum(LikelihoodEnum, name="risk_register_post_mitigated_likelihood"))
+    proximity = sqla.Column(sqla.Enum(ProximityEnum, name="risk_register_proximity"))
+
+    # TODO: Should this reference contact? Contact does not have a "role" field
+    risk_owner_role = sqla.Column(sqla.String(), nullable=False)
+
+    # TODO: does this unique index look right?
+    # Unique index for data integrity. There can't be multiple outcomes for a single project with
+    # the same date range and outcome metrics.
```
nsnape, line 299: I think the index is correct, but the TODO comment still has reference to "outcomes", "date range" and "outcome metrics"

nsnape (changes requested):

core/db/entities.py
cyrusdobbs: yeah good spot I missed that!

core/db/entities.py
cyrusdobbs: ahh ok makes sense!

core/db/entities.py
cyrusdobbs: ah mb i just copied and pasted

core/db/entities.py
```diff
@@ -109,6 +109,61 @@ class Project(db.Model):
...
+
+
+class Procurement(db.Model):
+    """Stores Procurement data for projects."""
+
+    __tablename__ = "procurement"
+
+    id = sqla.Column(GUID(), default=uuid.uuid4, primary_key=True)  # this should be UUIDType once using Postgres
+
+    construction_contract = sqla.Column(sqla.String(), nullable=False, unique=True)
```
nsnape: i think we still should remove this unique constraint on a single column

nsnape (approved):

PR #29, opened 2023-05-09
[FIX] incorrect procurement status enum

cyrusdobbs:
### Change description
Fixes a model that references the wrong Enum class

### How to test

### DIFF
--- a/core/db/entities.py
+++ b/core/db/entities.py
@@ -148,7 +148,8 @@ class Procurement(db.Model):
     project_id = sqla.Column(GUID(), sqla.ForeignKey("project_dim.id"), nullable=False)
     start_date = sqla.Column(sqla.DateTime(), nullable=False)
     end_date = sqla.Column(sqla.DateTime(), nullable=False)
-    status = sqla.Column(sqla.Enum(const.StatusEnum, name="procurement_plan_status"), nullable=False)
+    status = sqla.Column(sqla.Enum(const.StatusEnum, name="status"), nullable=False)
+    procurement_status = sqla.Column(sqla.Enum(const.ProcurementStatusEnum, name="procurement_status"), nullable=False)
 
     # TODO: Should this be nullable? Is Status alone enough in some circumstances?
     comments = sqla.Column(sqla.String(), nullable=False)


core/db/entities.py
```diff
@@ -148,7 +148,7 @@ class Procurement(db.Model):
     project_id = sqla.Column(sqla.String(), sqla.ForeignKey("project_dim.id"), nullable=False)
     start_date = sqla.Column(sqla.DateTime(), nullable=False)
     end_date = sqla.Column(sqla.DateTime(), nullable=False)
-    status = sqla.Column(sqla.Enum(const.StatusEnum, name="procurement_plan_status"), nullable=False)
```
nsnape, line 151: Good spot. Is this definitely correct now? According to the Enum Spreadsheet this is now right, but the original data model spreadsheet had 2 different status types in it for Procurement. Was this superseded when the Enum definitions were created (I'm just checking this isn't a missed field)?

cyrusdobbs: You're right, nice spot! Glad you checked it haha. This didn't fall over during run time because I had a mistake where I was overwriting 'status' with 'procurement_status' in the ingest PR

cyrusdobbs: So just deleting the original "status" data with "procuement_status" basically

cyrusdobbs: fixed this 👍

core/db/entities.py
```diff
@@ -149,6 +149,9 @@ class Procurement(db.Model):
     start_date = sqla.Column(sqla.DateTime(), nullable=False)
     end_date = sqla.Column(sqla.DateTime(), nullable=False)
     status = sqla.Column(sqla.Enum(const.StatusEnum, name="procurement_plan_status"), nullable=False)
+    procurement_status = sqla.Column(
+        sqla.Enum(const.ProcurementStatusEnum, name="procurement_plan_status"), nullable=False
```
nsnape, line 153: I think we should give this a unique name (it's currently set to same as the other "status")

nsnape (changes requested):

core/db/entities.py
cyrusdobbs: Fixed now

nsnape (approved):

PR #32, opened 2023-05-10
[#PAB-273] endpoints use sqlite

cyrusdobbs:
https://trello.com/c/Zzxwvzsa/273-hook-up-endpoints-to-sqlite-db

### Change description
Endpoints query the SQL db rather than the FakeDB

### How to test


core/controllers/project.py
```diff
@@ -28,109 +26,12 @@ def get_project(project_id: str):
     :param project_id: the project_id provided as a string
     :return: all data associated with a project
     """
-    project_data = get_by_project_id(table="project", project_id=project_id)
+    project = Project.query.filter_by(project_id=project_id).first()
 
-    if not project_data:
+    if not project:
```
nsnape, line 31: you're right, this is now "disgustingly simple" 🙌

core/db/entities.py
```diff
@@ -1,58 +1,65 @@
...
 from core import const
 from core.db import db
 from core.db.types import GUID
 
 
-class Organisation(db.Model):
+class BaseModel(db.Model):
+    __abstract__ = True
+
+    id: Mapped[int] = sqla.orm.mapped_column(
```
nsnape, line 15: I know this is a "pythonic" way of reducing the repeatability of defining the same id definition in every table, but I definitely think there's an argument for keeping model definitions verbose and individually defined. 

In this case I prefer the verbose way, as they are easy to read at a glance, and also easy to change if we, for example, decide to switch some PK's to be natural keys rather than surrogates. 

But happy to discuss. @cyrusdobbs @Joel-Pearce @gidsg @bobbi-hbrown any preference?

cyrusdobbs: Hmm yeah I get you. I do think the state of the models is still quite clear because _every_ model inherits `BaseModel` and has an `id` field. I think if some models did inherit `id` and some didn't then I think the argument for verboseness is stronger.

If we switched to natural keys then it would definitely have to change.

I'm happy for it to be kept like this or change it - I don't really have strong feelings.

core/db/entities.py
```diff
@@ -62,41 +69,41 @@ class Package(db.Model):
...
+    m_and_e_contact_id = sqla.orm.mapped_column(GUID(), sqla.ForeignKey("contact_dim.id"), nullable=False)
     m_and_e_contact = sqla.orm.relationship(
         "Contact",
-        back_populates="m_and_e_packages",
         foreign_keys=[m_and_e_contact_id],
     )
 
-    projects = sqla.orm.relationship("Project", back_populates="package")
+    projects: Mapped[List["Project"]] = sqla.orm.relationship()
+    progress: Mapped[List["ProjectProgress"]] = sqla.orm.relationship()
```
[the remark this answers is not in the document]
cyrusdobbs: Going to change to `progress_answers` if that sounds ok?

core/db/entities.py
```diff
@@ -107,17 +114,24 @@ class Project(db.Model):
     # TODO: should this be a fk to organisation?
     secondary_organisation = sqla.Column(sqla.String(), nullable=True)
 
-    package = sqla.orm.relationship("Package", back_populates="projects")
+    project_delivery_plans: Mapped[List["ProjectDeliveryPlan"]] = sqla.orm.relationship()
+    procurement_contracts: Mapped[List["Procurement"]] = sqla.orm.relationship()
+    direct_funds: Mapped[List["DirectFund"]] = sqla.orm.relationship()
+    capital: Mapped[List["Capital"]] = sqla.orm.relationship()
```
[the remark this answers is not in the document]
cyrusdobbs: `capital_records` and `progress_records`? 

I guess if we're going to make one plural we should probs do the same for the other. As progress can also be used a plural for progress...

core/db/entities.py
```diff
@@ -166,13 +178,13 @@ class Procurement(db.Model):
     )
 
 
-class ProjectProgress(db.Model):
+class ProjectProgress(BaseModel):
```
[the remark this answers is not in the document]
cyrusdobbs: sounds good will do

core/controllers/project.py
cyrusdobbs: 😆

core/db/entities.py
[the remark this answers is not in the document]
nsnape: sorry my bad, it should include id :/ I'll make that change.

core/db/entities.py
nsnape: will resolve for now, we can revisit if it becomes a problem

nsnape (approved):

PR #34, opened 2023-05-15
Feature/228 populate test data

bobbi-hbrown:
https://trello.com/c/pHGBCS8S/228-create-test-seed-script-for-sql-db

### Change description
Populates the test database with the rest of the test data.

- [X ] Unit tests and other appropriate tests added or updated
- [ X] README and other documentation has been updated / added (if needed)
- [ X] Commit messages are meaningful and follow good commit message guidelines (e.g. "FS-XXXX: Add margin to nav items preventing overlapping of logo")

### DIFF
--- a/tests/db_tests/resources/db_seed_data.py
+++ b/tests/db_tests/resources/db_seed_data.py
@@ -117,7 +117,7 @@ INDIRECT_FUND_SECURED_DATA = {
         enums.FundingSourceCategoryEnum.THIRD_SECTOR_FUNDING,
     ],
     "state": [enums.StateEnum.ACTUAL, enums.StateEnum.FORECAST, enums.StateEnum.FORECAST],
-    "amount": ["£0.00", "£1,000,000", "£12,345,678.00"],
+    "amount": [0.00, 1000000.00, 12345678.00],
 }
 
 INDIRECT_FUND_UNSECURED_DATA = {
@@ -132,14 +132,14 @@ INDIRECT_FUND_UNSECURED_DATA = {
         enums.FundingSourceCategoryEnum.THIRD_SECTOR_FUNDING,
     ],
     "state": [enums.StateEnum.ACTUAL, enums.StateEnum.FORECAST, enums.StateEnum.FORECAST],
-    "amount": ["£1,000,000", "£12,345,678.00", "£0.00"],
+    "amount": [1000000.00, 12, 345, 678.00, 0.00],
     "current_status": ["1. Not yet started", "2. Ongoing - on track", "3. Ongoing - at risk"],
     "comments": [
         "Lorem ipsum dolor sit amet, consectetur adipiscing elit.",
         "sed do eiusmod tempor incididunt ut labore et dolore magna aliqua",
         "Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat.",
     ],
-    "potential_secure_date": ["01/04/2021", "11/03/2022", "14/01/2023"],
+    "potential_secure_date": [datetime.date(2021, 4, 1), datetime.date(2022, 11, 3), datetime.date(2023, 1, 14)],
 }
 
 OUTPUT_DATA = {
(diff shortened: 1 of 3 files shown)


tests/db_tests/test_entities.py
```diff
@@ -97,10 +97,21 @@ def test_database_seed(seeded_app_ctx):
...
+    # Check parent table's row fields available as ORM attributes
     assert read_package.cfo_contact.organisation.organisation_name == "Test Organisation"
 
+    # Check ability to query/read a project
+    read_project = Project.query.first()
+    assert read_project.project_name == "Project 1"
+    assert read_project.address == "SW1A 2AA"
+    assert read_project.secondary_organisation == "Org 1"
+    assert read_project.package_id == read_package.id
+
```
nsnape, line 114: Could we add a new test to access the children of Project, rather than just the parents. These are now accessible as ORM attributes directly from project, and should be returned in a list. In your seed data most projects just have one child in each child table so your list should be 1 long. Example implementation:

`assert read_project.outputs[0] == OutputData.query.first()`

nsnape: The reason I've asked for this is that this method of accessing child attributes is specific to the way we've implemented ORM relationships from parent to child. So the test is helpful in case that method gets inadvertently changed.

It may be worth adding it as a completely new test for this reason. The test will probably only have 1 line (ie the example line above) but that's ok - it's just testing the method, not the whole test data integrity.

nsnape (changes requested):

nsnape (changes requested):
re Git workflow. Could you either rebase them into more succinct commit messages/groupings, or if that's too complicated now, squash into 1 commit and give it a good message, referencing ticket and clear description of overall work done. For example, you don't want references to WIP in your committed feature history - it's irrelevant and misleading. 

Is that okay? Let me know if you need any help

tests/db_tests/resources/db_seed_data.py
```diff
@@ -28,3 +31,193 @@
...
+    "contractually_committed_amount": [1000000.00, 12345678.00, 0.00],
+}
+
+CAPITAL_DATA = {
+    "id": [uid["uuid25"], uid["uuid26"], uid["uuid27"]],
+    "project_id": [uid["uuid10"], uid["uuid11"], uid["uuid12"]],
+    "start_date": [datetime.date(2021, 4, 1), datetime.date(2022, 11, 3), datetime.date(2023, 1, 14)],
+    "end_date": [datetime.date(2022, 3, 20), datetime.date(2023, 3, 10), datetime.date(2024, 1, 13)],
+    "state": [enums.StateEnum.ACTUAL, enums.StateEnum.FORECAST, enums.StateEnum.FORECAST],
+    "amount": [2, 000, 000.00, 0.00, 10, 345, 000.00],
```
cyrusdobbs, line 105: It doesn't make the code fall over so not major but I think we only need 3 values here? The same for the amount value in `INDIRECT_FUND_UNSECURED_DATA`

nsnape (changes requested):

tests/db_tests/resources/db_seed_data.py
cyrusdobbs: Looks like `INDIRECT_FUND_UNSECURED_DATA` still has 5 values

tests/db_tests/resources/db_seed_data.py
```diff
@@ -28,3 +31,193 @@
...
+    "start_date": [datetime.date(2021, 4, 1), datetime.date(2022, 11, 3), datetime.date(2023, 1, 14)],
+    "end_date": [datetime.date(2022, 3, 20), datetime.date(2023, 3, 10), datetime.date(2024, 1, 13)],
+    "funding_source_name": ["Private Funding", "Local Authority", "Private Funding"],
+    "funding_source_category": [
+        enums.FundingSourceCategoryEnum.THIRD_SECTOR_FUNDING,
+        enums.FundingSourceCategoryEnum.OTHER_PUBLIC_FUNDING,
+        enums.FundingSourceCategoryEnum.THIRD_SECTOR_FUNDING,
+    ],
+    "state": [enums.StateEnum.ACTUAL, enums.StateEnum.FORECAST, enums.StateEnum.FORECAST],
+    "amount": [1000000.00, 12, 345, 678.00, 0.00],
```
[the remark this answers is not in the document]
nsnape: Hey, this still has 5, please switch to 3, then we can merge.

nsnape (changes requested):

nsnape (approved):

PR #39, opened 2023-05-22
PAB-297: update copilot app to deploy as a backend service

gidsg:
### Change description
In preparation for the Pen Test, we need to re-configure the API to be a [Backend Service](https://aws.github.io/copilot-cli/docs/manifest/backend-service/), so it is not exposed on the public internet.
This also changes the parent 'app' name to post-award to be in better keeping with copilot conventions

### How to test
Should deploy successfully after merge

### DIFF
--- a/copilot/.workspace
+++ b/copilot/.workspace
@@ -1,1 +1,1 @@
-application: post-award-data-store
+application: post-award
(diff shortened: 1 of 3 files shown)


cyrusdobbs (approved):
Approving to test deployment

PR #41, opened 2023-05-23
Feature/ingestion round 1 data

Joel-Pearce:

### Change description

### How to test

### DIFF
--- a/core/extraction/round_1.py
+++ b/core/extraction/round_1.py
@@ -1,0 +1,417 @@
+import pandas as pd
+
+
+def ingest_round_1_data(df_map: pd.DataFrame) -> dict[pd.DataFrame]:
+    """
+    Extract data from Round 1 Reporting Template into column headed Pandas DataFrames.
+
+    :param df_ingest: DataFrame of parsed Excel data.
+    :return: Dictionary of extracted "tables" as DataFrames.
+    """
+    df_dictionary = dict()
+
+    df_dictionary["programme_outcomes"] = extract_programme_outcomes(df_map["programme_outcomes"])
+    df_dictionary["programme_risks"] = extract_programme_risks(df_map["programme_risks"])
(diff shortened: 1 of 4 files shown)


core/extraction/round_1.py
```diff
@@ -0,0 +1,208 @@
...
+    Extract programme outcomes information from a DataFrame.
+
+    Input dataframe is parsed from Excel spreadsheet: "Round 1 Reporting - TF_aggregated_data_DD.MM.YYYY.xlsx".
+    Specifically Project work sheet, parsed as dataframe.
+
+    :param df: Input DataFrame containing data.
+    :return: Extracted DataFrame containing programme data.
+    """
+
+    df_programme = df_programme.reset_index(drop=True)
```
nsnape, line 42: I don't think this actually does anything as we haven't altered the dataframe at all (I just checked on debugger and the indexes look ok without resetting them)

core/extraction/round_1.py
```diff
@@ -0,0 +1,208 @@
...
+def ingest_round_1_data(df_map: pd.DataFrame) -> dict[pd.DataFrame]:
+    """
+    Extract data from Round 1 Reporting Template into column headed Pandas DataFrames.
+
+    :param df_ingest: DataFrame of parsed Excel data.
+    :return: Dictionary of extracted "tables" as DataFrames.
+    """
+    df_dictionary = dict()
+
+    df_dictionary["programme_outcomes"] = extract_programme_outcomes(df_map["programme_outcomes"])
```
nsnape, line 13: ah, so I've just realised, that since we don't actually need to do any manipulation to get these sheets in, and we are just naming them the same thing they are already called on the tabs, we don't need to run any individual extraction methods at all. pd.read_excel already does all that. Sorry. 

Effectively, input param df_map is the same as df_dictionary in this function.

However, what we have is probably useful as a placeholder structure for when we do some transformation, so it can be left in the current format for now :)

nsnape (changes requested):

PR #45, opened 2023-05-26
[#PAB-361] seed db on startup

cyrusdobbs:
https://trello.com/c/XriMEDz1/361-prep-api-for-pen-test

### Change description
Seed the database on start-up with some test data because the ingest endpoint is inaccessible in the test environment.

### How to test

### DIFF
--- a/app.py
+++ b/app.py
@@ -9,6 +9,7 @@ from fsd_utils.healthchecks.healthcheck import Healthcheck
 from fsd_utils.logging import logging
 from sqlalchemy import event
 
+from core.cli import create_cli
 from core.db import db
 from core.errors import ValidationError, validation_error_handler
 from core.validation.schema import parse_schema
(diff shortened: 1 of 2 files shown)


app.py
```diff
@@ -48,6 +49,7 @@ def _fk_pragma_on_connect(dbapi_con, con_record):  # noqa
     with flask_app.app_context():
         event.listen(db.engine, "connect", _fk_pragma_on_connect)
         db.create_all()
+        seed_db()  # TODO: Remove this after pen test
```
gidsg, line 52: Sorry for change of heart, can we model this as a [Flask Custom Command](https://flask.palletsprojects.com/en/2.3.x/cli/#custom-commands) as well, that way it will be easier to run once we connect up to Postgres and we need to run it one time.

tests/seed.py
```diff
@@ -0,0 +1,24 @@
+from core.db import db
+from core.db import entities as ents
+from tests.db_tests.resources import db_seed_data as db_seed
+from tests.util_test import seed_test_database
+
+
+def seed_db():
```
nsnape, line 7: maybe a basic docstring?

cyrusdobbs: removed this change now

app.py
cyrusdobbs: nice idea - implemented 👍

gidsg (approved):
LGTM 👍 will test on AWS once deployed and we can tweak if needs be.

PR #46, opened 2023-05-26
[#PAB-362] add download endpoint

cyrusdobbs:
https://trello.com/c/wBlZvwpn/362-create-data-as-file-api-endpoint

### Change description
Adds an endpoint that returns some data in either xlsx or json format. The data returned is just example data for now.

### How to test

### DIFF
--- a/openapi/api.yml
+++ b/openapi/api.yml
@@ -6,20 +6,35 @@ info:
   version: 0.1.9
 
 paths:
-  /data:
+  /download:
     get:
-      operationId: core.controllers.examples.example
-      summary: Returns a list of data.
-      description: Optional extended description in CommonMark or HTML.
+      operationId: core.controllers.download.download
+      summary: Returns some data in the requested format.
+      description: Takes a file format and returns some data in that format.
+      parameters:
+        - in: query
+          name: file_format
+          schema:
+            type: string
+            enum: [ json, xlsx ]
+          required: true
+          description: Specify the desired file format (json or xlsx)
       responses:
-        '200': # status code
-          description: A JSON array of data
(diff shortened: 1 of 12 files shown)


openapi/api.yml
```diff
@@ -6,20 +6,31 @@ info:
...
 paths:
-  /data:
+  /download:
     get:
-      operationId: core.controllers.examples.example
-      summary: Returns a list of data.
-      description: Optional extended description in CommonMark or HTML.
+      operationId: core.controllers.download.download
+      summary: Returns some data in a binary format.
+      description: Takes a file format and returns some data in that format.
```
gidsg, line 13: Guess it's only binary if excel, so maybe say, 'in requested format' or similar?

nsnape (commented):
LGTM, nice and simple. I guess this will hopefully just work when we want to output the updated DM data? 

The only thing we may need to add is a little transformation before it gets spat out, but handy if it's in a Dict of dataframes

openapi/api.yml
cyrusdobbs: changed this

gidsg (approved):
LGTM 👍 🌸

cyrusdobbs:
> LGTM, nice and simple. I guess this will hopefully just work when we want to output the updated DM data?
> The only thing we may need to add is a little transformation before it gets spat out, but handy if it's in a Dict of dataframes

It will work seamlessly for excel but for json we'll need to change the call to the Package model to the Programme model, but yeah. It will need more changes than this though obvs because at the moment it just grabs _something_ from the db and returns it because we dont care about the contents yet

PR #50, opened 2023-05-30
BAU: Use absolute db file path

gidsg:
### Change description
App is failing with permission denied errors, hypothesis is that it is trying to write the sqlite db to the root path and hitting permissions issues, change to system tempfile path (which should be writable)

### How to test
Should deploy successfully and data should be persisted after running seed script

### DIFF
--- a/app.py
+++ b/app.py
@@ -1,4 +1,5 @@
 import os
+import tempfile
 from copy import deepcopy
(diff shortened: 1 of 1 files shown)


endamccormack (approved):
Lgtm

PR #51, opened 2023-05-30
BAU: seed data after deployment

gidsg:
### Change description
We need to seed data after deployment, so it is not wiped after each merge.

### How to test
Data should be available post-deployment
Deploy should complete successfully


cyrusdobbs (approved):

endamccormack (approved):
lgtm 👍🏻

PR #52, opened 2023-05-30
[#PAB-295] use postgres db

cyrusdobbs:
https://trello.com/c/zZ7XRHeA/295-setup-postgres-database

### Change description
Add configuration options to connect to prostgres db. Also adds a new CLI command to drop all data from the database.

### How to test

### DIFF
--- a/app.py
+++ b/app.py
@@ -1,5 +1,3 @@
-import os
-import tempfile
 from copy import deepcopy
 from pathlib import Path
 
@@ -10,6 +8,7 @@ from fsd_utils.healthchecks.healthcheck import Healthcheck
 from fsd_utils.logging import logging
 from sqlalchemy import event
 
+from config import Config
 from core.cli import create_cli
 from core.db import db
 from core.errors import ValidationError, validation_error_handler
@@ -20,7 +19,7 @@ from schemas.towns_fund import TF_SCHEMA
 WORKING_DIR = Path(__file__).parent
 
 
-def create_app() -> Flask:
+def create_app(config_class=Config) -> Flask:
     connexion_options = {"swagger_url": "/"}
     connexion_app = connexion.FlaskApp(
         "Sample API",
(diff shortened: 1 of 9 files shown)


app.py
```diff
@@ -61,4 +54,12 @@ def _fk_pragma_on_connect(dbapi_con, con_record):  # noqa
     return flask_app
 
 
+def enable_sqlite_fk_constraints(flask_app):
```
nsnape, line 57: could we stick a very basic docstring on this now it's in it's own function? Just so it's clear what it does. As not 100% obvious at first. The comment out of the old code would probably suffice.

config/__init__.py
```diff
@@ -0,0 +1,34 @@
...
+
+if not FLASK_ENV:
+    raise KeyError("FLASK_ENV does not exist in environ")
+
+match FLASK_ENV:
+    case "unit_test":
+        from config.envs.unit_test import UnitTestConfig as Config
+    case "development":
+        from config.envs.development import DevelopmentConfig as Config
+    # case "dev":
```
nsnape, line 14: can these be removed?

config/__init__.py
```diff
@@ -0,0 +1,34 @@
...
+    #     )
+    # case "uat" | "production":
+    #     from config.envs.production import (
+    #         ProductionConfig as Config,
+    #     )
+    case _:
+        from config.envs.default import DefaultConfig as Config
+
+try:
+    Config.pretty_print()
```
nsnape, line 30: Is this just a user message? Do we need it, or can it be replaced with logging?

app.py
nsnape: Maybe with a slight addition about what the intended use is? (is it for unit testing? reading the PR content below)

config/envs/default.py
```diff
@@ -0,0 +1,11 @@
...
+from pathlib import Path
+
+from fsd_utils import configclass
+
+
+@configclass
+class DefaultConfig(object):
+    FLASK_ROOT = str(Path(__file__).parent.parent.parent)
+
+    SQLALCHEMY_DATABASE_URI = environ.get("DATABASE_URL")
```
gidsg, line 11: As it stands now this will break the AWS environment as it needs the temp file path which is removed by this PR, please restore that functionality, i.e. make this something like:
```
    SQLALCHEMY_DATABASE_URI = environ.get("DATABASE_URL", f"sqlite:///{tempfile.gettempdir()}/sqlite.db")
```

config/__init__.py
cyrusdobbs: It's from `fsd_utils` and prints the config into a table which shows where each value is defined when the app starts - seems useful to keep IMO

https://github.com/communitiesuk/funding-service-design-utils/blob/8ab0dd38130ea12039fb15d7f79224a57f51e665/fsd_utils/config/configclass.py#LL29C28-L29C28

gidsg (changes requested):

app.py
cyrusdobbs: yeah good shout 👍

config/envs/default.py
cyrusdobbs: 👍 nice spot

app.py
```diff
@@ -61,4 +54,20 @@ def _fk_pragma_on_connect(dbapi_con, con_record):  # noqa
     return flask_app
 
 
+def enable_sqlite_fk_constraints(flask_app: Flask) -> None:
+    """Enable FK constraints for an SQLite DB.
+
+    NOTE: We currently only use SQLite for unit testing.
```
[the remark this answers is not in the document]
cyrusdobbs: ah true

nsnape (approved):
LGTM

gidsg (approved):
🍡 👍

PR #53, opened 2023-05-30
PAB 297: add postgres connection string for AWS

gidsg:
### Change description
Add required postgres connection string for AWS 

### How to test
Deployment should work
Should connect to Postgres instance on AWS


cyrusdobbs (approved):
LGTM 👍

PR #55, opened 2023-05-31
PAB-297: drop db prior to seed

gidsg:
### Change description
Now we are using Postgres which persists state we need to drop tables in order to pick up schema changes between deployments Note: we will change to use proper migrations prior to production deployment

### How to test
Should deploy successfully on merge

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -30,6 +30,7 @@ jobs:
         run: |
           copilot deploy
 
-      - name: Seed data
+      - name: Recreate DB and Seed data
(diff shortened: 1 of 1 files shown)


endamccormack (approved):
temporary change I expect?

gidsg:
> temporary change I expect?
@endamccormack yes, see comment in PR https://github.com/communitiesuk/funding-service-design-post-award-data-store/blob/a9d9054fc65b486efdaeb3768513605a66c55e34/.github/workflows/deploy.yml#L34

PR #58, opened 2023-05-31
[#PAB-331] R3 Transformation: Project Details

nsnape:
[#PAB-331] https://trello.com/c/97RfNJ9h

Project Details DataFrame transformed to match "Load" format (minus submission id) for ingestion into DB.

Changes in PR:
+++ Project id lookup table extracted from form
+++ lookup project id on project details DataFrame 
+++ Squash multiple/single columns together

### Change description

### DIFF
--- a/core/extraction/towns_fund.py
+++ b/core/extraction/towns_fund.py
@@ -16,7 +16,10 @@ def ingest_towns_fund_data(df_ingest: pd.DataFrame) -> Tuple[Dict[str, pd.DataFr
     """
 
     towns_fund_extracted = {"df_place_extracted": extract_place_details(df_ingest["2 - Project Admin"])}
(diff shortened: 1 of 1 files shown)


cyrusdobbs (commented):
nice one - good to see some output from the ETL stuff!

from comparing the output project df to the validation / db schema + other thoughts (not really looked at the code):

- missing submission id (I guess this will be added in a later step?)
- missing programme id
- db will accept any string for locations and lat_long but the current output contains escape characters and other characters - I assume we'll want to clean this prior to db load
- lat long is just a copy of locations atm - i guess this is intentional for now?
- the db will accept Yes / No or null for gis_provided but I see `< select >` in the df - this doesn't look like a numpy null so I assume we'll need to clean that too

DonShelly (approved):

cyrusdobbs (approved):
LGTM

PR #61, opened 2023-06-01
[FIX] remove organisation_id from Organisation entity

cyrusdobbs:
### Change description
The data model example contained a natural key `organisation_id` for Organisation but we don't need this and will just use the UUID `id` as a surrogate key instead.

### How to test


core/validation_schema.py
```diff
@@ -13,21 +13,20 @@
     },
     "Organisation_Ref": {
         "columns": {
-            "Organisation ID": "str",
             "Organisation": "str",
             "Geography": "str",
         },
-        "uniques": ["Organisation ID", "Organisation"],  # TODO: Assuming we dont want multiple Orgs with the same name
+        "uniques": ["Organisation"],  # TODO: Assuming we dont want multiple Orgs with the same name
```
nsnape, line 19: yes, that's correct, can remove the comment

core/controllers/mappings.py
```diff
@@ -101,9 +100,9 @@ def lookup_fk_row(model: db.Model, lookup_field: str, lookup_val: Any) -> str |
             "Programme ID": "programme_id",
             "Programme Name": "programme_name",
             "FundType_ID": "fund_type_id",
-            "Organisation ID": "organisation_id",
+            "Organisation": "organisation",
         },
-        fk_relations=[("organisation_id", ents.Organisation, "organisation_id", "organisation_id")],
+        fk_relations=[("organisation_name", ents.Organisation, "organisation_id", "organisation")],
```
nsnape, line 105: should item 3 in this tuple now be "organisation_name"?

nsnape (changes requested):

core/controllers/mappings.py
cyrusdobbs: no, because `Programme.organisation_id` is what we call the FK attribute that relates to the `Organisation.id`

core/validation_schema.py
cyrusdobbs: removed 👍

nsnape (approved):

PR #63, opened 2023-06-01
[#PAB-332] Transform round 3 Project Progress and Funding Questions

nsnape:
https://trello.com/c/97RfNJ9h

### Change description

Transform `Programme Progress` ready to load into DB
Transform `Funding Question` ready to load
Minor change to nullability on funding Q DB schema

### DIFF
--- a/core/db/entities.py
+++ b/core/db/entities.py
@@ -135,7 +135,7 @@ class FundingQuestion(BaseModel):
     programme_id: Mapped[int] = sqla.orm.mapped_column(sqla.ForeignKey("programme_dim.id"), nullable=False)
 
     question = sqla.Column(sqla.String(), nullable=False)
-    indicator = sqla.Column(sqla.String(), nullable=False)
+    indicator = sqla.Column(sqla.String(), nullable=True)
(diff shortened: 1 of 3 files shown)


cyrusdobbs:
We have `date_of_important_milestone` down as a string atm (I think we said it was free text) but the transformed data contains datetimes - can this field actually be a `DataTime`?

cyrusdobbs:
`FundingQuestion.response` data contains `< Select >` - does this want to be null?

nsnape:
> `FundingQuestion.response` data contains `< Select >` - does this want to be null?

I think in this case Im happy to leave is as '< Select >' - as it indicates they failed to choose a drop-down option, rather than leaving a free-text field empty. It's because we have a mix of "indicators" in that column, as the table is "unpivoted"

nsnape:
> We have `ProjectProgress.date_of_important_milestone` down as a string atm (I think we said it was free text) but the transformed data contains datetimes - can this field actually be a `DataTime`?

yeah, I'll change on this PR

cyrusdobbs (approved):
LGTM

PR #65, opened 2023-06-02
#PAB-421: added function for getting organisation names for filter

Joel-Pearce:
Added function to get a list of organisation names from the database for the frontend filter.

### Change description

### How to test

### DIFF
--- a/tests/controller_tests/test_filters.py
+++ b/tests/controller_tests/test_filters.py
@@ -1,0 +1,27 @@
+from flask.testing import FlaskClient
+
+
+def test_get_organisation_names_failure(app: FlaskClient):
+    """Asserts failed retrieval of organisation names."""
+
+    response = app.get("/organisations")
+
+    assert response.status_code == 404
+    assert response.json["detail"] == "No organisation names found."
+
+
+def test_get_organisation_names(seeded_app_ctx):
+    """Asserts successful retrieval of organisation names."""
+
+    response = seeded_app_ctx.get("/organisations")
+
+    assert response.status_code == 200
+    assert response.content_type == "application/json"
+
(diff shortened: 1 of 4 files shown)


tests/controller_tests/test_filters.py
```diff
@@ -0,0 +1,19 @@
...
+    assert response.json["detail"] == "No organisation names found."
+
+
+def test_get_organisation_names(seeded_app_ctx):
+    """Asserts successful retrieval of organisation names."""
+
+    response = seeded_app_ctx.get("/organisations")
+
+    assert response.status_code == 200
+    assert response.content_type == "application/json"
```
gidsg, line 19: Can we assert something about the request body also?

nsnape: For me this is a weird test since we are evaluating one long desterilised sting output rather than the JSON object. For leaving it like this, maybe a better assertion would be:
`assert '"name"' in response...` to make it check more specifically a JSON key. 

Or even better, do your assertions on response.json rather than response.data? Because then we can test the structure and positions of arguments in the response.

nsnape (changes requested):

tests/controller_tests/test_filters.py
Joel-Pearce: I've changed the test to now inspect response.json to check if 'name' and 'id' are keys.

```
    response_json = response.json

    assert any("name" in org for org in response_json)
    assert all(isinstance(org["name"], str) for org in response_json)

    assert any("id" in org for org in response_json)
    assert all(isinstance(org["id"], str) for org in response_json)
```

nsnape (approved):

PR #68, opened 2023-06-05
[#PAB-332] Transform round 3 PSI and Risk Register

nsnape:
https://trello.com/c/97RfNJ9h

### Change description
Transform PSI ready to load into DB ` PrivateInvestments`
Transform and combine Project and programme risks, into `RiskRegister`
Tweak DB table to allow data for cancelled project, as per example (data rows are empty, but cancellation summary provided in risk name column)

### DIFF
--- a/core/db/entities.py
+++ b/core/db/entities.py
@@ -405,23 +405,30 @@ class RiskRegister(BaseModel):
     project_id: Mapped[int] = sqla.orm.mapped_column(sqla.ForeignKey("project_dim.id"), nullable=True)
 
     risk_name = sqla.Column(sqla.String(), nullable=False)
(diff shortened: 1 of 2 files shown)


cyrusdobbs:
Row 18 of the Risk Register has null values (all but project id and riskname) where the db is defined as non-nullable

nsnape:
> ll values (all bu

Yeah, that's a valid input on the form I think. I changed the nullability of these fields in the DB as part of this PR, to allow for partially completed risks. Might be we add more logic around this as part of validation, after MVP1

cyrusdobbs:
ah my bad I was comparing with the previous versions entity class

cyrusdobbs (approved):
LGTM

PR #73, opened 2023-06-06
BAU: Adds script to create tunnel to container

gidsg:
Co-authored-by: Aaron Williams <aaron.williams@version1.com>

### Change description
Adds script to tunnel to container.

(this will be used as the initial approach to POST to the ingest endpoint
on the api)

### How to test
Requires AWS access but have confirmed it works as described.


cyrusdobbs (approved):
Nice :)

PR #74, opened 2023-06-06
[#PAB-426] get outcome categories endpoint

cyrusdobbs:
https://trello.com/c/oluqlVNU/426-endpoint-list-of-outcome-categories

### Change description
Implements a GET endpoint for the frontend to get all the possible outcome categories in the data.

### How to test

### DIFF
--- a/core/controllers/get_filters.py
+++ b/core/controllers/get_filters.py
@@ -1,7 +1,7 @@
 from flask import abort
 
 from core.const import FUND_ID_TO_NAME
-from core.db.entities import Organisation, Programme
+from core.db.entities import Organisation, OutcomeDim, Programme
 
 
 def get_organisation_names():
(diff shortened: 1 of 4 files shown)


core/controllers/get_filters.py
```diff
@@ -32,3 +32,18 @@ def get_funds():
...
+
+
+def get_outcome_categories():
+    """Returns a list of all outcome categories.
+
+    :return: List of outcome categories
+    """
+    outcome_dims = OutcomeDim.query.with_entities(OutcomeDim.outcome_category).distinct().all()
+
+    if not outcome_dims:
```
nsnape, line 44: is this definitely how we want it to behave? (it probably is, I'm just checking). As there's a real chance we won't have any Categories stored (ie the column may have nan's in if we don't get given some mappings), in this (hopefully unavoidable) case, should we get an empty list? Or will the error be handled?

cyrusdobbs: I don't know if it'd been decided as such but this follows the existing pattern on these endpoints. I guess it doesn't really matter as long as the error is handled. I would imagine we'd want to handle it in the same way for all of these filters. The bit of work for calling this endpoint from the front-end hasn't been done yet so I'll put a note on the goal and related tickets.

nsnape (approved):
LGTM

PR #77, opened 2023-06-06
[BAU] working towards full etl

cyrusdobbs:
### Change description
Changes to enable full ETL of towns_fund

### How to test


core/controllers/ingest.py
```diff
@@ -10,28 +10,49 @@
...
+    If source_type is specified, the data will be put through an etl pipeline, prior
+    to validation.
+    If the data fails validation, a `ValidationError` is raised. Otherwise, the data is cleaned and ingested into a
+    database.
 
     :return: A tuple containing a string message and an integer HTTP status code.
     :raises ValidationError: If the data fails validation against the specified schema.
     """
     excel_file = connexion.request.files["excel_file"]  # required
+    source_type = connexion.request.form.get("source_type")  # required
```
gidsg, line 39: Are these not passed through as method args rather than having to fetch them from connexion? e.g. as per: https://connexion.readthedocs.io/en/stable/request.html#automatic-parameter-handling ?

core/controllers/ingest.py
```diff
@@ -10,28 +10,49 @@
...
+    specified schema, and validates it.
+    If source_type is specified, the data will be put through an etl pipeline, prior
+    to validation.
+    If the data fails validation, a `ValidationError` is raised. Otherwise, the data is cleaned and ingested into a
+    database.
 
     :return: A tuple containing a string message and an integer HTTP status code.
     :raises ValidationError: If the data fails validation against the specified schema.
     """
     excel_file = connexion.request.files["excel_file"]  # required
```
gidsg, line 38: Can we get this from a function parameter too?

cyrusdobbs: ah yeah nice! You can just put `excel_file` in the function params.

core/controllers/ingest.py
```diff
@@ -95,3 +119,17 @@ def populate_db(workbook: dict[str, pd.DataFrame], mappings: tuple[DataMapping])
...
+
+    :param excel_file: The Excel file to save.
+    :param submission_id: The ID of the submission to be updated.
+    """
+    submission = Submission.query.filter_by(submission_id=submission_id).first()
+    submission.submission_filename = excel_file.filename
+    excel_file.stream.seek(0)
+    submission.submission_file = excel_file.stream.read()
+    db.session.add(submission)
+    db.session.commit()
```
gidsg, line 135: Are we committing the whole binary or something else?

tests/controller_tests/test_ingest.py
```diff
@@ -138,3 +142,21 @@ def test_extract_data_extracts_from_multiple_sheets(example_data_model_file):
...
+    )
+    db.session.add(sub)
+
+    filename = "example.xlsx"
+    filebytes = b"example file contents"
+    file = FileStorage(BytesIO(filebytes), filename=filename)
+
+    save_submission_file(file, submission_id=sub.submission_id)
+    assert Submission.query.first().submission_filename == filename
+    assert Submission.query.first().submission_file == filebytes
```
gidsg, line 162: Can we persist this as a dataframe or some picked python format in a database Text field or similar? Not sure if it's a great idea to store raw files directly in the db?

cyrusdobbs: A potential problem with that approach is that if we pickle the DataFrame and the dependencies go out of sync we could run into problems unpickling later on. I have experienced this problem with pickling before..

core/extraction/towns_fund.py
```diff
@@ -65,11 +66,47 @@ def ingest_towns_fund_data(df_ingest: pd.DataFrame) -> Tuple[Dict[str, pd.DataFr
         project_lookup,
         programme_id,
     )
+    towns_fund_extracted["Submission_Ref"] = pd.DataFrame(
+        data=[
+            {
+                "Submission ID": 1,
```
nsnape, line 72: Most of this is now done in a separate PR, is this just to "hack" the data in in the mean time?

core/extraction/towns_fund.py
```diff
@@ -65,11 +66,47 @@ def ingest_towns_fund_data(df_ingest: pd.DataFrame) -> Tuple[Dict[str, pd.DataFr
...
+
+    add_submission_ids(towns_fund_extracted)
+
     reporting_period = df_ingest["1 - Start Here"].iloc[4, 1]
 
     return towns_fund_extracted, reporting_period
 
 
+def add_submission_ids(dfs):
+    tables_need_sub_ids = [
```
nsnape, line 89: We might not be able to do it like this at extraction time, as we'll only create the submission ID when we load into the DB (as it needs to check the DB for the previous one). We need to think about transaction control to prevent race conditions

nsnape: Unless you propose using the dummy id for linking the entities, then updating the submission_id afterwards?

nsnape (changes requested):

tests/controller_tests/test_ingest.py
gidsg: OK, let's leave as binary for now, but add a TODO to migrate to S3.

core/extraction/towns_fund.py
cyrusdobbs: This is just a hacky solution to single ingest - wasn't proposing anything. That being said, that doesn't sound like a bad idea on the face of it?

tests/controller_tests/test_ingest.py
cyrusdobbs: Added a TODO

core/extraction/towns_fund.py
cyrusdobbs: this is now properly implemented by a prior PR

core/extraction/towns_fund.py
cyrusdobbs: this has all been removed now

core/db/entities.py
```diff
@@ -41,7 +41,8 @@ class Submission(BaseModel):
 
     __tablename__ = "submission_dim"
 
-    submission_id = sqla.Column(sqla.String(), nullable=False, unique=True)
+    # TODO: Change submission ID from an int value to a string that includes programme or fund type?
```
nsnape, line 44: can remove this todo now?

core/controllers/ingest.py
```diff
@@ -45,6 +65,9 @@ def ingest():
     clean_data(workbook)
     populate_db(workbook, INGEST_MAPPINGS)
 
+    submission_id = workbook["Submission_Ref"].iloc[0]["Submission ID"]
```
nsnape, line 68: does this now need fetching from DB (or fetching as param from when it was created)?

core/db/entities.py
```diff
@@ -41,7 +41,8 @@ class Submission(BaseModel):
...
+    submission_id = sqla.Column(sqla.Integer(), nullable=False, unique=True)
```
nsnape, line 45: Should this now be string again?

core/extraction/towns_fund.py
```diff
@@ -118,6 +118,28 @@ def extract_submission_details(submission_period: str) -> pd.DataFrame:
     return df_submission
 
 
+def add_submission_ids(dfs):
```
nsnape, line 121: sorry, I might be getting confused - do we still need this?

nsnape (changes requested):

core/db/entities.py
cyrusdobbs: old rebase was messing with the branch - should be fixed now

core/controllers/ingest.py
cyrusdobbs: It's inserted into the dataframe so can grab it from here. We could return it from populate_db instead?

core/db/entities.py
cyrusdobbs: symptom of an old rebase - dropped the commit now and this is gone

core/controllers/ingest.py
nsnape: nah, no preference, that's fine

nsnape (approved):

PR #81, opened 2023-06-07
[BAU] submission ID added after validation

cyrusdobbs:
### Change description
Modified the ingest flow to get the submission ID from the DB, rather than expecting it from the submission data.

### How to test


core/controllers/ingest.py
```diff
@@ -82,6 +83,19 @@ def clean_data(workbook: dict[str, pd.DataFrame]) -> None:
         worksheet.replace({np.nan: None}, inplace=True)  # replaces np.NAT with None
 
 
+def next_submission_id() -> int:
+    """Get the next submission ID by incrementing the last in the DB.
+
+    :return: The next submission ID.
+    """
+    latest_submission = Submission.query.order_by(Submission.submission_id.desc()).first()
```
nsnape, line 91: So, I probably should have told you this, but we were planning on a set specification for the Submission ID. It will be a string in the format "S-R03-1", "S-R03-2"  etc with the Rxx number referring to the round (taken from submission DataFrame). then we can easily, and in human readable format see where the submission was made (and makes it easier to ingest historical data)

core/db/entities.py
```diff
@@ -41,7 +41,8 @@ class Submission(BaseModel):
 
     __tablename__ = "submission_dim"
 
-    submission_id = sqla.Column(sqla.String(), nullable=False, unique=True)
+    # TODO: Change submission ID from an int value to a string that includes programme or fund type?
+    submission_id = sqla.Column(sqla.Integer(), nullable=False, unique=True)
```
nsnape, line 45: see previous comment. We should have enough info to build this. Doesn't need to reference programme or project, since they will actually reference this "parent" submission ID. Fund type can always be looked up via db queries to Programe_Ref.fund_type, don't need it in the submission id

nsnape (changes requested):
Do these changes I requested make sense? Are the appropriate for the scope of this ticket?

core/controllers/ingest.py
cyrusdobbs: Updated to this format of `submission_id`

core/controllers/ingest.py
```diff
@@ -82,6 +83,20 @@ def clean_data(workbook: dict[str, pd.DataFrame]) -> None:
...
+def next_submission_id(reporting_round: int) -> str:
+    """Get the next submission ID by incrementing the last in the DB.
+
+    :return: The next submission ID.
+    """
+    submissions = Submission.query.filter_by(reporting_round=reporting_round).all()
+    if not submissions:
+        return SUBMISSION_ID_FORMAT.format(reporting_round, 1)  # the first submission
+
+    latest_submission = sorted(submissions, key=lambda x: x.submission_number)[0]
```
nsnape, line 95: I think you should be able to order the data in the original query, with something like `Submission.query.filter_by(reporting_round=reporting_round).order_by(Submission.submission_number).all()`

nsnape: Or even better, sort and just return the top value, something like:
`Submission.query.filter_by(reporting_round=reporting_round).order_by(desc(Submission.submission_number)).first()`

core/controllers/ingest.py
```diff
@@ -90,8 +105,12 @@ def populate_db(workbook: dict[str, pd.DataFrame], mappings: tuple[DataMapping])
                      the workbook to the database.
     :return: None
     """
+    reporting_round = workbook["Submission_Ref"].iloc[0]["Reporting Round"]
```
nsnape, line 108: By co-incidence I was reading about these lookups earlier, apparently this is slightly faster if you select by column first, as the slice doesn't mix datatypes. Not that it really mattes here for a 1 line df  :D but thought you might like to know. So something like:
`workbook["Submission_Ref"]["Reporting Round"].iloc[0]`

nsnape (changes requested):
A couple comments, but I think this is a really tidy solution

core/controllers/ingest.py
cyrusdobbs: I tried this originally but it doesn't look like you can query on hybrid parameters like this. I imagine because the query is executed at SQL level but the hybrid param is at Python/ORM level? I may have been implementing it slightly wrong though so will give it another go

core/controllers/ingest.py
cyrusdobbs: ah good to know! looks slightly more readable too

core/controllers/ingest.py
cyrusdobbs: yeah this errors....

nsnape: ah ok, shame

nsnape: as discussed, try ordering on Submission_id itself? should work on a string

nsnape (changes requested):

nsnape (approved):

PR #86, opened 2023-06-08
[#PAB-391] Round One Load

Joel-Pearce:
- Load for Round 1 ingest
- Also includes minor changes to Round 1 transformation

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,6 @@
 FROM python:3.11-slim-bullseye
 ENV FLASK_ENV=dev
-ENV GUNICORN_CMD_ARGS="--timeout 60 --workers 3"
+ENV GUNICORN_CMD_ARGS="--timeout 300 --workers 3"
 WORKDIR /app
 COPY requirements.txt requirements.txt
 RUN python3 -m pip install --upgrade pip && pip install -r requirements.txt
--- a/core/controllers/ingest.py
+++ b/core/controllers/ingest.py
@@ -1,4 +1,5 @@
 """Provides a controller for spreadsheet ingestion."""
+import copy
 from io import BytesIO
 
 import numpy as np
@@ -16,11 +17,13 @@ from core.db.entities import Organisation, Programme, Project, Submission
 
 # isort: on
 from core.errors import ValidationError
+from core.extraction.round_one import ingest_round_1_data_towns_fund
 from core.extraction.towns_fund import ingest_towns_fund_data
 from core.validation.casting import cast_to_schema
 from core.validation.validate import validate
 
 ETL_PIPELINES = {
+    "tf_round_one": ingest_round_1_data_towns_fund,
     "tf_round_three": ingest_towns_fund_data,
 }
 
@@ -48,7 +51,11 @@ def ingest(body, excel_file):
         etl_pipeline = ETL_PIPELINES[source_type]
         workbook = etl_pipeline(workbook)
 
-    schema = current_app.config["VALIDATION_SCHEMA"]
+    schema = (
+        modify_schema(current_app.config["VALIDATION_SCHEMA"])
+        if source_type == "tf_round_one"
+        else current_app.config["VALIDATION_SCHEMA"]
+    )
     cast_to_schema(workbook, schema)
     validation_failures = validate(workbook, schema)
 
@@ -56,7 +63,10 @@ def ingest(body, excel_file):
         raise ValidationError(validation_failures=validation_failures)
 
     clean_data(workbook)
-    populate_db(workbook, INGEST_MAPPINGS)
+    if source_type == "tf_round_one":
+        populate_db_round_one(workbook, INGEST_MAPPINGS)
+    else:
+        populate_db(workbook, INGEST_MAPPINGS)
 
     submission_id = workbook["Submission_Ref"]["Submission ID"].iloc[0]
     save_submission_file(excel_file, submission_id)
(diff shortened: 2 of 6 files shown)


nsnape (changes requested):
Can you update PR title to describe everything in it, and to give context (ir Round 1 transformation of....)?

nsnape (changes requested):

myenv/bin/Activate.ps1
```diff
@@ -0,0 +1,247 @@
+<#
```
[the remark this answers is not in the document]
nsnape: You also need to get it removed in the commit so the file  doesn't show up here

tests/controller_tests/test_integration.py
```diff
@@ -1,6 +1,8 @@
 import pandas as pd
 import pytest
 
+from core.controllers.ingest import ingest
+from core.controllers.ingest import ingest
```
nsnape, line 5: double import?

nsnape (commented):
couple more comments

.gitignore
```diff
@@ -140,3 +140,11 @@ Round 1 Reporting - TF_aggregated_data_23.09.2022.xlsx
...
 # output files
 data_store_ERD.png
+ROUND_1_DATA_INGESTION.xlsx
+Data Model v3.7.xlsx
+.~lock.Round 1 Reporting - TF_aggregated_data_23.09.2022.xlsx#
+.~lock.ROUND_1_DATA_INGESTION.xlsx#
+~$Post_transform_EXAMPLE_round_one_data.xlsx
+
+# environments
+myenv/
```
gidsg, line 150: Can we revert this stuff? You can put it in your local gitignore global if you want.

core/controllers/ingest.py
```diff
@@ -48,15 +51,21 @@ def ingest(body, excel_file):
...
     validation_failures = validate(workbook, schema)
 
     if validation_failures:
         raise ValidationError(validation_failures=validation_failures)
 
     clean_data(workbook)
-    populate_db(workbook, INGEST_MAPPINGS)
+    populate_db_round_one(workbook, INGEST_MAPPINGS) if source_type == "tf_round_one" else populate_db(
+        workbook, INGEST_MAPPINGS
+    )
```
gidsg, line 68: Can we maybe use a case match for this? I guess this approach will get quite messy once we have multiple funds/schemas.

core/controllers/ingest.py
```diff
@@ -199,6 +208,80 @@ def populate_db(workbook: dict[str, pd.DataFrame], mappings: tuple[DataMapping])
     db.session.commit()
 
 
+def populate_db_round_one(workbook: dict[str, pd.DataFrame], mappings: tuple[DataMapping]) -> None:
```
gidsg, line 211: Should we give this a name to indicate it's specific to Towns Fund (as other funds may have a Round One in future?

tests/controller_tests/test_integration.py
```diff
@@ -47,6 +49,15 @@ def test_ingest_round_two_historical():
...
+
+
+# TODO: Remove / update this test once ingest connected into main work-flow
+@pytest.mark.skip(reason="currently this is just a pytest/pycharm debug entrypoint for ingest work")
+def test_ingest_round_one_historical_end_to_end():
+    # assumes that Round 1 Reporting data has had the tab 'Place Identifiers' added from EXAMPLE_TF_REPORTING_TEMPLATE
+    towns_fund_data = "Round 1 Reporting - TF_aggregated_data_23.09.2022.xlsx"
+    body = {"source_type": "tf_round_one"}
+
+    ingest(body, towns_fund_data)
```
gidsg, line 63: Can we revert this?

Dockerfile
```diff
@@ -1,6 +1,6 @@
 FROM python:3.11-slim-bullseye
 ENV FLASK_ENV=dev
-ENV GUNICORN_CMD_ARGS="--timeout 60 --workers 3"
+ENV GUNICORN_CMD_ARGS="--timeout 300 --workers 3"
```
cyrusdobbs, line 3: I made this change to fix an issue in test yesterday where the worker timed out during ingest. We should probably add a TODO though to check if this is a suitable long term solution to the problem

core/controllers/ingest.py
```diff
@@ -214,3 +304,46 @@ def save_submission_file(excel_file, submission_id):
...
     db.session.commit()
+
+
+def modify_schema(schema):
+    """
+    Modifies the schema by removing specified tables and adding "Submission ID" as the first column to specified tables.
+
+    :param schema: The original schema dictionary.
+    :return: The updated schema dictionary.
+    """
```
cyrusdobbs, line 315: I know I suggested this way of doing it but I've thought on it and I actually think it's more maintainable and less error prone to have 2 separate verbose validation schemas. 

- The main reason is that if we make changes to _the_ validation schema in the current setup - it may have unintended affects on R1. 
- It'll also be clearer what the validation logic is for each round if we give round 1 it's own verbose schema

cyrusdobbs: Starting with the current schema and just making these changes shouldn't take very long so probably worth doing IMO

tests/controller_tests/conftest.py
```diff
@@ -13,3 +13,10 @@ def example_data_model_file() -> BinaryIO:
     """An example spreadsheet in towns-fund schema format."""
     with open(DefaultConfig.EXAMPLE_DATA_MODEL_PATH, "rb") as file:
         yield file
+
+
+@pytest.fixture(scope="function")
+def example_round_one_file() -> BinaryIO:
+    """An example spreadsheet in Round One format."""
+    with open(DefaultConfig.EXAMPLE_ROUND_ONE_PATH, "rb") as file:
+        yield file
```
cyrusdobbs, line 22: Doesn't look like this is being used?

cyrusdobbs (changes requested):

cyrusdobbs (approved):

PR #90, opened 2023-06-09
[#PAB-332] Map Outcomes/Outputs to Categories

nsnape:
https://trello.com/c/97RfNJ9h

### Change description
Categories have been provided, now mapped to outcomes/outputs at ingest time.


nsnape:
> All looks good. I've tested it by ingesting a round three spreadsheet and looking at DB contents. All are mapped and look fine.
> 
> Only thing is, there is an Outcome like this:
> 
> ```
> {
>      "Outcome_Name": "fgjsrj",
>      "Outcome_Category": "Custom"
>  },
> ```
> 
> Which I don't really understand if it's from the real example spreadsheet?

Probably I accidentally saved the spreadsheet after messing around with rows. But it's a good example, we want non-mapped ones to be "custom"

cyrusdobbs:
Just had a thought.. do we want to make `outcome_category` non-nullable now?

nsnape:
> Just had a thought.. do we want to make `outcome_category` non-nullable now?

Good shout, I'll add it here before merging

PR #97, opened 2023-06-12
PAB-400: add migration framework and initial migration

gidsg:
### Change description
Add migration framework and initial migration 

### How to test
Steps in README should work locally
**Note** will need to be merged alongside sister PR


db/migrations/README
cyrusdobbs, line 1: Does this want to be `README.txt` or `README.md`?

gidsg: this is just autogenerated by Alembic so would rather leave as-is.

cyrusdobbs (changes requested):
Is this going to break local dev when using SQLite?

gidsg:
@cyrusdobbs I don't believe it breaks local development using sqlite, but feel free to check-out the branch and verify for yourself?

cyrusdobbs:
> @cyrusdobbs I don't believe it breaks local development using sqlite, but feel free to check-out the branch and verify for yourself?

It did break because `db.create_all()` was never called so the tables weren't built. I've added:
```
        with flask_app.app_context():
            db.create_all()
```
to app.py that's called if sqlite is in the DB URI

cyrusdobbs (approved):
LGTM. Maybe just give it a check to see it's the same version you expect as I did pull and rebase. Should be fine but just to be sure.

PR #99, opened 2023-06-12
BAU_ FIX: blank postcode type errors

nsnape:
### Change description

If user leaves postcode field blank transform gives a weird error when regex tries to run on np.nan (as it's from pandas). 

Handle this case and return empty list.

(NOTE: postcode aka locations field should not actually be nullable on DB, but that's a separate validation issue)

### DIFF
--- a/core/util.py
+++ b/core/util.py
@@ -1,5 +1,6 @@
 import re
 
+import numpy as np
 from flask_sqlalchemy.model import Model
 
 from core.const import POSTCODE_AREA_TO_ITL1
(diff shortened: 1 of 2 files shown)


core/util.py
```diff
@@ -9,13 +11,16 @@
 POSTCODE_REGEX = r"[A-Z]{1,2}[0-9][A-Z0-9]? ?[0-9][A-Z]{2}"
 
 
-def extract_postcodes(s: str) -> list[str]:
+def extract_postcodes(s: Union[str, float]) -> List[str]:
```
cyrusdobbs, line 14: Minor point but Python 3.11 introduced a new (and preferred notation for typing) - `Union[str, float]` can now be `str | float`

and also we don't need to use `List` any more - can just do `list[str]`

cyrusdobbs (changes requested):
Nice

core/util.py
cyrusdobbs: seems really insignificant in the scheme of today but the change you made is slightly wrong.. `Union[str | float]` should be just `str | float`

cyrusdobbs (approved):
LGTM

PR #112, opened 2023-06-13
[#PAB-451] filter download results by region

cyrusdobbs:
https://trello.com/c/q9Doffni/451-backend-implement-itl-region-filter-in-download

### Change description
Filter download results by region

### How to test


core/const.py
```diff
@@ -393,7 +393,7 @@ class YesNoEnum(StrEnum):
     "Stavely": "STV",
     "Sutton Town Centre High Street": "SUT",
 }
-FUND_ID_TO_NAME = {"FHSF": "High Street Fund", "TD": "Towns Fund"}
+FUND_ID_TO_NAME = {"HS": "High Street Fund", "FHSF": "High Street Fund", "TD": "Towns Fund"}
```
nsnape, line 396: according to the lookups ingest pipeline does, the options are 

```
{
        "Town_Deal": "TD",
        "Future_High_Street_Fund": "HS",
    }
```
Should these match up, and is the TD one like an "all" ?

nsnape, line 396: Could you add a comment to explain what the const dict actually is?

nsnape (changes requested):

core/const.py
cyrusdobbs: Ah - I think yours is correct. We also don't need "FSHS" after looking at the real data. We had HS as FSHS in our test data but I think that's wrong. Going to change that in the post-transform test ss.

nsnape (approved):

PR #125, opened 2023-06-14
BAU_ FIX: Handle blank/invalid Fund Type

nsnape:
### Change description
The ingest had no way of dealing with a form that had either a null or a "< Select >" set for Fund Type.

This is required in order to do many things in the ingest, including recreating the project code. (Also the form would likely have no data in any of the other tabs). Handle the case where none is entered, so it can be caught at validation.

Small refactor to minor misuse of try/except

### How to test


cyrusdobbs (approved):
LGTM

PR #126, opened 2023-06-15
BAU_Updating bash script

aaronwilliamsv1:
### Change description

Updating ssm bash script to be a little more user friendly

+semver: minor

- [ N/A ] Unit tests and other appropriate tests added or updated
- [ N/A ] README and other documentation has been updated / added (if needed)
- [ X ] Commit messages are meaningful and follow good commit message guidelines (e.g. "FS-XXXX: Add margin to nav items preventing overlapping of logo")

### DIFF
--- a/scripts/ssm-conn.sh
+++ b/scripts/ssm-conn.sh
@@ -2,49 +2,88 @@
 
 usage () {
     echo "Connects a container to a local port, for example to allow you to 'curl' requests from your local machine to the instance."
-    echo "Usage: $(basename ${0}) local_port_number [port_number] [cluster_name] [service_name] [task_name] [container_name]"
+    echo "Usage: $(basename ${0}) local_port_number -p [port_number] -c [cluster_name] -s [service_name] -t [task_name] -d [docker_container_name]"
     exit 1
 }
 
-if [[ "${#}" -lt 1 ]]; then
+while [ $OPTIND -le "$#" ]
+do
+    if getopts p:c:s:t:d: opt
+    then
+        case $opt in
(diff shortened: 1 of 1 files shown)


scripts/ssm-conn.sh
```diff
@@ -2,49 +2,88 @@
...
-else
-    task_name=${5}
 fi
 
-if [ -z "${6}" ]; then
-    container_name='data-store'
-else
-    container_name=${6}
+if [ -z "${docker_container_name}" ]; then
+    docker_container_name='data-store'
```
gidsg, line 57: Can we just call this 'container_name' docker doesn't really make sense once we're on AWS?

aaronwilliamsv1: I've called it 'docker container name' because that matches with the -d parameter pass. I can't use -c twice, and changing -d to match with docker container name seemed like the better option than trying to change cluster

Can keep it as container_name here in the code, but then the -d doesn't make much sense. Happy to accept suggestions!

gidsg (approved):

aaronwilliamsv1:
> ### Change description
> Updating ssm bash script to be a little more user friendly
> 
> +semver: minor
> 
> * [ N/A ] Unit tests and other appropriate tests added or updated

In theory could add shellcheck as a pre-commit hook https://github.com/kward/shunit2/blob/master/.githooks/pre-commit.shellcheck to test the bash script. However, it's not complicated enough to require it, in my opinion, and a more 'friendly' method should be developed instead, to allow anyone with permissions to do so

PR #129, opened 2023-06-15
[BAU_FIX] r3 ingest support for projects with colons in the name

cyrusdobbs:
### Change description
Ingest couldn't match project names to ID's if the project name contained a colon ":", resulting in non-nullable ID failures during validation. This change manages this scenario.

### How to test
This can be tested by attempting to ingest the `TF_Reporting_Template_(v3.0)_-_TD_-_CASTLEFORD_-_080623.xlsx` ingest ss. This was previously returning lots of validation errors but now passes and ingests.

### DIFF
--- a/core/extraction/towns_fund.py
+++ b/core/extraction/towns_fund.py
@@ -437,7 +437,7 @@ def extract_funding_comments(df_input: pd.DataFrame, project_lookup: dict) -> pd
 
     for idx in range(len(project_lookup)):
         line_idx = 28 * idx
-        current_project = df_input.iloc[line_idx, 0].split(": ")[1]
+        current_project = ": ".join(df_input.iloc[line_idx, 0].split(": ")[1:])  # omit the "Project X: " prefix
         comment = df_input.iloc[line_idx + 26, 0]
         fund_row = pd.DataFrame([[current_project, comment]], columns=header)
         df_fund_comments = df_fund_comments.append(fund_row)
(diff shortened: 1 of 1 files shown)


core/extraction/towns_fund.py
```diff
@@ -437,7 +437,7 @@ def extract_funding_comments(df_input: pd.DataFrame, project_lookup: dict) -> pd
 
     for idx in range(len(project_lookup)):
         line_idx = 28 * idx
-        current_project = df_input.iloc[line_idx, 0].split(": ")[1]
+        current_project = ": ".join(df_input.iloc[line_idx, 0].split(": ")[1:])  # omit the "Project X: " prefix
```
gidsg, line 440: is this testable?

gidsg: Is `:` treated as a special character by the library or something? could we encode it differently or do a regex on it? Removing it like this may have unintended consequences.

cyrusdobbs: none of this is unit tested.. it's a bigger job to think about than this fix I think

cyrusdobbs: It's because the cell is always 
```
Project X: A city or something
```
but this SS was:
```
Project X: A city: something else
```

so it was originally just grabbing the "A city" bit, without the end, and that didnt match the project ID lookup table so it all fell over. This should work regardless of how many colons you have it in now

gidsg: Could we extract the logic to a function and unit test that?

cyrusdobbs: I could extract that line into it's own function and unit test that individual line's logic if you wanted but as I said, I think its small fry in comparison to all this un(unit)tested logic

gidsg (approved):

PR #133, opened 2023-06-15
[#PAB-457]: add initial pre transformation validation

Joel-Pearce:
Added feature to validate file before transformations occur. 

These checks are intended to ensure the following:

- Form version must be: “Town Deals and Future High Streets Fund Reporting Template (v3.0)” → Sheet 1A, column B8
 - Reporting period MUST be round 3. Date input string can be mapped to rounds using the dict in core.extraction.towns_fund.extract_submission_details. It would be worth refactoring the code to have this lookup dict in constants folder since it’s being re-used.
- There must be a valid answer provided for sheet 2, Section A, Question A1 (fund type) → if left blank or “< Select >” then ingest will break.
- As previous point but with A2 (place name)

### How to test

For manual testing, the user must attempt to ingest both a valid and invalid file in order to catch the errors above. These checks are only for Round 3 data, and will not apply to other rounds. Therefore, test using Round 3 data.

### DIFF
--- a/core/extraction/towns_fund.py
+++ b/core/extraction/towns_fund.py
@@ -8,7 +8,10 @@ import numpy as np
 import pandas as pd
 from pandas.tseries.offsets import MonthEnd
 
-from core.const import OUTCOME_CATEGORIES, OUTPUT_CATEGORIES, FundTypeIdEnum
+# isort: off
+from core.const import OUTCOME_CATEGORIES, OUTPUT_CATEGORIES, REPORTING_PERIOD_DICT, FundTypeIdEnum
+
+# isort: on
 from core.extraction.utils import convert_financial_halves, drop_empty_rows
 from core.util import extract_postcodes
 
@@ -87,17 +90,7 @@ def extract_submission_details(submission_period: str) -> pd.DataFrame:
     first_period = "1 April 2019"
     # TODO: Review this - next round is actually Round 4. If the wrong round selected on form, it could wipe data
     #  for the same programme in a different round - we should probably hard-code for now.
-    for period, reporting_round in {
-        "2019/20 to 31 March 2022": 1,
-        "1 April 2022 to 30 September 2022": 2,
-        "1 October 2022 to 31 March 2023": 3,
-        "1 April 2023 to 30 September 2023": 4,
-        "1 October 2023 to 31 March 2024": 5,
-        "1 April 2024 to 30 September 2024": 6,
-        "1 October 2024 to 31 March 2025": 7,
-        "1 April 2025 to 30 September 2025": 8,
-        "1 October 2025 to 31 March 2026": 9,
-    }.items():
+    for period, reporting_round in REPORTING_PERIOD_DICT.items():
         start_str, end_str = period.split(" to ")
 
         # assuming start date of 1st round is 1st April 2019, otherwise extract from string
--- a/core/validation/failures.py
+++ b/core/validation/failures.py
@@ -41,7 +41,7 @@ class EmptySheetFailure(ValidationFailure):
 
     def __str__(self):
         """Method to get the string representation of the empty sheet failure."""
-        return f'Empty Sheets Failure: The sheet named "{self.empty_sheet}" contain no ' "data."
+        return f'Empty Sheets Failure: The sheet named "{self.empty_sheet}" contains no ' "data."
 
 
 @dataclass
@@ -153,4 +153,50 @@ class NonNullableConstraintFailure(ValidationFailure):
         return (
             f'Non-nullable Constraint Failure: Sheet "{self.sheet}" Column "{self.column}" '
             f"is non-nullable but contains a null value(s)."
+        )
+
+
+@dataclass
+class PreTransformationFailure(ValidationFailure):
+    """Class representing a pre-transformation failure."""
+
+    value_descriptor: str
+    entered_value: str
+    expected_values: set
+
+    def __str__(self):
+        """
+        Method to get the string representation of the pre-transformation failure.
+        """
+        return (
+            f"Pre-transformation Failure: The workbook failed a pre-transformation check for {self.value_descriptor} "
+            f'where the entered value "{self.entered_value}" '
+            f'was outside of the expected values [{", ".join(self.expected_values)}].'
+        )
+
+
+@dataclass
+class NoInputFailure(ValidationFailure):
+    """Class representing a no input failure."""
+
+    value_descriptor: str
+
+    def __str__(self):
+        """
+        Method to get the string representation of the no input failure.
+        """
+        return f"No Input Failure: Expected an input value for {self.value_descriptor}"
+
+
+@dataclass
+class InvalidSheetFailure(ValidationFailure):
+    """Class representing an invalid sheet failure."""
+
+    invalid_sheet: str
+
+    def __str__(self):
+        """Method to get the string representation of the empty sheet failure."""
+        return (
+            f"Invalid Sheets Failure: The sheet named {self.invalid_sheet} is invalid "
+            f"as it is missing expected values"
         )
--- a/core/validation/initial_check.py
+++ b/core/validation/initial_check.py
@@ -1,0 +1,120 @@
+from typing import Any
+
+import numpy as np
+import pandas as pd
+
+import core.validation.failures as vf
+from core.const import EXPECTED_ROUND_THREE_SHEETS
+
+
+def extract_round_three_submission_details(
+    workbook: dict[str, pd.DataFrame]
+) -> dict[str, list[str]] | dict[str, dict[Any, Any]]:
+    """
+    Extract submission details for round three from the given workbook.
+
+    :param workbook: A dictionary where keys are sheet names and values are pandas DataFrames.
+    :return: A dictionary containing the extracted submission details.
+    """
+    details_dict = {"ValueChecks": {}, "NullChecks": {}}
+
+    missing_sheets = check_missing_sheets(EXPECTED_ROUND_THREE_SHEETS, workbook)
+
+    if missing_sheets:
+        return {"Missing Sheets": missing_sheets}
+
+    invalid_sheets = []
+
+    sheet_a1 = workbook.get("1 - Start Here")
+    sheet_a2 = workbook.get("2 - Project Admin")
+    try:
+        details_dict["ValueChecks"]["Form Version"] = (
+            sheet_a1.iloc[6][1],
+            {"Town Deals and Future High Streets Fund Reporting Template (v3.0)"},
+        )
+        details_dict["ValueChecks"]["Reporting Period"] = (sheet_a1.iloc[4][1], {"1 October 2022 to 31 March 2023"})
+    except IndexError:
+        invalid_sheets.append("1 - Start Here")
+
+    try:
+        details_dict["ValueChecks"]["Fund Type"] = (sheet_a2.iloc[5][4], {"Town_Deal", "Future_High_Street_Fund"})
+        details_dict["NullChecks"]["Place Name"] = sheet_a2.iloc[6][4]
+    except IndexError:
+        invalid_sheets.append("2 - Project Admin")
+
+    if invalid_sheets:
+        return {"Invalid Sheets": missing_sheets}
+
+    return details_dict
+
+
+def pre_transformation_check(submission_details: dict[str, dict[str, dict]]) -> list[vf.ValidationFailure]:
+    """
+    Perform pre-transformation checks on the given workbook.
+
+    :param submission_details: A dictionary containing submission details.
+    :return: A list of validation failures encountered during pre-transformation checks.
+    """
+
+    if missing_sheets := submission_details.get("Missing Sheets"):
+        return [vf.EmptySheetFailure(empty_sheet) for empty_sheet in missing_sheets]
+
+    if invalid_sheets := submission_details.get("Invalid Sheets"):
+        return [vf.InvalidSheetFailure(invalid_sheet) for invalid_sheet in invalid_sheets]
+
+    failures = []
+
+    for value_descriptor, (entered_value, expected_values) in submission_details["ValueChecks"].items():
+        if failure := check_values(value_descriptor, entered_value, expected_values):
+            failures.append(failure)
+
+    for value_descriptor, value in submission_details["NullChecks"].items():
+        if failure := check_nulls(value_descriptor, value):
+            failures.append(failure)
+
+    return failures
+
+
+def check_values(value_descriptor: str, entered_value: str, expected_values: set) -> vf.PreTransformationFailure | None:
+    """
+    Check the form input for pre-transformation failures.
+
+    :param value_descriptor: A string describing the form input value.
+    :param entered_value: A string containing the entered value
+    :param expected_values: A set containing the set of expected values.
+    :return: A ValidationFailure object representing the failure, if any.
+    """
+
+    if entered_value not in expected_values:
+        return vf.PreTransformationFailure(
+            value_descriptor=value_descriptor, entered_value=entered_value, expected_values=expected_values
+        )
+
+
+def check_nulls(value_descriptor: str, value: str) -> vf.NoInputFailure | None:
+    """
+    Check the form input for pre-transformation failures.
+
+    :param value_descriptor: A string describing the form input value.
+    :param value: A string containing the form input value.
+    :return: A ValidationFailure object representing the failure, if any.
+    """
+
+    if value in ["", np.nan]:
+        return vf.NoInputFailure(value_descriptor=value_descriptor)
+
+
+def check_missing_sheets(expected_sheets: list[str], workbook: dict[str, pd.DataFrame]) -> list[str]:
+    """
+    Check for missing sheets in the workbook.
+
+    :param expected_sheets: A list of expected sheet names.
+    :param workbook: A dictionary where keys are sheet names and values are pandas DataFrames.
+    :return: A list of missing sheets.
+    """
+    missing_sheets = []
+    for sheet in expected_sheets:
+        if workbook.get(sheet) is None or workbook.get(sheet).empty:
+            missing_sheets.append(sheet)
+
+    return missing_sheets
--- a/tests/validation_tests/test_pre_checks.py
+++ b/tests/validation_tests/test_pre_checks.py
@@ -1,0 +1,199 @@
+import pandas as pd
+import pytest
+
+import core.validation.failures as vf
+
+# isort: off
+from core.validation.initial_check import (
+    extract_round_three_submission_details,
+    pre_transformation_check,
+)
+
+
+@pytest.fixture
+def valid_workbook():
+    valid_workbook = {
+        "1 - Start Here": pd.DataFrame(
+            {
+                0: ["", "", "", "", "", "", "", "", ""],
+                1: [
+                    "",
+                    "",
+                    "",
+                    "",
+                    "1 October 2022 to 31 March 2023",
+                    "",
+                    "Town Deals and Future High Streets Fund Reporting Template (v3.0)",
+                    "",
+                    "",
+                ],
+            }
+        ),
+        "2 - Project Admin": pd.DataFrame(
+            {
+                0: ["", "", "", "", "", "", "", "", ""],
+                1: ["", "", "", "", "", "", "", "", ""],
+                2: [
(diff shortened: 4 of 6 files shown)


core/validation/initial_check.py
cyrusdobbs, line 1: I think it would be best to separate the data extraction and the validation - that way we can make the validation code spreadsheet type agnostic. This would mean we could have different data extraction steps for different spreadsheet types but the validation code can accept a standard input and be used across rounds / funds etc.

Joel-Pearce: Do you mean calling the extract_data() method after the pre-transformation file check? That's what I was looking at first, but I decided against it as I had to read the xlsx file into a dataframe anyway and it was just extra unneeded code. 

Could you give me a quick pseudocode example so I can get a better idea of what you mean if I've misunderstood?

core/validation/initial_check.py
```diff
@@ -0,0 +1,150 @@
...
+    """
+    details_dict = {}
+
+    sheet_a1 = workbook.get("1 - Start Here")
+    if sheet_a1 is not None:
+        details_dict["Form Version"] = sheet_a1.iloc[6][1]
+        details_dict["Reporting Period"] = sheet_a1.iloc[4][1]
+        details_dict["A1"] = "valid"
+    else:
+        details_dict["Form Version"] = None
```
nsnape, line 27: this feels unnecessary. First of all it's probably better not to call the else at all, and just access the expected keys with a get method. Then missing data is implied by it's absence, not entered into a dict as None values (this is better practice in general). 

But in this case we shouldn't worry about that, instead I think it'd make more sense to add a check before these that checks if these sheets exist, if not immediately cease validating and return an error (since the other code will not find anything if all is missing).

core/validation/initial_check.py
```diff
@@ -0,0 +1,150 @@
+"""Initial Pre-transformation Validation
```
nsnape, line 1: As a general observation: This code is super-specific and not really reusable at all (other rounds, different funds, different version of forms etc), since lots of the validation values are hard-coded into the functions. 

Since most of the fields this is validating are single fields, being compared against a single value, you could for the most part just re-use something like your "check_form_input" function, for all (or most) of the fields we are checking, and invoke that function with params, field_value_entered and field_value_expected. That way it's much more versatile, when we come to the next round, we can simply pass the same function a new pre-defined selection of expected field values. Does that make sense?

core/validation/initial_check.py
```diff
@@ -0,0 +1,150 @@
...
+            vf.WrongRoundFailure(
+                expected_round=3,
+                actual_round=round_value,
+            )
+        )
+
+
+def check_form_input(submission_details: dict[str, str], question: str, failures: list[vf.ValidationFailure]) -> None:
+    """
+    Check if the form input is valid.
```
nsnape, line 120: Docstring a bit vague, could you specifically mention this function is used to check any specified field does not contain one of ["", "< Select >"] ?

I think this function could potentially be re-used in other places in future, so good to have accurate and descriptive docstring

core/validation/initial_check.py
```diff
@@ -0,0 +1,150 @@
...
+                invalid_reporting_period=reporting_period,
+                valid_reporting_periods=REPORTING_PERIOD_DICT.keys(),
+            )
+        )
+
+        return
+
+    round_value = REPORTING_PERIOD_DICT[reporting_period]
+
+    if round_value != 3:
```
nsnape, line 109: my feeling is this bit should be refactored out into a separate function that is passed the expected round of "3" as a param, hard-coding it like this makes the function very specific

nsnape (dismissed):

Joel-Pearce:
@nsnape 

Thanks for the feedback, I have refactored the code to make it more reusable. "Form Version" and "Reporting Round" are now specific to extract_round_three_submission_details(). Each round can have its own extraction function that will populate these two fields in accordance with the information we would expect from those rounds.

core/validation/initial_check.py
```diff
@@ -0,0 +1,134 @@
...
+    sheet_a1 = workbook.get("1 - Start Here")
+    if sheet_a1 is None:
+        return {"a1": "1 - Start Here"}
+    details_dict["Form Version"] = sheet_a1.iloc[6][1]
+    details_dict["Reporting Period"] = sheet_a1.iloc[4][1]
+
+    sheet_a2 = workbook.get("2 - Project Admin")
+    if sheet_a2 is None:
+        return {"a2": "2 - Project Admin"}
+    details_dict["Are you filling this in for a Town Deal or Future High Street Fund?"] = sheet_a2.iloc[5][4]
```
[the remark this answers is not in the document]
cyrusdobbs: I imagine we want to validate that the Fund Type is in a given range of values - TD or HS for this fund? So may want to add those as expected values like round etc

tests/validation_tests/test_pre_checks.py
```diff
@@ -0,0 +1,186 @@
+import pandas as pd
```
nsnape, line 1: could you add a test for `pre_transformation_check `please? Everything else covered but this one not actually tested.

nsnape: and test the `file_validation_failures ` look ok when raised?

core/validation/failures.py
```diff
@@ -154,3 +154,57 @@ def __str__(self):
...
+
+
+@dataclass
+class InvalidReportingPeriodFailure(ValidationFailure):
+    """Class representing a invalid form input failure."""
+
+    invalid_reporting_period: str
+    valid_reporting_periods: str
+
+    def __str__(self):
```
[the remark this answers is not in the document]
nsnape: could you reword as I've suggested above? the "invalid" part is a little misleading as an error message. Unexpected is better (as the input they've used will be "valid", just not for this round.

core/validation/initial_check.py
```diff
@@ -0,0 +1,168 @@
...
+        failures.append(
+            vf.InvalidReportingPeriodFailure(
+                invalid_reporting_period=reporting_period,
+                valid_reporting_periods=REPORTING_PERIOD_DICT.keys(),
+            )
+        )
+
+        return
+
+    round_value = REPORTING_PERIOD_DICT[reporting_period]
```
nsnape, line 106: I think it'd be more useful to actually return the date period strings here rather than the reporting round number (since that's our concept, users may only understand in terms of the date range they put on the spreadsheet form). So could just not lookup the values from REPORTING_PERIOD_DICT for this bit, just effectively use the `reporting_period`

core/extraction/towns_fund.py
```diff
@@ -8,7 +8,8 @@
 import pandas as pd
 from pandas.tseries.offsets import MonthEnd
 
-from core.const import OUTCOME_CATEGORIES, OUTPUT_CATEGORIES, FundTypeIdEnum
+# isort: off
+from core.const import OUTCOME_CATEGORIES, OUTPUT_CATEGORIES, REPORTING_PERIOD_DICT, FundTypeIdEnum
```
nsnape, line 12: can you turn isort back on on the next line?

nsnape (commented):
Looking pretty good now. A few more things to change in the comments please (I'll not mark as request changes or else you'll be blocked from merging). Thanks

core/validation/initial_check.py
cyrusdobbs: @Joel-Pearce this hasn't change hasn't been implemented and I think it's key to what this pre validation pattern should look like. 

check_form_input should look something like this
```
def check_form_input(entered_value: str,  expected_values: str | set[str]) -> vf.PreTransformationFailure:
    if isinstance(expected_values, str):
        expected_values = {expected_values}

    if entered_value not in expected_values:
        return vf.PreTransformationFailure(
                entered_value=entered_value,
                expected_values=expected_values,
            )
        )
```
which we can then just use in a loop on the values extracted from the submission i.e. fund type, form version, place name...

core/validation/initial_check.py
cyrusdobbs: This is probably redundant if the above change is implemented

cyrusdobbs: In my mind, there's two types of validation we want to be able this module to do: 

1. is where we want to check if a value e.g. Fund Type is in a set of valid values e.g. HS or TD - the above mentioned `check_form_input` covers these cases
2. we also want to check that some values just aren't null e.g. Place Name - I think this is where this function comes in - I don't think we should be checking "< Select >" here because the extraction function can deal with that specific detail. This function should just be used to check if some values aren't null and raising a failure saying it wasn't entered

core/validation/initial_check.py
Joel-Pearce: I've refactored the code to make pre_transformation_check() round agnositic. Only the extraction function will differ per round.

core/validation/initial_check.py
Joel-Pearce: Code now checks for an expected value for a given round, performing a check to see if a relevant value has been added to catch nulls as well as mismatched values.

core/validation/initial_check.py
```diff
@@ -0,0 +1,94 @@
...
+    """
+
+    if "Missing Sheets" in submission_details:
+        return [vf.EmptySheetFailure(empty_sheet=empty_sheet) for empty_sheet in submission_details["Missing Sheets"]]
+
+    pre_transformation_failures = [
+        check_form_input(key, value)
+        for key, value in submission_details.items()
+        if check_form_input(key, value) is not None
+    ]
```
cyrusdobbs, line 57: This code will unnecessary call `check_form_input` a second time. Better to do something like this:
```
failures = []
for value_descriptor, values in submission_details.items():
    if failure := check_form_input(value_descriptor, values):
        failures.append(failure)
```

core/validation/initial_check.py
```diff
@@ -0,0 +1,94 @@
...
+    :param values: A string containing the form input values.
+    :return: A ValidationFailure object representing the failure, if any.
+    """
+    if values[0] in ["", np.nan]:
+        return vf.NoInputFailure(value_descriptor=value_descriptor)
+
+    if value_descriptor != "Place Name" and values[0] not in values[1:]:
+        return vf.PreTransformationFailure(
+            value_descriptor=value_descriptor, entered_value=values[0], expected_values=values[1:]
+        )
```
cyrusdobbs, line 78: Couple of points:

- Shouldn't combine the entered value and expected into a single list of strings like this. 
- The null check should be a separate function that is only used on a separate list of values because otherwise we have to make hard coded checks like the one for "Place Name".

I'd consider changing the submission_details to this structure:
```
submission_details = {
    "Missing Sheets": [],
    "NullChecks": {
        "Place Name": "Some Place not an Empty String",  #  this would pass the check
        "Something else": ""  # this would fail
    },
    "ValueChecks": {
        "Fund Type": ("TD", {"TD", "HS"}),  # a tuple of two values, the entered, and the set of expected
        "Something else": ("Entered", {"Expected1", "Expected2"})
    }
}
```
This means we can iterate over the null checks and call the null check function for those, and for the value checks we can call this function.

A dictionary will do for now but we'll probably want to make this an object to better define and enforce the structure at some point.

Joel-Pearce: I've changed the structure to look like this.

Was thinking we could actually make the extraction spreadsheet agnostic if the expected_values for each round were tied to an object in const, and then we could just add source_type as a param and wouldn't need to add any code for subsequent rounds, but Round 1 and Round 2 won't have any pre-transformation validations and we can't be sure about the specific cell location in subsequent rounds so I've stuck with that you've suggested.

cyrusdobbs (commented):
Looking good - just a couple of minor points

core/validation/initial_check.py
```diff
@@ -0,0 +1,107 @@
...
+    :return: A list of validation failures encountered during pre-transformation checks.
+    """
+
+    if "Missing Sheets" in submission_details:
+        return [vf.EmptySheetFailure(empty_sheet) for empty_sheet in submission_details["Missing Sheets"]]
+
+    failures = []
+    for value_descriptor, values in submission_details["ValueChecks"].items():
+        if failure := check_values(value_descriptor, values):
+            failures.append(failure)
```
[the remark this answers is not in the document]
Joel-Pearce: Amended.

core/validation/initial_check.py
```diff
@@ -0,0 +1,107 @@
...
+
+    for value_descriptor, values in submission_details["NullChecks"].items():
```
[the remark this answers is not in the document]
Joel-Pearce: Amended.

core/validation/initial_check.py
```diff
@@ -0,0 +1,107 @@
...
+def pre_transformation_check(submission_details: dict[str, list[str]]) -> list[vf.ValidationFailure]:
+    """
+    Perform pre-transformation checks on the given workbook.
+
+    :param submission_details: A dictionary containing submission details.
+    :return: A list of validation failures encountered during pre-transformation checks.
+    """
+
+    if "Missing Sheets" in submission_details:
+        return [vf.EmptySheetFailure(empty_sheet) for empty_sheet in submission_details["Missing Sheets"]]
```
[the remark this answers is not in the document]
Joel-Pearce: Amended.

core/validation/initial_check.py
```diff
@@ -0,0 +1,107 @@
...
+    sheet_a1 = workbook.get("1 - Start Here")
+    if sheet_a1 is not None:
+        details_dict["ValueChecks"]["Form Version"] = (
+            sheet_a1.iloc[6][1],
+            {"Town Deals and Future High Streets Fund Reporting Template (v3.0)"},
+        )
+        details_dict["ValueChecks"]["Reporting Period"] = (sheet_a1.iloc[4][1], {"1 October 2022 to 31 March 2023"})
+
+    sheet_a2 = workbook.get("2 - Project Admin")
+    if sheet_a2 is not None:
```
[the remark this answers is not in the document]
Joel-Pearce: It's not possible to evaluate the truth of a dataframe like this. This leaves us with if sheet_a1 is not None, or if sheet_at1 is in df.values. 

Changed for mapping.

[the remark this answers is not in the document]
cyrusdobbs: ah good point

core/validation/initial_check.py
```diff
@@ -0,0 +1,111 @@
...
+        details_dict["ValueChecks"]["Fund Type"] = (sheet_a2.iloc[5][4], {"Town_Deal", "Future_High_Street_Fund"})
+        details_dict["NullChecks"]["Place Name"] = sheet_a2.iloc[6][4]
```
cyrusdobbs, line 38: Just realised we'll get index errors if a spreadsheet that is wildly wrong tries to be ingest... e.g. df.iloc[3][1] will fall over if the df is tiny. Not sure if we want to account for this... The fact we pull named sheets out and don't run the index lookups if they don't exist will protect against this in the vast majority of cases.. 

One solution that wouldn't require much work could be to add a try except round the uses of `iloc` and if IndexError is found then just add the sheet as missing and continue.. 

What do you think? It seems a bit excessive but it is definitely a possible edge case

core/validation/initial_check.py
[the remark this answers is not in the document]
cyrusdobbs: can't see that this has changed

core/validation/initial_check.py
```diff
@@ -0,0 +1,111 @@
...
+
+    return details_dict
+
+
+def pre_transformation_check(submission_details: dict[str, list[str]]) -> list[vf.ValidationFailure]:
```
cyrusdobbs, line 43: submission details typing hint is wrong - `dict[str, dict]` might do?

core/validation/initial_check.py
```diff
@@ -0,0 +1,111 @@
...
+
+    for value_descriptor, values in submission_details["NullChecks"].items():
+        if failure := check_nulls(value_descriptor, values):
+            failures.append(failure)
+
+    return failures
+
+
+def check_values(
+    value_descriptor: str, entered_value: str, expected_values: tuple
```
cyrusdobbs, line 68: Looks like `expected_values` is a set rather than a tuple

core/validation/initial_check.py
Joel-Pearce: Amended.

core/validation/initial_check.py
Joel-Pearce: The extraction function is written under the assumption that the form will be formatted in the same way for each round, which is why we may need bespoke extraction functions for different rounds. I haven't encountered any R3 forms that don't adhere to the specific format presumed by this extraction method. 

To hedge our bets we could do something like:
```
    sheet_a1 = workbook.get("1 - Start Here")
    sheet_a2 = workbook.get("2 - Project Admin")
    try:
        details_dict["ValueChecks"]["Form Version"] = (
            sheet_a1.iloc[6][1],
            {"Town Deals and Future High Streets Fund Reporting Template (v3.0)"},
        )
        details_dict["ValueChecks"]["Reporting Period"] = (sheet_a1.iloc[4][1], {"1 October 2022 to 31 March 2023"})
        details_dict["ValueChecks"]["Fund Type"] = (sheet_a2.iloc[5][4], {"Town_Deal", "Future_High_Street_Fund"})
        details_dict["NullChecks"]["Place Name"] = sheet_a2.iloc[6][4]
    except IndexError as e:
        return details_dict
```

As the missing sheet will already be caught above, and so we merely want to return the dictionary containing this. However, this won't account for the sheet being present but with insufficient data (e.g. if it just had headers), in which case it'd try and fail to locate the specified cells.

Another way would be as you suggested:

```
missing_sheets = []

    sheet_a1 = workbook.get("1 - Start Here")
    sheet_a2 = workbook.get("2 - Project Admin")
    try:
        details_dict["ValueChecks"]["Form Version"] = (
            sheet_a1.iloc[6][1],
            {"Town Deals and Future High Streets Fund Reporting Template (v3.0)"},
        )
        details_dict["ValueChecks"]["Reporting Period"] = (sheet_a1.iloc[4][1], {"1 October 2022 to 31 March 2023"})
    except IndexError:    
        missing_sheets.append("1 - Start Here")
    
    try:
        details_dict["ValueChecks"]["Fund Type"] = (sheet_a2.iloc[5][4], {"Town_Deal", "Future_High_Street_Fund"})
        details_dict["NullChecks"]["Place Name"] = sheet_a2.iloc[6][4]
    except IndexError:    
        missing_sheets.append("2 - Project Admin")
```
In this instance, we don't check to see if sheets are missing before trying to retrieve the values from the specified cells as otherwise we'd potentially be adding the same sheet twice to the list. We no longer iterate through all of the sheets to check if present, but that was only included after an edge case in QA testing, and as post-transformation validation captures the EmptySheet phenomenon it should be alright.

Realistically, this edge case will only happen if for whatever reason the user has sheets with some data in for sheets a1 and a2, but not enough, and as the sheet is protected and cannot be altered to delete most of the cells it's very unlikely to happen, but if you think of the above two solutions works well enough and helps protect against this I am happy to implement. What do you think is the best way forward?

cyrusdobbs: The first solution could still incorrectly pass validation. If neither sheet was empty, but didn't conform to the Round 3 structure and in doing so raised an Index Error, then the details dict could be returned with either no ValueChecks or some valid ValueChecks that would then still pass validation. 

The second ensures that we raise a failure. Although I'm not sure an EmptySheet really describes the problem. It's more like an invalid sheet.

core/validation/initial_check.py
cyrusdobbs: docstring needs updating to reflect

core/validation/initial_check.py
cyrusdobbs: The second solution would fall over if there was a legitimate empty sheet now though, because we'd call iloc on None, which would raise a different exception to an Index Error. We'd also need to check if either sheet was None, and add a missing sheet in that case and skip the try except block in that case

core/validation/initial_check.py
Joel-Pearce: Amended.

core/validation/initial_check.py
Joel-Pearce: I've pushed a new solution, which separates missing sheets from invalid sheets. If there are any missing sheets, they will be returned before the code block that identifies values by cell position. This will prevent the programme calling iloc on None. Invalid sheets are then treated in the same way missing sheets are, only raising a different type of ValidationFailure.

cyrusdobbs (approved):
LGTM 💯

PR #142, opened 2023-06-21
BAU_ FIX: Upsert on Programme to include PREVIOUS rounds' programmes

nsnape:
### Change description
Previously, upsert (merge) only worked if an incoming programme had already been added in the same round. 

It now does it if one was added in this or any prior round (but not subsequent ones). Previously, this gap caused a unique error constraint

Unit test added for this specific case


cyrusdobbs (approved):
LGTM

PR #156, opened 2023-06-30
[#PAB-391] Round One Load

Joel-Pearce:
- Loading of Round One data for TF
- Slight alterations to Round One transformations

### Change description

### DIFF
--- a/core/controllers/ingest.py
+++ b/core/controllers/ingest.py
@@ -16,11 +16,13 @@ from core.db.entities import Organisation, Programme, Project, Submission
 
 # isort: on
 from core.errors import ValidationError
+from core.extraction.round_one import ingest_round_1_data_towns_fund
 from core.extraction.towns_fund import ingest_towns_fund_data
 from core.validation.casting import cast_to_schema
 from core.validation.validate import validate
 
 ETL_PIPELINES = {
+    "tf_round_one": ingest_round_1_data_towns_fund,
     "tf_round_three": ingest_towns_fund_data,
 }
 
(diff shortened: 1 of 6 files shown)


core/controllers/ingest.py
```diff
@@ -214,3 +304,46 @@ def save_submission_file(excel_file, submission_id):
...
     db.session.commit()
+
+
+def modify_schema(schema):
+    """
+    Modifies the schema by removing specified tables and adding "Submission ID" as the first column to specified tables.
+
+    :param schema: The original schema dictionary.
+    :return: The updated schema dictionary.
+    """
```
cyrusdobbs, line 315: I know I suggested this way of doing it but I've thought on it and I actually think it's more maintainable and less error prone to have 2 separate verbose validation schemas.

- The main reason is that if we make changes to the validation schema in the current setup - it may have unintended affects on R1.
- It'll also be clearer what the validation logic is for each round if we give round 1 it's own verbose schema
- Member

Starting with the current schema and just making these changes shouldn't take very long so probably worth doing IMO

cyrusdobbs (changes requested):
Looks good just one comment about the validation schema..

We'll also probably need this change from the previous PR where I added it to fix a timeout issue occurring on test due to ingest taking longer than 60 secs.. The comment is also still valid on this though regarding adding a TODO
https://github.com/communitiesuk/funding-service-design-post-award-data-store/pull/86/files#r1247641527

core/controllers/ingest.py
Joel-Pearce: I agree, and have updated the code to follow this approach.

```
schema = (
        current_app.config["ROUND_ONE_TF_VALIDATION_SCHEMA"]
        if source_type == "tf_round_one"
        else current_app.config["VALIDATION_SCHEMA"]
    )
```

This is going to quickly become messy with more rounds, but we can't define these in a dictionary outside of the function like for ETL_Pipelines as current app requires a context in which to operate. 

I am happy to refactor this - and other instances of if-statements to check r1 v. r3 - in another PR, but it could be somewhat time-consuming and is arguably beyond the scope of just loading in R1 if we're aiming to get this in today.

cyrusdobbs: Completely agree - I've been holding off commenting on things because we'll need to go back and think about this and refactor a lot of this code so it's more extendable. Will be good to get clear direction on the expected functionality going forwards and refactor accordingly.

cyrusdobbs (approved):
LGTM 👍

PR #158, opened 2023-06-30
Fix/458 Replace enums from database schema definitions with String

DonShelly:
[Ticket](https://trello.com/c/5O4rAQi0/458-remove-enums-from-database-schema-definitions)

Replaced instances of "sqla.Enum(...), ..." with "sqla.String, ..." in core/db directory


DonShelly:
[Remove Enum's from database schema definitions](https://trello.com/c/5O4rAQi0/458-remove-enums-from-database-schema-definitions)

DonShelly:
Test fails because of the addition of ""Output_Data": {"enums": {"Actual/Forecast": enums.StateEnum}},"
to the TF round 1 data schema

cyrusdobbs:
> Test fails because of the addition of ""Output_Data": {"enums": {"Actual/Forecast": enums.StateEnum}}," to the TF round 1 data schema
> 
> @cyrusdobbs or @Joel-Pearce please advise

You've tried to add "Output_Data" as a table to the round 1 schema but you haven't defined any columns so it's falling over because the schema is invalid. You don't actually want to do this though because if a table isn't already present for a schema then it's because it doesn't include data of that type - e.g. we don't ingest Outputs from Round 1 in this case, and changing the enums in the database won't change that fact. 

It's slightly outdated but [this README](https://github.com/communitiesuk/funding-service-design-post-award-data-store/tree/main/core/validation) might help to understand the schema.

core/db/entities.py
```diff
@@ -309,12 +308,10 @@ class Project(BaseModel):
...
     project_name = sqla.Column(sqla.String(), nullable=False)
     primary_intervention_theme = sqla.Column(sqla.String(), nullable=False)
-    location_multiplicity = sqla.Column(
-        sqla.Enum(const.MultiplicityEnum, name="project_location_multiplicity"), nullable=True
-    )
+    location_multiplicity = sqla.Column(sqla.String, nullable=True)
     locations = sqla.Column(sqla.String, nullable=False)
     postcodes = sqla.Column(sqla.String, nullable=True)
-    gis_provided = sqla.Column(sqla.Enum(const.YesNoEnum), nullable=True)
+    gis_provided = sqla.Column(sqla.String, nullable=True)
```
gidsg, line 314: Do we not need to generate a migration for these changes??

gidsg (changes requested):

gidsg:
...also tests are failing

core/db/entities.py
DonShelly: working on it

db/migrations/versions/006_replace_enums_with_strings.py
```diff
@@ -0,0 +1,232 @@
...
+"""
+import sqlalchemy as sa
+from alembic import op
+from sqlalchemy.dialects import postgresql
+
+# revision identifiers, used by Alembic.
+revision = "720d88c96ef6"
+down_revision = "005_relax_proj_prog_nulls"
+branch_labels = None
+depends_on = None
```
cyrusdobbs, line 16: `revision` and `Revision ID` should reflect the custom filename you've given

DonShelly: done

cyrusdobbs: they shouldn't have the `.py` bit - should be like line 14 `down_revision = "005_relax_proj_prog_nulls"`

core/db/entities.py
DonShelly: Done

db/migrations/versions/006_replace_enums_with_strings.py
```diff
@@ -0,0 +1,232 @@
...
+Revises: 005_relax_proj_prog_nulls
+Create Date: 2023-07-03 12:10:30.753706
+
+"""
+import sqlalchemy as sa
+from alembic import op
+from sqlalchemy.dialects import postgresql
+
+# revision identifiers, used by Alembic.
+revision = "006_replace_enums_with_strings.py"
```
cyrusdobbs, line 13: `006_replace_enums_with_strings.py` should be `006_replace_enums_with_strings`

cyrusdobbs (changes requested):

db/migrations/versions/006_replace_enums_with_strings.py
DonShelly: done

cyrusdobbs (approved):
LGTM - best to squash on merge as the history contains merge commits

gidsg (approved):
LGTM 👍 
(We should test the migrated fields on Test before deploying to Production)

PR #161, opened 2023-07-05
[FIX] org endpoint only returns orgs referenced by at least one prog

cyrusdobbs:
https://trello.com/c/wr8jrP7N/17-organisation-filter-on-hambleton-district-council-returns-no-data

### Change description
Modifies the organisation endpoint to only return organisations referenced by at least one programme. The organisation table itself can not be trusted as the source because we currently ingest unreferenced organisations & organisations can be left unreferenced if re-ingestion with different source data occurs.

### How to test

### DIFF
--- a/core/controllers/get_filters.py
+++ b/core/controllers/get_filters.py
@@ -12,11 +12,11 @@ from core.db.entities import Organisation, OutcomeDim, Programme, Project, Submi
 
 
 def get_organisation_names():
-    """Returns a list of all distinct organisation names.
+    """Returns a list of all distinct organisation names that are referenced by at least one programme.
 
     :return: List of organisation names
     """
-    organisations = Organisation.query.with_entities(Organisation.id, Organisation.organisation_name).distinct().all()
+    organisations = db.session.query(Organisation).join(Programme).distinct().all()
 
     if not organisations:
         return abort(404, "No organisation names found.")
--- a/tests/controller_tests/test_filters.py
+++ b/tests/controller_tests/test_filters.py
@@ -4,7 +4,12 @@ from flask.testing import FlaskClient
 
 from core.const import DATETIME_ISO_8610
 from core.db import db
-from core.db.entities import Submission
+
+# isort: off
+from core.db.entities import Organisation, Programme, Submission
+
+
+# isort: on
 
 # TODO: Test these endpoints with specific seeded data that doesn't rely on being seeded via the example data model ss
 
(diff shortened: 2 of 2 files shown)


Joel-Pearce:
Looks good to me, I've tested, and it's working, but another issue flagged at the same time was the Organisation name of 'None'. Each Programme will be tied to an Organisation, but in the case of TD-STH (St Helen's) there was no provided Organisation name in the R1 data. It's possible the same issue will arise in R2, unless we make the decision to drop all data parented by an unnamed Organisation. 

This solution prevents the issue from arising if any subsequent round of data contains an Organisation name for TD-STH, but even so we probably want to prevent None from appearing in the drop-down as it's misleading (implying the user is not selecting an Organisation name when in fact they're selecting somewhere in St Helens). 

Perhaps we can fix this by excluding None in the filters, and then we can look to either drop or manually overwrite Programmes for which there is no Organisation name going forward?

cyrusdobbs:
> Looks good to me, I've tested, and it's working, but another issue flagged at the same time was the Organisation name of 'None'. Each Programme will be tied to an Organisation, but in the case of TD-STH (St Helen's) there was no provided Organisation name in the R1 data. It's possible the same issue will arise in R2, unless we make the decision to drop all data parented by an unnamed Organisation.
> 
> This solution prevents the issue from arising if any subsequent round of data contains an Organisation name for TD-STH, but even so we probably want to prevent None from appearing in the drop-down as it's misleading (implying the user is not selecting an Organisation name when in fact they're selecting somewhere in St Helens).
> 
> Perhaps we can fix this by excluding None in the filters, and then we can look to either drop or manually overwrite Programmes for which there is no Organisation name going forward?

I noticed this yesterday as well.. We could filter out certain values but I'm not sure it's the right solution. This kind of data issue should be caught on ingest / validation and failing that testing IMO. I feel as if it opens the gates to filtering out other values like "N/A" etc if they ever occur - what about on the other filters too? This would hide the underlying data from users and could also lead to us taking longer to fix it (although we could warning log any values that we see as None etc to help with this).

I lean towards fixing the data issue in the source data and the live database. We should probably also add validation in to protect against this at ingest time - e.g. flag when non-nullable cells are None, N/A etc.

Joel-Pearce:
We could add Organisation Name to the pre-transformation validation to account for all subsequent data ingests, and will have to either manually update or remove those in historical lacking an Organisation name (can raise with TF).

I'll approve this PR.

Joel-Pearce (approved):

PR #165, opened 2023-07-07
Feature/360 model extract tab updates

bobbi-hbrown:
https://trello.com/c/uWubpFxg/360-priority-1-data-model-extract-to-include-project-name-place-organisation-on-each-tab

- Adds Organisation and Place to all (but 3) tables of data. 
- Adds Project Name to all project level data tables. 
- Adds ExtractedPostcodes to the project table

### Change description

- [X] Unit tests and other appropriate tests added or updated - these need to be addressed

### DIFF
--- a/core/serialization/download_xlsx_serializer.py
+++ b/core/serialization/download_xlsx_serializer.py
@@ -11,13 +11,17 @@ from core.db.queries import get_programme_child_with_natural_keys_query, get_pro
 from core.util import ids
 
 
-def serialize_download_data(programmes, programme_outcomes, projects, project_outcomes) -> dict[str, list[dict]]:
+def serialize_xlsx_data(programmes, programme_outcomes, projects, project_outcomes) -> dict[str, list[dict]]:
+    programme_ids = ids(programmes)
+
+    # organisation_metadata = org_metadata_programme_id(programme_ids)
+    # places_metadata = places_metadata_programme_id(programme_ids)
+
     """Top level serialization of the download data."""
     # Organisation level data
     organisation_refs = [OrganisationSerializer(programme.organisation).to_dict() for programme in programmes]
 
     # Programme level data
-    programme_ids = ids(programmes)
     programme_refs = [ProgrammeSerializer(prog).to_dict() for prog in programmes]
     programme_progresses = get_programme_progress(programme_ids)
     place_details = get_place_details(programme_ids)
@@ -26,7 +30,7 @@ def serialize_download_data(programmes, programme_outcomes, projects, project_ou
 
     # Project level data
     project_ids = ids(projects)
-    project_details = [ProjectSerializer(proj).to_dict() for proj in projects]
+    project_details = get_project_details(projects)
     project_progresses = get_project_progresses(project_ids)
     funding = get_funding(project_ids)
     funding_comments = get_funding_comments(project_ids)
@@ -60,6 +64,41 @@ def serialize_download_data(programmes, programme_outcomes, projects, project_ou
     return download_data
 
 
+def org_metadata_programme_id(programme_ids):
+    # ProjectName, Place, OrganisationName
+
+    organisations = {"organisation": [], "programme_id": []}
+
+    organisation_names = get_programme_child_with_natural_keys_query(ents.PlaceDetail, programme_ids).filter_by(
+        question="Grant Recipient:"
+    )
+
+    for org in organisation_names:
+        organisations["organisation"].append(org.answer)
+        organisations["programme_id"].append(str(org.programme_id))
+
+    return organisations
+
+
+def places_metadata_programme_id(programme_ids):
+    places = {"place_name": [], "programme_id": []}
+
+    place_name = get_programme_child_with_natural_keys_query(ents.PlaceDetail, programme_ids).filter_by(
+        question="Please select your place name"
+    )
+
+    for place in place_name:
+        places["place_name"].append(place.answer)
+        places["programme_id"].append(place.programme_id)
+
+    return places
+
+
+def get_project_details(projects):
+    project_details = [ProjectSerializer(proj).to_dict() for proj in projects]
+    return project_details
+
+
 def get_programme_progress(programme_ids) -> list[dict[str, str]]:
     """Returns serialized ProgrammeProgress models related to provided programmes.
 
@@ -79,6 +118,7 @@ def get_place_details(programme_ids) -> list[dict[str, str]]:
     """
     place_detail_models = get_programme_child_with_natural_keys_query(ents.PlaceDetail, programme_ids).all()
     output = [PlaceDetailSerializer(model).to_dict() for model in place_detail_models]
+
     return output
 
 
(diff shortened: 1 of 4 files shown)


core/serialization/download_xlsx_serializer.py
```diff
@@ -0,0 +1,608 @@
...
+
+def org_metadata_programme_id(programme_ids):
+    organisations = {}
+
+    organisation_names = get_programme_child_with_natural_keys_query(ents.PlaceDetail, programme_ids).filter_by(
+        question="Grant Recipient:"
+    )
+
+    for org in organisation_names:
+        organisations[org.programme.programme_id] = org.answer
```
cyrusdobbs, line 96: I think it might be better to use organisation_name from the organisation_dim table via programme_dim because we do some standardisation on the organisation names

cyrusdobbs: We'll also need mappings from project_ids to the organisation_name via the programme's child projects  - we'll need to pre-load these when we do the query above.

This will be less expensive than querying for the projects parent programme during each loop at serialization time.

core/serialization/download_xlsx_serializer.py
```diff
@@ -0,0 +1,608 @@
...
+
+
+def places_metadata_programme_id(programme_ids):
+    places = {}
+    place_name = get_programme_child_with_natural_keys_query(ents.PlaceDetail, programme_ids).filter_by(
+        question="Please select your place name"
+    )
+
+    for place in place_name:
+        places[place.programme.programme_id] = place.answer
```
cyrusdobbs, line 107: Same point as above...

We'll also need mappings from project_ids to the place name via the programme's child projects - we'll need to pre-load these when we do the query above.

This will be less expensive than querying for the projects parent programme during each loop at serialization time.

core/serialization/download_xlsx_serializer.py
```diff
@@ -0,0 +1,608 @@
...
+    for org in organisation_names:
+        organisations[org.programme.programme_id] = org.answer
+    return organisations
+
+
+def places_metadata_programme_id(programme_ids):
+    places = {}
+    place_name = get_programme_child_with_natural_keys_query(ents.PlaceDetail, programme_ids).filter_by(
+        question="Please select your place name"
+    )
```
cyrusdobbs, line 104: I think the filter for a particular question should be as part of the previous query for efficiency. Might be best to just make a bespoke query for this function rather than reusing the existing function in this case

cyrusdobbs (changes requested):

core/serialization/download_xlsx_serializer.py
```diff
@@ -0,0 +1,608 @@
...
+    funding_comments = get_funding_comments(project_ids, places_metadata, organisation_metadata)
+    private_investments = get_private_investments(project_ids, places_metadata, organisation_metadata)
+    outputs_ref = [OutputDimSerializer(output_dim).to_dict() for output_dim in ents.OutputDim.query.all()]
+    output_data = get_outputs(project_ids, places_metadata, organisation_metadata)
+    outcome_ref = [OutcomeDimSerializer(outcome_dim).to_dict() for outcome_dim in ents.OutcomeDim.query.all()]
+    project_outcomes = [
+        OutcomeDataSerializer(
+            outcome_data,
+            places_metadata[outcome_data.programme.programme_id],
+            organisation_metadata[outcome_data.programme.programme_id],
```
cyrusdobbs, line 57: tests are failing because this should be `outcome_data.project.project_id` because this is project level outcome data

cyrusdobbs (changes requested):
looks good! mostly just an optimisation change to make 👍

core/serialization/download_xlsx_serializer.py
```diff
@@ -0,0 +1,660 @@
...
+        "OutputsRef": outputs_ref,
+        "OutputData": output_data,
+        "OutcomeRef": outcome_ref,
+        "OutcomeData": outcomes,
+        "RiskRegister": risks,
+    }
+    return download_data
+
+
+def org_metadata_programme_id(programme_ids):
```
nsnape, line 93: can we avoid abbreviations in function names please

nsnape: and add a docstring

core/serialization/download_xlsx_serializer.py
```diff
@@ -0,0 +1,660 @@
...
+        # add the project -> org mapping for each child project of programme
+        for project in programme.projects:
+            organisations[project.project_id] = programme.organisation.organisation_name
+
+    del programmes  # don't rely on the garbage collector, just delete now to preserve memory
+
+    return organisations
+
+
+def places_metadata_programme_id(programme_ids):
```
nsnape, line 118: docstring please

core/serialization/download_xlsx_serializer.py
```diff
@@ -0,0 +1,660 @@
...
+        # add the project -> org mapping for each child project of programme
+        for project in programme.projects:
+            places[project.project_id] = place.answer
+
+    del place_name  # don't rely on the garbage collector, just delete now to preserve memory
+
+    return places
+
+
+def get_project_details(projects, places_metadata, organisation_metadata):
```
nsnape, line 145: docstring please

core/serialization/download_xlsx_serializer.py
```diff
@@ -0,0 +1,660 @@
...
+        }
+
+
+class FundingQuestionSerializer:
+    def __init__(self, funding_question: ents.FundingQuestion, place: str, org: str):
+        self.funding_question = funding_question
+        self.place = place
+        self.org = org
+
+    # TODO ProjectName
```
nsnape, line 438: should this be left as a TODO? In ticket https://trello.com/c/uWubpFxg it says this has been done (item crossed off)

nsnape (changes requested):

core/serialization/download_xlsx_serializer.py
cyrusdobbs: ah didn't notice these names

core/serialization/download_xlsx_serializer.py
```diff
@@ -0,0 +1,660 @@
...
+
+    for programme in programmes:
+        # add the programme -> org mapping
+        organisations[programme.programme_id] = programme.organisation.organisation_name
+
+        # add the project -> org mapping for each child project of programme
+        for project in programme.projects:
+            organisations[project.project_id] = programme.organisation.organisation_name
+
+    del programmes  # don't rely on the garbage collector, just delete now to preserve memory
```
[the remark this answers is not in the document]
cyrusdobbs: was thinking to just use this as a precaution instead of relying on the garbage collector but thinking about it more I agree, I don't think it is necessary / probably wont make any difference. It's not memory that is the issue and garbage collector should just clean it up as soon as the function ends anyway...

core/serialization/download_xlsx_serializer.py
cyrusdobbs: The ticket says this:
> FundingQuestions: Remove Guidance Notes, added ProjectID. Note:  just fyi, some of these are programme level Qs and some are Project level. So Project Name may be blank

We can't add ProjectID to this as it's programme level data... 

Ideas?

cyrusdobbs (approved):
Approving to remove my requested changes from when I reviewed Bobbi's changes

core/serialization/download_xlsx_serializer.py
cyrusdobbs: Assuming this is just wrong on the ticket.. I've removed the TODO

PR #167, opened 2023-07-07
511 Add composite_key validation

DonShelly:
[Trello Ticket](https://trello.com/c/xclaqIcd/511-unique-together-validation)

Added Function, tests for unique together validation

### How to test
Run tests

### DIFF
--- a/core/validation/failures.py
+++ b/core/validation/failures.py
@@ -90,12 +90,18 @@ class NonUniqueCompositeKeyFailure(ValidationFailure):
     """Class representing a non-unique-composite_key failure."""
 
     sheet: str
-    row: tuple
+    cols: tuple
+    row: list
 
     def __str__(self):
         """Method to get the string representation of the non-unique-composite_key failure."""
-        row_str = ", ".join(str(i) for i in self.row)
-        return f'Non Unique Row Failure: Sheet "{self.sheet}"; Row "{row_str}" This is a duplicate row'
+        cols_str = ", ".join(str(i) for i in self.cols)
+        row_str = list(self.row)
+        return (
+            f'Non Unique Row Failure: Sheet "{self.sheet}"; '
+            f'Columns "{cols_str}" contains contains a duplicate row consisting of the values: '
+            f'"{row_str}"'
+        )
 
 
 @dataclass
--- a/core/validation/failures.py
+++ b/core/validation/failures.py
@@ -86,6 +86,19 @@ class NonUniqueFailure(ValidationFailure):
 
 
 @dataclass
+class NonUniqueCompositeKeyFailure(ValidationFailure):
+    """Class representing a non-unique-composite_key failure."""
+
+    sheet: str
+    row: tuple
+
+    def __str__(self):
+        """Method to get the string representation of the non-unique-composite_key failure."""
+        row_str = ", ".join(str(i) for i in self.row)
+        return f'Non Unique Row Failure: Sheet "{self.sheet}"; Row "{row_str}" This is a duplicate row'
+
+
+@dataclass
 class WrongTypeFailure(ValidationFailure):
     """Class representing a wrong type failure."""
 
--- a/core/validation/validate.py
+++ b/core/validation/validate.py
@@ -66,6 +66,7 @@ def validate_workbook(workbook: dict[str, pd.DataFrame], schema: dict) -> list[v
         (validate_columns, "columns"),
         (validate_types, "columns"),
         (validate_uniques, "uniques"),
+        (validate_unique_composite_keys, "composite_keys"),
         (validate_foreign_keys, "foreign_keys"),
         (validate_enums, "enums"),
         (validate_nullable, "non-nullable"),  # TODO: add tests for this
@@ -167,6 +168,35 @@ def validate_uniques(
     ]
 
     return non_unique_failures
+
+
+def validate_unique_composite_keys(
+    workbook: dict[str, pd.DataFrame], sheet_name: str, composite_keys: tuple
+) -> list[vf.NonUniqueCompositeKeyFailure]:
+    """
+    Validates the uniqueness of composite keys across multiple columns.
+
+    :param workbook: A dictionary containing sheet names as keys and corresponding pandas DataFrames as values.
+    :param sheet_name: The name of the worksheet to be validated.
+    :param composite_keys: A list of tuples, where each tuple contains the column names
+                           that should have combined uniqueness.
+    :return: A list of non-unique composite key failures, if any exist.
+    """
+
+    sheet = workbook[sheet_name]
+    non_unique_composite_key_failures = []
+
+    # filter dataframe by these columns and find duplicated rows including NaN/Empty
+    mask = sheet[list(composite_keys)].duplicated(keep=False)
+    mask.drop_duplicates(inplace=True)
+
+    if mask.any():
+        failures = [
+            vf.NonUniqueCompositeKeyFailure(sheet=sheet_name, row=row) for row in sheet[mask].itertuples(index=False)
+        ]
+        non_unique_composite_key_failures.extend(failures)
+
+    return non_unique_composite_key_failures
 
 
 def validate_foreign_keys(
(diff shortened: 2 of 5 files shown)


core/validation/validate.py
```diff
@@ -169,6 +170,36 @@ def validate_uniques(
...
+
+    for columns in unique_togethers:
+        # filter dataframe by these columns and find duplicated rows
+        mask = sheet[list(columns)].duplicated(keep=False)
+
+        if mask.any():
+            failures = [
+                vf.NonUniqueTogetherFailure(sheet=sheet_name, row=tuple(row[1:])) for row in sheet[mask].itertuples()
+            ]
+            non_unique_together_failures.extend(failures)
```
cyrusdobbs, line 198: Looks good and almost does what we want:

- I think it's a bit more readable if we just use the mask to filter the rows on row 193 rather than in the list comp. We also need to use `drop_duplicates` on the result because we only want one failure pure duplicate (rather than one for each duplicate row)
```
        mask = sheet[list(columns)].duplicated(keep=False)
        duplicated_rows = sheet[mask].drop_duplicates()
```
- we don't need to cast the row to a tuple on 196 because itertuples already return `namedTuple`s.  
- can also increase readability if we set the `index` param in `itertuples` to `False` then we don't have to slice the row

core/validation_schema.py
```diff
@@ -26,6 +26,7 @@
             "Organisation": "str",
         },
         "uniques": ["Programme ID"],
+        "unique_togethers": ("Programme ID", "Organisation"),
```
cyrusdobbs, line 29: We either need to make the constraint function accept a single composite_key per table like how you've defined the schema here or put these tuples in a list

cyrusdobbs (changes requested):
My bad suggesting `unique_togethers`, I think `composite_keys` is what we want to use after thinking about it.

core/validation/validate.py
DonShelly: done

core/validation_schema.py
DonShelly: done

DonShelly: opted to modify the operation

Joel-Pearce:
My understanding is that these values corresponded more to the unique_together constraint than the composite_key constraint. A composite key is used as a PK for a table, but the PK used for the tables in our db is the UUID "id" column. Then again, this is at the point of ingest, and we haven't actually generated our UUIDs at the point of validation. Do you know what the rationale is for composite_keys over unique_together?

```
 "composite_keys": ("Project ID", "Programme ID")
```
I do, however, have a bit of a concern about this. 

These values can not act as a composite key in the database for the project_dim table as different rounds will have the same human-readable "Project ID", but unique ids; also, we expect all projects in "Project Details" to be unique at the point of ingest so I'm not entirely sure what the point of the composite key is anyway? It might not be an issue for validation of a given worksheet - where rounds will not interact - but I would have it's redundant code as each Project has to be unique at the point of ingest anyway, and a project can't belong to two different programmes.

I will test this with R1 data. Might be worth  getting some QA testing from Hrishi if you haven't already.

Joel-Pearce (changes requested):
Falls over if try to ingest R1.

image

This is an unfortunate consequence of us being so rushed to get historical data in that we couldn't scaffold it with robust unit-testing, but for the time being we will have to QA test all ingest changes using both rounds as there's a possibility we will have to re-ingest R1.

DonShelly:
> Falls over if try to ingest R1.
> 
> image
> 
> This is an unfortunate consequence of us being so rushed to get historical data in that we couldn't scaffold it with robust unit-testing, but for the time being we will have to QA test all ingest changes using both rounds as there's a possibility we will have to re-ingest R1.

Removed the R1 schema changes

core/validation/validate.py
```diff
@@ -169,6 +170,38 @@ def validate_uniques(
...
+    """
+
+    sheet = workbook[sheet_name]
+    non_unique_composite_key_failures = []
+
+    # removing duplicates as to avoid index mismatch
+    sheet.drop_duplicates(subset=list(composite_keys), inplace=True)
+    # filter dataframe by these columns and find duplicated rows including NaN/Empty
+    mask = sheet[list(composite_keys)].duplicated(keep=False)
+    mask.drop_duplicates(inplace=True)
```
cyrusdobbs, line 193: removing the duplicates before identifying the duplicates breaks the intended functionality because no duplicates will ever be detected.

the test for this function `test_validate_unique_composite_keys_invalid` needs correcting too. At the moment the test iterates over each failure returned, making assertions on each failure - but this means if no failures are returned then we just never make an assertion and test passes incorrectly. Need to be really careful with things like this. I would opt to be verbose and explicit in tests and just write out each assertion without using loops etc. It will help with readability and also protect against problems like this.

core/validation/validate.py
```diff
@@ -169,6 +170,38 @@ def validate_uniques(
...
+
+    if mask.any():
+        failures = [
+            vf.NonUniqueCompositeKeyFailure(sheet=sheet_name, row=tuple(row[1:]))
+            for row in sheet[mask].itertuples(index=False)
+        ]
```
cyrusdobbs, line 199: you've implemented the suggestion of using `.itertuples(index=False)` but left in the row slicing `row[1:]` and therefore omitting the first legitimate value from the offending row

cyrusdobbs: also don't need to cast to a tuple - as the itertuples returns a type of tuple anyway

core/validation_schema.py
```diff
@@ -26,6 +26,7 @@
             "Organisation": "str",
         },
         "uniques": ["Programme ID"],
+        "composite_keys": ("Programme ID", "Organisation"),
```
cyrusdobbs, line 29: I think if the implementation expects a single composite key like this then the name of this attribute should be singular i.e. "composite_key" rather than "composite_keys"

cyrusdobbs: same for the rest of the instances of this in this file, the code that references it, and the tests

cyrusdobbs (dismissed):
The functionality is currently broken, but it isn't being picked up by any failing tests.. I explain why this is the case in the comments.

The [README.md](https://github.com/communitiesuk/funding-service-design-post-award-data-store/blob/main/core/validation/README.md) in the `validation` package also needs to be updated to include documentation of this functionality.

core/validation/validate.py
cyrusdobbs: I've added some more comments on the newest review as follow on from this

core/validation/failures.py
```diff
@@ -85,6 +85,23 @@ def __str__(self):
...
+    sheet: str
+    cols: tuple
+    index: int
+
+    def __str__(self):
+        """Method to get the string representation of the non-unique-composite_key failure."""
+        cols_str = ", ".join(str(i) for i in self.cols)
+        return (
+            f'Non Unique Row Failure: Sheet "{self.sheet}"; '
+            f'Columns "{cols_str}" contain non-unique values at row "{self.index}"'
```
nsnape, line 101: maybe this could be a bit clearer? ie something like:
"Columns "{cols_str}" contain non-unique combination of values at row "{self.index}"

nsnape (changes requested):

Joel-Pearce (approved):

core/validation/validate.py
```diff
@@ -169,6 +170,36 @@ def validate_uniques(
...
+    # filter dataframe by these columns and find duplicated rows including NaN/Empty
+    mask = sheet[list(composite_key)].duplicated()
+
+    if mask.any():
+        failures = [
+            vf.NonUniqueCompositeKeyFailure(sheet=sheet_name, cols=composite_key, index=idx + 1)
+            for idx, duplicate in mask.iteritems()
+            if duplicate is True
+        ]
+        non_unique_composite_key_failures.extend(failures)
```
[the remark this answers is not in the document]
Joel-Pearce: Yes, can implement that now.

Not sure on keep=False though, as that will give us both the original row, and the duplicated row (e.g. row 1 and row 3). Do we want to want to return both, or just the ones that come after the first occurrence as technically the first is not a duplicate?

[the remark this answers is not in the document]
cyrusdobbs: looks good now 👍

PR #171, opened 2023-07-11
Revert "Revert "Feature/360 model extract tab updates""

gidsg:
Restore these changes.
Reverts communitiesuk/funding-service-design-post-award-data-store#170

### DIFF
--- a/core/serialization/download_xlsx_serializer.py
+++ b/core/serialization/download_xlsx_serializer.py
@@ -1,0 +1,676 @@
+from sqlalchemy import or_
+from sqlalchemy.orm import joinedload
+
+import core.db.entities as ents
+
+# isort: off
+from core.db.queries import get_programme_child_with_natural_keys_query, get_project_child_with_natural_keys_query
+
+# isort: on
+
+from core.util import ids
+
+
+def serialize_xlsx_data(programmes, programme_outcomes, projects, project_outcomes) -> dict[str, list[dict]]:
+    """Top level serialization of the download data."""
+
+    programme_ids = ids(programmes)
+
+    # get mappings used to supplement (almost) every table with additional information
+    organisation_mapping = get_organisation_mapping(programme_ids)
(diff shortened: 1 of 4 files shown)


core/serialization/download_xlsx_serializer.py
```diff
@@ -38,8 +38,8 @@ def serialize_xlsx_data(programmes, programme_outcomes, projects, project_outcom
     programme_outcomes = [
         OutcomeDataSerializer(
             outcome_data,
-            programme_name_mapping[outcome_data.programme.programme_id],
-            organisation_mapping[outcome_data.programme.programme_id],
+            programme_name_mapping.get(outcome_data.programme.programme_id, None),
```
gidsg, line 41: Do we need the `None` second argument? Is that not the default behaviour of `.get`?

nsnape: No we don't, I forgot to remove after checking

tests/test_serialization.py
```diff
@@ -1,21 +1,62 @@
...
+    assert serialized_data.get("ProjectProgress")
+    assert serialized_data.get("FundingQuestions")
+    assert serialized_data.get("Funding")
+    assert serialized_data.get("FundingComments")
+    assert serialized_data.get("PrivateInvestments")
+    assert serialized_data.get("OutputsRef")
+    assert serialized_data.get("OutputData")
+    assert serialized_data.get("OutcomeRef")
+    assert serialized_data.get("OutcomeData")
+    assert serialized_data.get("RiskRegister")
```
gidsg, line 45: Are we expecting `ExtractedPostcodes` in the XSLX serialisation? I think we should have it if possible.

nsnape (approved):
approving for Gideon as he can't do it as creator

tests/test_serialization.py
gidsg: Scrap that I found it now in the ProjectDetails tab.

PR #172, opened 2023-07-11
Feature/506 frontend filters alphabetically sorted

DonShelly:
Options for Funds, Organisation and Outcomes need to be in alphabetical order on the frontend.

https://trello.com/c/7reotPPo/506-front-end-filters-alphabetically-sorted

Testing:
Tests included


tests/controller_tests/test_filters.py
```diff
@@ -89,6 +89,51 @@ def test_get_funds(seeded_test_client):
...
+class FundTypeIdEnum:
+    TOWN_DEAL = "TD"
+    HIGH_STREET_FUND = "HS"
+
+
+def test_get_funds_alphabetically(seeded_test_client):
+    """
+    Test the function that retrieves funds in alphabetical order.
+
+    :Args:
```
nsnape, line 106: as a general rule, you don't really need to add the args and raises to a test if you don't want since the function is not used by anything (not asking to do anything, just a general observation)

tests/controller_tests/test_filters.py
```diff
@@ -89,6 +89,51 @@ def test_get_funds(seeded_test_client):
     assert all(isinstance(fund["id"], str) for fund in response_json)
 
 
+class MockProgramme:
```
nsnape, line 92: is this redundant? Look like it can be removed

tests/controller_tests/test_filters.py
```diff
@@ -89,6 +89,51 @@ def test_get_funds(seeded_test_client):
...
+    def __init__(self, fund_type_id):
+        self.fund_type_id = fund_type_id
+
+
+class FundTypeIdEnum:
```
nsnape, line 97: likewise, remove this now?

tests/controller_tests/test_filters.py
```diff
@@ -89,6 +89,51 @@ def test_get_funds(seeded_test_client):
...
+    )
+    db.session.add(programme)
+
+    programme = entities.Programme(
+        programme_id="2",
+        programme_name="test programme2",
+        fund_type_id="HS",
+        organisation_id=read_org.id,
+    )
+    db.session.add(programme)
```
nsnape, line 128: for future reference, when adding multiple entities you can add all with one db method call like:

`db.session.add_all((programme_1, programme_2, blah...))`

nsnape: Also, just 2 values is not the best test I think, might be worth adding a 3rd with a value that sits in-between the other 2, then you actually check more thoroughly it is re-ordered at query time. With 2 you could just be getting lucky on the order of 2 binary option

nsnape (changes requested):

tests/controller_tests/test_filters.py
DonShelly: done

tests/controller_tests/test_filters.py
DonShelly: done

tests/controller_tests/test_filters.py
```diff
@@ -89,6 +89,51 @@ def test_get_funds(seeded_test_client):
...
+
+
+class FundTypeIdEnum:
+    TOWN_DEAL = "TD"
+    HIGH_STREET_FUND = "HS"
+
+
+def test_get_funds_alphabetically(seeded_test_client):
+    """
+    Test the function that retrieves funds in alphabetical order.
```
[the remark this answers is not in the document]
DonShelly: done

tests/controller_tests/test_filters.py
DonShelly: understood implementing

DonShelly: done

nsnape: nah, it's fine, I forgot it is set to enum on DB

core/controllers/get_filters.py
```diff
@@ -27,11 +27,12 @@ def get_organisation_names():
 
 
 def get_funds():
-    """Returns a list of all distinct funds.
```
gidsg, line 30: @DonShelly This seems to be just updating the `/funds` endpoint? Are we updating others e.g. `/organisations`?

tests/controller_tests/test_filters.py
```diff
@@ -112,6 +189,154 @@ def test_get_outcome_categories(seeded_test_client):
...
+        organisation_name="Some new Org",
+        geography="Mars",
+    )
+    db.session.add(organisation)
+    db.session.commit()
+
+    read_org = Organisation.query.first()
+
+    prog1 = Programme(
+        programme_id="FHSF001",  # matches id for incoming ingest. Should be replaced along with all children
```
nsnape, line 234: You appear to have copied and pasted all these test setup steps from another test, including comments and names that refer specifically to things not happening in this test. You could either:

- refactor this test to remove all the misleading / unnecessary commenting / naming.
- or, if all the code is similar enough, re-use existing code rather than having a duplicate here

nsnape: I think I'd prefer option 1 as much of this data is non-pertinent to this test

nsnape (changes requested):

nsnape (approved):

PR #173, opened 2023-07-11
Transform Round 2 Outcomes

Joel-Pearce:
Transformation for Outcomes.

### Change description

### How to test
Set breakpoint at relevant point & inspect df.

### DIFF
--- a/core/extraction/round_two.py
+++ b/core/extraction/round_two.py
@@ -17,13 +17,16 @@ from core.extraction.utils import convert_financial_halves, datetime_excel_to_pa
 from core.util import extract_postcodes
 
 
-def ingest_round_two_data(df_ingest: pd.DataFrame) -> Dict[str, pd.DataFrame]:
+def ingest_round_two_data(df_dict: Dict[str, pd.DataFrame]) -> Dict[str, pd.DataFrame]:
     """
     Extract data from Consolidated Round 2 data spreadsheet into column headed Pandas DataFrames.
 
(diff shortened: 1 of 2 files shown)


core/extraction/round_two.py
```diff
@@ -806,29 +811,109 @@ def extract_outcomes(df_input: pd.DataFrame) -> pd.DataFrame:
...
+            "Tab 6 - Outcomes: Section B - Higher Frequency": "Higher Frequency",
+        }
+    )
+
+    df_outcomes["Actual/Forecast"] = df_outcomes.apply(get_actual_forecast, axis=1)
+    df_outcomes["Higher Frequency"] = df_outcomes["Higher Frequency"].replace(0.0, np.nan)
+
+    df_outcomes = df_outcomes[final_cols]
+
+    return df_outcomes
```
nsnape, line 916: as a general observation about this function, there is a lot of local variables, some of which are quite large. We probably get away with it here as it's a rarely used data pipeline, but, for instance, this function has 8 large dataframes in local namespace by the time it gets to the end, and this is as a general rule pretty bad practice (apologies if some of this was inherited from the mess I left btw). But for some of them, there is no actual need to create a new dataframe, for example 

unpivoted_df -> df_unpivoted_financial_years -> df_outcomes

Each time these new df's are created, it is effectively creating an extra copy that it holds in memory, but the one before is never used again, so you could just reuse the variable name each time.  Does that make sense, happy to elaborate more if not?

nsnape (changes requested):

nsnape (approved):

PR #174, opened 2023-07-12
Transform Round 2 Footfall

Joel-Pearce:
Transformation for footfall outcomes. This concatenates the dataframes for outcomes and footfall outcomes.

Also small modifications to the function for attaching a project id to outcomes.

### Change description

### How to test

Please note that the subset used for testing Round 2 will produce an empty dataframe for footfall outcomes. This is because all of the rows for footfall in the subset are effectively null. For this reason, it is advisable to test with the full dataset.

### DIFF
--- a/core/extraction/round_two.py
+++ b/core/extraction/round_two.py
@@ -45,10 +45,11 @@ def ingest_round_two_data(df_dict: Dict[str, pd.DataFrame]) -> Dict[str, pd.Data
     # Note: No data fields for funding comments in Round 2 data-set
     # Note: No data for PSI in Round 2 data-set
     extracted_data["Outputs"] = extract_outputs(df_ingest)
-    extracted_data["Outcomes"] = extract_outcomes(df_ingest, df_lookup)
-
-    # TODO: some project ID's are "multiple". Add ref to programme.
-    extracted_data["df_outcomes_footfall_extracted"] = extract_footfall_outcomes(df_ingest)
+    extracted_data["Outcomes"] = pd.concat(
+        [extract_outcomes(df_ingest, df_lookup), extract_footfall_outcomes(df_ingest, df_lookup)],
+        ignore_index=True,
+        axis=0,
+    )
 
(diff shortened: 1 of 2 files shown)


core/extraction/round_two.py
```diff
@@ -887,7 +888,7 @@ def extract_outcomes(df_input: pd.DataFrame, lookup: pd.DataFrame) -> pd.DataFra
     return unpivoted_df
 
 
-def extract_footfall_outcomes(df_input: pd.DataFrame) -> pd.DataFrame:
+def extract_footfall_outcomes(df_input: pd.DataFrame, df_lookup: pd.DataFrame) -> pd.DataFrame:
     """
     Extract Project Outcome rows from DataFrame.
```
nsnape, line 893: Could just mention in the docstring that it's specifically footfall ones here?

core/extraction/round_two.py
```diff
@@ -898,44 +899,95 @@ def extract_footfall_outcomes(df_input: pd.DataFrame) -> pd.DataFrame:
...
+            "Tab 6 - Outcomes: Section C - Footfall Indicator 1 - Unit": "UnitofMeasurement",
+        }
+    )
+
+    unpivoted_df["Actual/Forecast"] = unpivoted_df.apply(get_actual_forecast, axis=1)
+    unpivoted_df["Higher Frequency"] = np.nan
+
+    unpivoted_df = unpivoted_df[final_cols]
+
+    return unpivoted_df
```
nsnape, line 990: in the returned DF, there are custom Outcome names, but they should all be just ""Year on Year monthly % change in footfall" as pre Round 3 (after discussions with TF we assumed the footfall outcome name SHOULD always be hard-coded, and set it to the above). 

This means it can then be mapped to the correct outcome category (from the dict in const). So for this DF can you just hard code these values please? (It looks like you only have 3 projects with footfall data in the whole data extract?)

nsnape (changes requested):

nsnape (approved):

PR #179, opened 2023-07-14
BAU FIX: bump timeout

gidsg:
### Change description
Requests were timing out on the gunicorn side, curtailing long running ingest tasks. Bumping this timeout solves the immediate issue, although we should also look at query optimisations

### How to test
Round 1 ingest should work after deployment.


Joel-Pearce (approved):
LGTM

PR #181, opened 2023-07-17
[FIX] None organisation is replaced with correct value on ingest

cyrusdobbs:
https://trello.com/c/XUbR38B5/21-bug-qa-board-literal-none-option-appearing-in-the-organisation-list

### Change description
St Helens' organisation is set to "None" in the data, which is obviously incorrect but also misleading when displayed on the frontend as a checkbox option. This change replaces that "None" value with the correct value before ingest.

### How to test
I've tested this locally by first loading the incorrect data in the database, then overwriting it with the data with this change. The /organisation  endpoint no longer returns "None" as an option and instead returns "St Helens Borough Council"


gidsg:
Change looks fine, could we possibly rename this file to `towns_fund_round_one` or similar? As might get confusing as we onboard other funds in future.

gidsg (approved):

PR #183, opened 2023-07-18
[FIX] filter all programme child data by submission

cyrusdobbs:
https://trello.com/c/WQPe8WIC/26-bug-anamolies-in-download-data-file-with-different-filters

### Change description
Fixes two issues:
1. not all data was being filtered by submission (i.e reporting period)
2. additional data was leaking into the returns due to using the wrong variable - this was possibly why it seemed as if region was integral to the issue

### How to test

### DIFF
--- a/core/controllers/download.py
+++ b/core/controllers/download.py
@@ -8,7 +8,9 @@ from datetime import datetime
 
 import pandas as pd
 from flask import abort, make_response, request
+from sqlalchemy import UUID
 from sqlalchemy.orm import joinedload
+from sqlalchemy.sql.operators import and_
 
 from core.const import DATETIME_ISO_8610, EXCEL_MIMETYPE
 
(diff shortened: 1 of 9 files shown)


core/controllers/download.py
```diff
@@ -135,17 +147,17 @@ def get_download_data(
...
-    # combine filtered programmes and project parent programmes
+    # COMBINE filtered programmes and project parent programmes
     combined_projects = {*final_projects, *programme_child_projects}  # unique projects
 
-    # sort by natural keys
+    # SORT by natural keys
     sorted_programmes = sorted(final_programmes, key=lambda x: x.programme_id)
     sorted_projects = sorted(combined_projects, key=lambda x: x.project_id)
-    return sorted_programmes, programme_outcomes, sorted_projects, project_outcomes
+    return sorted_programmes, programme_outcomes, sorted_projects, project_outcomes, submission_ids
```
nsnape, line 160: I may be misunderstanding, but is this sorting serving some purpose? As the excel files are coming out unsorted (see current/recent PR to fix this) in a seemingly random order - possibly due to data being cast to a dict downstream from here (or similar)? 

In fact, I just tried removing the sorted() from the 2 preceding lines, and all local tests still pass, so is this redundant?

tests/controller_tests/test_download.py
```diff
@@ -34,3 +44,144 @@ def test_download_excel_format_empty_db(test_client):
     response = test_client.get("/download?file_format=xlsx")
     assert response.status_code == 200
     assert response.content_type == EXCEL_MIMETYPE
+
+
+def test_download_contains_data_from_the_correct_time_period(seeded_test_client):
```
nsnape, line 49: this might be nit-picking, but this test doesn't appear to actually test what's it's name suggests? Ie it only actually asserts on stuff NOT in the download?

cyrusdobbs:
Closing this as it's outdated and fixed in https://github.com/communitiesuk/funding-service-design-post-award-data-store/pull/195

PR #184, opened 2023-07-18
[BAU_FIX] standardise council names

Joel-Pearce:
Due to multiple variations of organisation names, users were receiving a plethora of differently worded options for what were effectively the same organisation. This PR deals with the issue by:

- Mapping place names to their canonical grant recipient for all rounds.
- Updating ingest logic such that for R1 & R3 - the two rounds in production - whenever an Organisation is rendered childless (i.e. the Programme once pointing to it now points to the canonical Organisation), it is removed from the database.
- Tightening the logic with respect to R1's population of the organisation_ref table such that no childless Organisations will be ingested.


core/extraction/towns_fund.py
```diff
@@ -201,6 +207,13 @@ def get_programme_id(df_lookup: pd.DataFrame, df_place: pd.DataFrame) -> str:
     return "-".join([prefix, code])
 
 
+def get_canonical_organisation_name(df_place: pd.DataFrame) -> str:
```
nsnape, line 210: docstring

core/extraction/towns_fund_round_one.py
```diff
@@ -1370,24 +1373,21 @@ def extract_submission_refs(df_dict: pd.DataFrame) -> pd.DataFrame:
     return df_dict
 
 
-def extract_organisation_refs(df_dict: pd.DataFrame, round_1_data: pd.DataFrame) -> pd.DataFrame:
+def extract_organisation_refs(df_dict: dict[str, pd.DataFrame]) -> dict[str, pd.DataFrame]:
     """
-    Extracts submission IDs from the DataFrame dictionary and populates the 'Organisation_Ref' table
+    Populates the 'Organisation_Ref' table
```
nsnape, line 1378: Some of these docstrings don't meet PEP guidance, ie they are supposed to be in imperative tense, line layout etc. there is actually a tool to check them called pydocstyle, we could potentially add it to the pre-commit at some point to catch these, but you can run it locally if you want to check which ones are wrong. 

No worries if you're too busy though, as it will pick up a lot of problems including ones not part of this PR.

core/controllers/ingest.py
```diff
@@ -313,3 +317,15 @@ def save_submission_file(excel_file, submission_id):
     submission.submission_file = excel_file.stream.read()
     db.session.add(submission)
     db.session.commit()
+
+
+def remove_defunct_organisations():
```
nsnape, line 322: can we rename this function? Defunct is a bit of a loaded term, like the data has changed. We could maybe change it to something more explicitly descriptive. I think the correct database term is unreferenced row or dead row. Probably unreferenced is better

nsnape (changes requested):

core/const.py
```diff
@@ -398,8 +398,175 @@ class FundTypeIdEnum(StrEnum):
...
+    "Mansfield": "Mansfield District Council",
+    "Chatham Town Centre": "Medway Council",
+    "Middlesbrough": "Middlesbrough Council",
+    "Middlesbrough Centre": "Middlesbrough Council",
+    "Milton Keynes": "Milton Keynes City Council",
+    "Newark-on-Trent": "Newark and Sherwood District Council",
+    "Newark": "Newark and Sherwood District Council",
+    "Kidsgrove": "Newcastle-under-Lyme Borough Council",
+    "Newcastle-under-Lyme": "Newcastle-under-Lyme Borough Council",
+    "Newcastle-under-Lyme Town Centre": "Newcastle-under-Lyme Borough Council",
```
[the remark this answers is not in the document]
nsnape: fair, I missed that :)

core/controllers/ingest.py
```diff
@@ -313,3 +316,17 @@ def save_submission_file(excel_file, submission_id):
...
+
+
+def remove_unreferenced_organisations():
+    """Removes organisations no longer referenced by a Programme.
+
+    Some organisation names have been wrong in previous ingest.
+    This will clean them up everytime a new ingest is run updating those names
+    """
+
+    organisations_without_child = Organisation.query.filter(~Organisation.programmes.any()).all()
```
nsnape, line 328: I think this can be optimised further. Using list comprehensions to unpack attributes of a list of results returned is potentially expensive as:

- The original query returns ALL the organisation objects/attributes, not just the attributes we need(id only)
- The list comp then does a lot more in local memory.

This can be improved by using the with_entities method. Also, I think we might need the scalars method to put them all in a list (rather than a list of tuples only containing id in each). Now you SHOULD get a lsit of just the id's, with all the expense done on DB server side

`organisation_ids_without_child = Organisation.query.filter(~Organisation.programmes.any()).with_entities(Organisation.id).scalar()`

(I haven't tested the above at all).

core/controllers/ingest.py
```diff
@@ -313,3 +316,17 @@ def save_submission_file(excel_file, submission_id):
...
+    if organisations_without_child:
+        Organisation.query.filter(Organisation.id.not_in([org.id for org in organisations_without_child])).delete()
```
nsnape, line 330: does this work? It looks like you are trying to delete all the orgs Not without children, ie all the ones with children. Or am I getting confused here?

nsnape (changes requested):

core/controllers/ingest.py
Joel-Pearce: I think I've found a better solution. .with_entities() seems to return the memory address, and scalar() only retrieves the first object. Instead we can just call all() at the end, and then if there's any there delete them (without needing to look up if a particular attribute is in a particular list.

nsnape (approved):
LGTM

PR #186, opened 2023-07-21
Load Round Two

Joel-Pearce:
Loading for Round 2. Includes unit tests, refactoring of ingest, and appropriate schema changes.

### DIFF
--- a/tests/controller_tests/test_historical_ingest.py
+++ b/tests/controller_tests/test_historical_ingest.py
@@ -1,0 +1,459 @@
+from copy import deepcopy
+from datetime import datetime
+
+import pandas as pd
+import pytest as pytest
+
+from core.controllers.ingest import populate_db_historical_data
+from core.controllers.mappings import INGEST_MAPPINGS
+from core.db import db
+
+# isort: off
+from core.db.entities import Funding, Organisation, Programme, Project, Submission, RiskRegister
+
+
+@pytest.fixture
+def r1_workbook_mockup():
+    r1_workbook_mockup = {
+        "Submission_Ref": pd.DataFrame(
+            {
+                "Submission ID": [
+                    "S-R01-1",
+                    "S-R01-2",
+                ],
+                "Submission Date": [datetime(2023, 5, 1), datetime(2023, 5, 1)],
+                "Reporting Period Start": [datetime(2023, 4, 1), datetime(2023, 4, 1)],
+                "Reporting Period End": [datetime(2023, 4, 1), datetime(2023, 4, 1)],
+                "Reporting Round": [1, 1],
+            }
+        ),
+        "Organisation_Ref": pd.DataFrame(
+            {
+                "Organisation": ["Romulan Star Empire", "United Federation Of Planets"],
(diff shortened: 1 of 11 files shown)


tests/controller_tests/test_historical_ingest.py
```diff
@@ -0,0 +1,420 @@
...
+
+    # expect R3 programme to not be overwritten
+    prog = Programme.query.filter_by(programme_id="TD-ROM").all()
+    assert len(prog) == 1
+    assert prog[0] == read_prog
+
+    # expect both projects for the programme have been ingested
+    proj1 = Project.query.filter_by(project_id="TD-ROM-01").first()
+    proj2 = Project.query.filter_by(project_id="TD-ROM-02").first()
+    assert proj1 is not None and proj2 is not None
```
nsnape, line 420: The only thing we could potentially add to tests is the case where R1 exists, R3 does not, then R2 comes in and does update a programme ref'd by R1, without altering the R1 child project data itself?

nsnape (changes requested):

tests/controller_tests/test_historical_ingest.py
Joel-Pearce: I've updated the tests to include this scenario, checking to ensure (a) programme is updated, but that (b) programme's child is not deleted/changed.

tests/controller_tests/test_historical_ingest.py
```diff
@@ -0,0 +1,439 @@
...
+    populate_db_historical_data(r1_workbook_mockup, INGEST_MAPPINGS)
+    populate_db_historical_data(r2_workbook_mockup, INGEST_MAPPINGS)
+
+    # check programme updated (should have new name)
+    prog = Programme.query.filter_by(programme_id="TD-FED").with_entities(Programme.programme_name).all()
+    assert len(prog) == 1
+    assert ("Rebuild Bajor",) in prog
+
+    # ensure child with same programme not deleted or overwritten
+    child = RiskRegister.query.filter_by(risk_name="Dominion Attack").first()
```
nsnape, line 438: this test assertion would be more concise if it actually filtered the RiskRegister entity by filtering on the programme. As currently there is a lot of assumption implied by the comment.

Likewise the part above "# check programme updated (should have new name)" could do a more explicit test, ie, run the query before the R2 upsert and then after, and show they are not the same.

The risk with not doing that (ie just saying it is the same before the upsert in a comment is that any subsequent person editing this test have to trust that comment, but if it's tested that it's always explicitly demonstrated and tested.

nsnape (changes requested):

tests/controller_tests/test_historical_ingest.py
nsnape: For instance, in this current test, in order to check the details of exactly what is being tested for, I'd have to go searching all through the fixture setup code to work out what  the comment was referring to (in this case, that the child actually existed before we inserted R2, and go looking to find out what it was, and that it didn't match the one in the assertion here)

nsnape (approved):

PR #191, opened 2023-07-26
[PAB-476] Round 3 Ingest/Transformation tests

nsnape:
https://trello.com/c/H0eNkLxR

No test coverage on Round 3 Extract/Transformation pipelines. Adding to remove risks of tech-debt etc.

### Change description
+++ Refactor existing test folders
+++ Minor Refactor of pipeline to replace "empty" cells with np.nan - for consistency in testing
+++ Tests for extraction/transformation of project, programme, place, organisation


core/extraction/towns_fund.py
```diff
@@ -324,12 +323,14 @@ def extract_project(df_project: pd.DataFrame, project_lookup: dict, programme_id
...
         lambda row: ",".join(extract_postcodes(row["Locations"])), axis=1
     )  # TODO: keep as a list and store as array of strings in db (not supported by SQLite)
+
     single_lat_long = "Single location __Project Location - Lat/Long Coordinates (3.d.p e.g. 51.496, -0.129)"
     multiple_lat_long = "Multiple locations __Project Locations - Lat/Long Coordinates (3.d.p e.g. 51.496, -0.129)"
     df_project["Lat/Long"] = df_project.apply(
         lambda row: row[single_lat_long] if row["Single or Multiple Locations"] == "Single" else row[multiple_lat_long],
         axis=1,
     )
+    df_project["Postcodes"] = df_project["Postcodes"].replace("", np.nan, regex=True)
```
Joel-Pearce, line 333: Why have you set regex=True? If you're just replacing empty strings I don't think it's necessary.

tests/extraction_tests/test_round_three_extraction.py
```diff
@@ -0,0 +1,181 @@
...
+# Test intended only as a local debug tool
+@pytest.mark.skip(reason="currently this is just a pytest/pycharm debug entrypoint for ingest work")
+def test_ingest_round_two_historical():
+    # TODO: currently testing with small subset of data (to allow reasonable debugging speed)
+    round_two_data = pd.read_excel(
+        "Round 2 Reporting - Consolidation (MASTER).xlsx",
+        # "Round 2 Reporting - Consolidation.xlsx",
+        sheet_name=None,  # extract from all sheets
+    )
+    ingest_round_two_data_towns_fund(round_two_data)
```
Joel-Pearce, line 181: I know that Round 1's integration tested was removed - and that our intention is to potentially remove these tests when they're redundant - but if we're keeping the historical integration test for Round 2 it might be logical to keep the integration test for Round 1 unless we're certain there will be no need to re-ingest or alter the data transformations.

Joel-Pearce (changes requested):
Looks good. Left some small comments on which some further clarification might be helpful.

core/extraction/towns_fund.py
nsnape: oops, left that in by mistake, good spot

tests/extraction_tests/test_round_three_extraction.py
nsnape: Yeah, I changed my mind on removing them as I kept needing to debug the pipeline. I saw R1 had gone, it's probably not the end of the world, and can always add back in if we ever need it

Joel-Pearce (approved):
LGTM

PR #203, opened 2023-08-16
[FMD-23] Download refactor - serialise queried data

nsnape:
https://dluhcdigital.atlassian.net/jira/software/c/projects/FMD/boards/139?modal=detail&selectedIssue=FMD-23

### Change description
Common methods (between Excel and JSON serialisers) to query and serialise table specific data.

General architecture:
Output grouping specific query extenders (add methods to base_query) added as methods to queries module

new module data_serialiser ->  serialise_extract_data

Serialise extract data returns a dict of all tables data. For each "table" in a local mapping  it takes the base query, extends it according to the desired table (adds correct joins, specifies download columns etc), runs it on the DB to get data, passes this to corresponding serialiser schema class to serialise into format ready to transform directly into JSON / Excel

Serialiser function can take list of "tables" to reduce the number of sheets returned (optional)

### How to test


core/serialisation/data_serialiser.py
```diff
@@ -0,0 +1,333 @@
...
+    sheets_required = sheets_required if sheets_required else list(table_queries.keys())
+
+    serialised_results = {}
+
+    for sheet in sheets_required:
+        query_extender, schema = table_queries[sheet]
+        extended_query = query_extender(base_query)
+        table_data = extended_query.all()
+        table_serialised = schema(many=True).dump(table_data)
+        serialised_results[sheet] = table_serialised
```
gidsg, line 83: Could this be a comprehension rather than mutating the dict?

nsnape: I think in cases like this, you loose a lot of readability for limited performance gain, if switching to a comprehension.  After all it's only ever going to be doing like 14 iterations. Since there are quite a few lines executed for each iteration, this stuff can get unnecessarily complex in a comprehension. I'd prefer to keep it nice and verbose in a for loop, then we can easily follow it's steps, which also makes it easier to debug, and extend/change any of the functionality (for instance I may change it to a generator, depending on performance testing results)

core/serialisation/data_serialiser.py
```diff
@@ -0,0 +1,333 @@
...
+
+    class Meta:
+        model = PlaceDetail
+
+    submission_id = auto_field(model=Submission, data_key="SubmissionID")
+    programme_id = auto_field(model=Programme, data_key="ProgrammeID")
+    question = auto_field(data_key="Question")
+    indicator = auto_field(data_key="Indicator")
+    answer = auto_field(data_key="Answer")
+    organisation_name = auto_field(model=Organisation, data_key="OrganisationName")
```
cyrusdobbs, line 221: Looks like this is missing "Place"?

core/serialisation/data_serialiser.py
```diff
@@ -0,0 +1,333 @@
...
+class ProgrammeSchema(SQLAlchemySchema):
+    """Serialise Programme data."""
+
+    class Meta:
+        model = Programme
+
+    programme_id = auto_field(data_key="ProgrammeID")
+    programme_name = auto_field(data_key="ProgrammeName")
+    fund_type_id = auto_field(data_key="FundTypeID")
+    organisation_name = auto_field(model=Organisation, data_key="OrganisationName")
```
cyrusdobbs, line 265: I get why Place isn't included here but is this going to confuse users if they're expecting it as "Place"? Seems mental adding it in twice I know..

tests/serialisation_tests/test_serialisation.py
```diff
@@ -0,0 +1,314 @@
...
+    assert test_serialised_data.get("OutputsRef")
+    assert test_serialised_data.get("OutputData")
+    assert test_serialised_data.get("OutcomeRef")
+    assert test_serialised_data.get("OutcomeData")
+    assert test_serialised_data.get("RiskRegister")
+    assert len(test_serialised_data) == 15
+
+    # assert all tables contain place and organisation (apart from OrgRef, OutputsRef and OutcomeRef)
+    for section_name, data in test_serialised_data.items():
+        if section_name in ["ProgrammeRef", "PlaceDetails", "OrganisationRef", "OutputsRef", "OutcomeRef"]:
```
cyrusdobbs, line 121: Ah PlaceDetails is here... I guess the omission is intentional then?

tests/serialisation_tests/test_serialisation.py
```diff
@@ -0,0 +1,314 @@
...
+
+    #  read into pandas for ease of inspection
+    test_fund_filtered_df = pd.DataFrame.from_records(test_serialised_data["ProjectDetails"])
+
+    project4 = additional_test_data[7]
+    assert project4.project_id not in test_fund_filtered_df["ProjectID"]  # not in SW region
+    assert all(
+        ITLRegion.SouthWest in project.itl_regions
+        for project in Project.query.filter(Project.project_id.in_(test_fund_filtered_df["ProjectID"]))
+    )
```
cyrusdobbs, line 314: do we want some more tests like this for the other filters or happy with the queries tests?

cyrusdobbs (commented):
Looks good!

core/serialisation/data_serialiser.py
nsnape: good spot

core/serialisation/data_serialiser.py
nsnape: oh, it's not just me, SQLAlchemy also thinks it's not a good option, no super obvious option to duplicate the column. I'll hack it...

nsnape: hmm, no actual way to do this cleanly, so I'm going to reject it, as unreasonable and un-programmatic. sqlalchemy-marshmallow will not allow the same field twice, the second overwrites the first. If anyone complains we'll 1) explain why we shouldn't, and 2) if necessary, import base marshmallow field class, and add an aliased column to the query - but feels very hacky

tests/serialisation_tests/test_serialisation.py
nsnape: It basically does the same thing as queries tests, so no point re-inventing the wheel. My thoughts were that queries are tested in test_queries, serialisation is tested in test_serialisation. I started combining both here, but realised there was just a lot of duplication / testing the same thing twice. 

If I find any bugs when testing the download, I've generally been adding query/serialisation combination tests then, otherwise just assuming the individual test coverages on each is enough. Otherwise I'd be writing tests all month :D

nsnape:
> 

yeah, I checked and we're ok with this for now. We can alter later if anyone asks for it. We can do it via tickets in the backlog etc

core/serialisation/data_serialiser.py
cyrusdobbs: makes sense

tests/serialisation_tests/test_serialisation.py
cyrusdobbs: yeah it makes sense to do most of the testing in queries just wanted to make sure

cyrusdobbs (approved):
LGTM 🥞

PR #204, opened 2023-08-17
[SMD-38] adds unused dropdown user-centered validation messages

cyrusdobbs:
https://dluhcdigital.atlassian.net/jira/software/c/projects/SMD/boards/140?modal=detail&selectedIssue=SMD-38

### Change description
- Adds a new Abstract Base Class for validation Failures to implement user centred validation messages
- Adds an option to return only user-centred validation messages on unsuccessful ingestion

### How to test

### DIFF
--- a/core/const.py
+++ b/core/const.py
@@ -113,7 +113,6 @@ class ITLRegion(StrEnum):
 # maps a fund id to its full name (we only store ids in the data model)
 FUND_ID_TO_NAME = {FundTypeIdEnum.HIGH_STREET_FUND: "High Street Fund", FundTypeIdEnum.TOWN_DEAL: "Town Deal"}
 
-
 # TLZ is given to any location outside the primary 12 ITL 1 regions as stated on the link below (previously NUTS & UKZ)
 # This introduces the problem that TLZ now maps to multiple locations (Channel Islands, Isle of Man, Non-geographic)
 # https://en.wikipedia.org/wiki/International_Territorial_Level
@@ -409,7 +408,6 @@ PLACE_NAME_ABBREVIATIONS = {
     "Workington": "WRK",
     "Yeovil": "YEO",
 }
-
 
 TF_PLACE_NAMES_TO_ORGANISATIONS = {
     "Heanor": "Amber Valley Borough Council",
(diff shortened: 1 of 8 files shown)


core/controllers/ingest.py
```diff
@@ -47,6 +47,9 @@ def ingest(body, excel_file):
     :raises ValidationError: If the data fails validation against the specified schema.
     """
     source_type = body.get("source_type")  # required
+    # TODO: Do we want to retain the old functionality? If so, how do we want to toggle it? By request param like this?
+    # user_centered_messaging = body.get("user_centered_messaging")
+    user_centered_messaging = True  # TODO: remove hard-coded variable
```
Joel-Pearce, line 52: Why would we need to toggle unless we plan to have two sets of messages (e.g. one for us, one for the users)? If the users are now responsible for their own uploads, then we shouldn't require specific messages for us (or else we're more like admins). Also the end result of the user-centred messages should still be sufficient for us to find & fix problems.

core/validation/failures.py
```diff
@@ -159,9 +213,30 @@ def __str__(self):
...
+        # additional logic for risk location
+        if sheet == "Risk Register":
+            if self.row_values[1]:
+                # project risk
+                section = f"Project {int(self.row_values[1].split('-')[2])} Risks"
+            else:
+                # programme risk
+                section = "Programme Risks"
+
+        return sheet, section, column, message
```
Joel-Pearce, line 235: What are we doing with the value that is wrong? Are we removing all mention of it from the validation?

core/validation/failures.py
```diff
@@ -159,9 +213,30 @@ def __str__(self):
             f'Value "{self.value}" is not a valid enum value.'
         )
 
+    def uc_message_components(self) -> tuple[str, str, str, str]:
+        message = "You must select an option from the list provided. Do not populate the cell with your own content."
+        sheet, section = INTERNAL_TABLE_TO_FORM_LOCATION[self.sheet]
+        column = INTERNAL_COLUMN_TO_FORM_COLUMN[self.column]
```
Connor-OS, line 219: Could we use a line like 
```suggestion
        column = INTERNAL_COLUMN_TO_FORM_COLUMN.get(self.column) if not 'null' else self.column
```
No point mapping values onto themselves

core/validation/failures.py
```diff
@@ -17,6 +17,44 @@ def __str__(self):
         """Abstract method to get the string representation of the failure."""
 
 
+INTERNAL_TABLE_TO_FORM_LOCATION = {
+    "Project Details": ("Project Admin", "Project Details"),
+    "Project Progress": ("Programme Progress", "Projects Progress Summary"),
+    "Funding": ("Funding Profiles", "Project Funding Profiles"),
+    "Outcome_Data": ("Outcomes", "Outcome Indicators (excluding footfall)"),
+    "RiskRegister": ("Risk Register", "Programme / Project Risks"),
```
Connor-OS, line 25: Can we reliably map 1 section per sheet like this, what if a sheet has more than one section to consider?

core/controllers/ingest.py
cyrusdobbs: I was also thinking this.. just wasn't sure yet so left this in for now.
> "# TODO: Do we want to retain the old functionality?"

There are other "Failures" that we collect and that we won't add "UC" versions of, although in our use case we never return them due to the data that is outputted from transformation so maybe it's a void point. And if there ever were some that we didn't have UC messages for that were returned then they almost certainly should have UC messages.

The crux of the decision is will those old style messages ever be more useful for us as developers. If that was the case though it probably suggests we haven't done this well - rather than having the toggle is useful.

My thinking is that we don't need it and I'll remove the toggle.

core/validation/failures.py
cyrusdobbs: Good point. I don't think I saw it on Elizabeth's UC message but it seems illogical to not include it as it will be useful to use CTRL + F with. I'll look to format it into the message.

core/validation/failures.py
cyrusdobbs: I think it's better to be verbose and add the somewhat redundant mappings in this case. I think this reduces readability for no efficiency gain.

core/controllers/ingest.py
Joel-Pearce: The only use case in which the legacy messages would serve a purpose is if we're ingesting a dataset for which we don't have the original form (R1 & R2, for instance). Otherwise, provided we retain the value from which the error has arisen we should be able to locate the problem in the spreadsheet as easily as with the legacy messages (I think!).

core/validation/failures.py
cyrusdobbs: Good point, this will work for this UC message as there is only one section per sheet that it can occur on. Makes more sense to add the section to the column mapping instead.

core/validation/failures.py
```diff
@@ -159,9 +192,32 @@ def __str__(self):
...
+            section = "Footfall Indicator"
+
+        # additional logic for risk location
+        if sheet == "Risk Register":
+            if self.row_values[1]:
+                # project risk
+                section = f"Project {int(self.row_values[1].split('-')[2])} Risks"
+            else:
+                # programme risk
+                section = "Programme Risks"
```
Joel-Pearce, line 214: I feel like a lot of this could be refactored as it will be reusable across different validation types. Only the message is specific to the error. 

For instance, NonNullableConstraintFailure applies to pretty much every sheet, and so almost all of the code above could go into its UC component. 

I don't think this is urgent or mandatory though, and it's probably better to get the PR in so we can refactor if and when, or put it as a todo.

core/const.py
```diff
@@ -798,3 +796,33 @@ class ITLRegion(StrEnum):
...
+    "Spend (RAG)": ("Spend (RAG)", "Projects Progress Summary"),
+    "Risk (RAG)": ("Risk (RAG)", "Projects Progress Summary"),
+    "Secured": ("Has this funding source been secured?", "Project Funding Profiles"),
+    "GeographyIndicator": ("Geography Indicator", "Outcome Indicators (excluding footfall)"),
+    "Pre-mitigatedImpact": ("Pre-mitigated Impact", "Programme / Project Risks"),
+    "Pre-mitigatedLikelihood": ("Pre-mitigated Likelihood", "Programme / Project Risks"),
+    "PostMitigatedImpact": ("Post-Mitigated Impact", "Programme / Project Risks"),
+    "PostMitigatedLikelihood": ("Post-mitigated Likelihood", "Programme / Project Risks"),
+    "Proximity": ("Proximity", "Programme / Project Risks"),
+}
```
Joel-Pearce, line 828: We could potentially put these in const.py?

Joel-Pearce (approved):
LGTM, I think there's some refactoring we can do along the way but that can happen as we go.

Connor-OS:
Looks good, I've tested end to end and it works great

core/validation/failures.py
cyrusdobbs: Yep sounds good

PR #206, opened 2023-08-24
Feature/smd 41 nullable constraints

Joel-Pearce:
- Adds user-centred messages for non-nullable constraint failures.
- Refactors way in which UC messages are generated.

### DIFF
--- a/core/validation/failures.py
+++ b/core/validation/failures.py
@@ -201,7 +201,7 @@ class InvalidEnumValueFailure(ValidationFailure, TFUCFailureMessage):
         )
 
         # additional logic for outcomes to differentiate between footfall and non-footfall
-        if sheet == "Outcome_Data" and self.row_values[4] == "Year-on-year % change in monthly footfall":
+        if sheet == "Outcomes" and self.row_values[4] == "Year-on-year % change in monthly footfall":
             section = "Footfall Indicator"
 
         # additional logic for risk location
@@ -230,8 +230,21 @@ class NonNullableConstraintFailure(ValidationFailure, TFUCFailureMessage):
             f"is non-nullable but contains a null value(s)."
         )
 
-    def to_user_centered_components(self) -> tuple[str, str, str, str]:
-        return "Unimplemented", "Unimplemented", "Unimplemented", "Unimplemented"
+    def to_user_centered_components(self) -> tuple[str, str, str]:
+        sheet = INTERNAL_TABLE_TO_FORM_TAB[self.sheet]
+        column, section = INTERNAL_COLUMN_TO_FORM_COLUMN_AND_SECTION[self.column]
+
+        message = (
+            f'There are blank cells in column: "{column}". '
+            f"Use the space provided to tell us the relevant information"
+        )
+
+        # additional logic if Project Outputs as always has same section
+        # and can conflict with Outcomes for Unit of Measurement
+        if sheet == "Project Outputs":
+            section = "Project Outputs"
+
+        return sheet, section, message
 
 
 @dataclass
--- a/tests/validation_tests/test_failures.py
+++ b/tests/validation_tests/test_failures.py
@@ -1,10 +1,11 @@
 from core.validation.failures import (
     InvalidEnumValueFailure,
+    NonNullableConstraintFailure,
     serialise_user_centered_failures,
 )
 
 
-def test_serialise_user_centered_failures():
+def test_invalid_enum_user_centered_failures():
     failure1 = InvalidEnumValueFailure(
         sheet="Project Details",
         column="Single or Multiple Locations",
@@ -46,4 +47,46 @@ def test_serialise_user_centered_failures():
                 "which isn't correct. You must select an option from the list provided",
             ]
         }
+    }
+
+
+def test_non_nullable_user_centered_failures():
+    failure1 = NonNullableConstraintFailure(
+        sheet="Project Details",
+        column="Locations",
+    )
+    failure2 = NonNullableConstraintFailure(
+        sheet="Outcome_Data",
+        column="Unit of Measurement",
+    )
+    failure3 = NonNullableConstraintFailure(
+        sheet="Output_Data",
+        column="Unit of Measurement",
+    )
+
+    failures = [
+        failure1,
+        failure2,
+        failure3,
+    ]
+    output = serialise_user_centered_failures(failures)
+    assert output == {
+        "Project Admin": {
+            "Project Details": [
+                'There are blank cells in column: "Project Location(s) - Post Code (e.g. SW1P 4DF)". Use the space '
+                "provided to tell us the relevant information"
+            ]
+        },
+        "Outcomes": {
+            "Outcome Indicators (excluding footfall) / Footfall Indicator": [
+                'There are blank cells in column: "Unit of Measurement". Use the space provided to '
+                "tell us the relevant information"
+            ]
+        },
+        "Project Outputs": {
+            "Project Outputs": [
+                'There are blank cells in column: "Unit of Measurement". Use the space provided to '
+                "tell us the relevant information"
+            ]
+        },
     }
--- a/core/const.py
+++ b/core/const.py
@@ -801,9 +801,14 @@ TABLE_SORT_ORDERS = {
 INTERNAL_TABLE_TO_FORM_TAB = {
     "Project Details": "Project Admin",
     "Project Progress": "Programme Progress",
+    "Programme Progress": "Programme Progress",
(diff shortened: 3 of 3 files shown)


core/util.py
```diff
@@ -89,3 +93,33 @@ def group_by_first_element(tuples: list[tuple]) -> dict[str, list[tuple | Any]]:
...
+
+    # additional logic if column is 'Question'
+    if validation_failure.column == "Question" and sheet == "Project Admin":
+        section = "Place Details"
+
+    # additional logic if Project Outputs as always has same section
+    if sheet == "Project Outputs":
+        section = "Project Outputs"
+
+    return sheet, section, column
```
cyrusdobbs, line 125: The type hint for the validation_failure param in this function should be `TFUCFailureMessage` rather than `Any`.

I think this raises an issue now though. If we want a generic function that will work with different types of validation failures then we should standardise the `TFUCFailureMessage` objects to have a standard interface.

If we elevate the `column`, `sheet` and `row_values` out into the parent class `TFUCFailureMessage` we can be sure that a function like this can work with the objects it's passed and won't need to make checks like above where we do `hasattr` because we'll know the interface for all objects passed to the function.

This raises another issue though - e.g. `NonNullableConstraintFailure` cannot have `row_values` because the precision of the error is only at the column level, rather than cell. Does this mean we would change the validation capture logic to be at the cel level?

Another example is that `NonUniqueCompositeKeyFailure` uses `cols` rather than a single column attribute. 

I think this begs the question - do we standardise these 4 failures so a generic function like this can be robust - or do we just have individual implementations of this logic for each error? I'm not sure the logic within this is actually as generic as it seems - i.e. it covers so many individual edge cases and not all errors can occur on all sheets. Building for reusability and genericism is good practice as a general rule but I think we need to consider this? What do you think?

core/validation/failures.py
```diff
@@ -231,7 +214,14 @@ def __str__(self):
...
     def to_user_centered_components(self) -> tuple[str, str, str, str]:
-        return "Unimplemented", "Unimplemented", "Unimplemented", "Unimplemented"
+        sheet, section, column = get_sheet_section_and_column(self)
+
+        message = (
+            f'There are blank cells in column(s): "{column}".'
+            f"Use the space provided to tell us the relevant information."
+        )
+
+        return sheet, section, message
```
cyrusdobbs, line 225: I've seen Elizabeth state that GDS error messages don't end with a full stop

e.g. `Use the space provided to tell us the relevant information.` should be `Use the space provided to tell us the relevant information`

cyrusdobbs: Also can column ever be plural? do we need the (s) ?

tests/validation_tests/test_failures.py
```diff
@@ -47,3 +49,24 @@ def test_serialise_user_centered_failures():
...
+
+    sheet, section, column = get_sheet_section_and_column(failure1)
+    assert sheet == "Project Admin"
+    assert section == "Place Details"
+    assert column == "A1-A7"
+
+    sheet, section, column = get_sheet_section_and_column(failure2)
+    assert sheet == "Programme Progress"
+    assert section == "Programme-Wide Summary Progress"
+    assert column == "A1-A7"
```
cyrusdobbs, line 72: I think if we do go with a generic function like this then we need to test edge cases more thoroughly

cyrusdobbs (changes requested):

core/util.py
Joel-Pearce: The lack of ability to extract row_values from a NonUniqueCompositeKeyFailure does present some issues in combining all of the above logic into one piece of reusable code. It seemed like a sensible idea, but given how hyper specific each of these errors are going to be (relevant to different columns and sheets, returning different pieces of information from which we reverse engineer back to the original spreadsheet) I think I'll revert to getting the column, sheet, and section bespoke for each function.

Joel-Pearce: And then all of the functions are complete we can assess how much is reusable and potentially recombine those chunks of code into a function like the above.

core/validation/failures.py
Joel-Pearce: I'll remove the full-stop. 

There's a difficulty with reverse engineering the 'Question' and 'Answer' columns from our dataframe back to what they looked like in the original spreadsheet. If we don't have row_values, all we have is a column called 'Answer' being null, but that corresponds to a series of questions we don't have in the values for. Therefore I've opted to indicate the parts of spreadsheet where the questions are (i.e. A1-A7) in a certain section, hence the (s).

core/util.py
cyrusdobbs: sounds good to me

core/validation/failures.py
```diff
@@ -230,8 +230,21 @@ def __str__(self):
...
+            f'There are blank cells in column(s): "{column}". '
+            f"Use the space provided to tell us the relevant information"
+        )
+
+        # additional logic if Project Outputs as always has same section
+        # and can conflict with Outcomes for Unit of Measurement
+        if sheet == "Project Outputs":
+            section = "Project Outputs"
+
+        return sheet, section, message
```
cyrusdobbs, line 247: What are the individual cases within the submission that you expect this to appear for?

cyrusdobbs (changes requested):
just a point about unit tests

core/validation/failures.py
cyrusdobbs: this is mostly for my understanding

Joel-Pearce: A lot of the parent tables are populated by information generated at the point of ingestion, and others will be caught by pre-transformation checks. The questions in e.g. Programme Progress are not filled out by the user, so even though they're non-nullable we wouldn't expect them to be caught by this. Outcomes, Outputs, and RiskRegister won't fall under this as the first two have a dropdown that populates Unit of Measurement (but which if not dropped down will cause UoM to be null, but that's a temporary issue), and the latter won't be extracted if Risk Name (the only non-nullable field) is not present.

That leaves the answers in Place Details, various fields in Project Details, a couple of fields in Funding, and potentially Private Investments (although I'm unsure about the latter as have no experience ingesting rounds where it's present).

tests/validation_tests/test_failures.py
```diff
@@ -47,3 +48,39 @@ def test_serialise_user_centered_failures():
...
+                "information"
+            ]
+        },
+        "Project Outputs": {
+            "Project Outputs": [
+                'There are blank cells in column(s): "Unit of Measurement". Use the space provided to '
+                "tell us the relevant information"
+            ]
+        },
+    }
```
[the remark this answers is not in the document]
Joel-Pearce: For all 10 or so fields where we'd expect this error to occur, with one massive output we do an assertAs assertion on?

core/validation/failures.py
cyrusdobbs: - Place Details won't evoke this failure because Answer is nullable in the schema
- Funding doesn't have any evoke it either
- Private Investment doesn't evoke either because neither of the user fields are non-nullable in the schema

It sounds like this will change for Round 4, and we'll have to make the validation tighter. These changes should come in subsequent tickets IMO though as they've only just been loosely defined.

I can only think of Location (and that weird UoM issue that we've noted) that actually evokes this...

Just wanted to verify it with your understanding as well

Joel-Pearce: If it's only Location in Project Details that will be evoked by this then that makes testing easier. I was labouring under a misapprehension because this issue occurred for almost every sheet for historical rounds.

cyrusdobbs: Cool - we'll definitely have more to add cases for Round 4. We had to keep loosening the validation for Round 3 U seem to remember because they had edge cases where they didn't want to submit data for certain fields so we gave them an ultimatum and they decided to remove the validation. 

I think add a test for the UoM as well as Location and then it should change when we make the change to instead raise a dropdown error when that ticket is actioned.

tests/validation_tests/test_failures.py
```diff
@@ -47,3 +48,56 @@ def test_serialise_user_centered_failures():
...
+                "tell us the relevant information"
+            ]
+        },
+        "Project Outputs": {
+            "Project Outputs": [
+                'There are blank cells in column: "Unit of Measurement". Use the space provided to '
+                "tell us the relevant information"
+            ]
+        },
+    }
```
cyrusdobbs, line 103: I think it will be most useful to just have a single assertion on the full output. The initial granularity is a lot of repetition that I don't think makes it easier to debug.

cyrusdobbs (changes requested):

cyrusdobbs (approved):
LGTM

PR #207, opened 2023-08-24
[SMD-39] Implement WrongTypeFailure UC messages

Connor-OS:
## Wrong Type Error

Added error for wrong types including dates, and numbers usually input incoractly as text. Often simply the unit e.g m^2 is appended to the number.

These errors occur on sheets: Program Progress, Funding Profiles, PSI, Project outputs, Outcomes

### DIFF
--- a/core/const.py
+++ b/core/const.py
@@ -685,6 +685,10 @@ INTERNAL_COLUMN_TO_FORM_COLUMN_AND_SECTION = {
     "Primary Intervention Theme": ("Primary Intervention Theme", "Project Details"),
     "Locations": ("Project Location(s) - Post Code (e.g. SW1P 4DF)", "Project Details"),
     "Lat/Long": ("Project Location - Lat/Long Coordinates (3.d.p e.g. 51.496, -0.129)", "Project Details"),
+    "Private Sector Funding Required": ("Private Sector Funding Required", "Private Sector Investment"),
+    "Private Sector Funding Secured": ("Private Sector Funding Secured", "Private Sector Investment"),
+    "Spend for Reporting Period": ("Financial Year 2022/21 - Financial Year 2025/26", "Project Funding Profiles"),
+    "Amount": ("Financial Year 2022/21 - Financial Year 2025/26", "Project Outputs"),
 }
 
 # message back of uc pre-transformation failure messages for Round 4
@@ -704,3 +708,5 @@ GET_FORM_VERSION_AND_REPORTING_PERIOD = {
     3: ("Town Deals and Future High Streets Fund Reporting Template (v3.0)", "1 October 2022 to 31 March 2023"),
     4: ("Town Deals and Future High Streets Fund Reporting Template (v4.0)", "1 April 2023 to 30 September 2023"),
 }
+
+INTERNAL_TYPE_TO_USER_CENTERED_TYPE = {"datetime64[ns]": "a date", "float64": "a number", "object": "text"}
--- a/core/errors.py
+++ b/core/errors.py
@@ -18,3 +18,15 @@ def validation_error_handler(error: ValidationError):
         "status": 400,
         "title": "Bad Request",
     }, 400
+
+
+class UnimplementedUCException(Exception):
+    """Raised when a validation error occurs that is not supported as a user-centred error message."""
+
+
+def unimplemented_uc_error_handler():
+    return {
+        "detail": "Uncaught workbook validation failure",
+        "status": 500,
+        "title": "Bad Request",
+    }, 500
--- a/core/validation/failures.py
+++ b/core/validation/failures.py
@@ -8,9 +8,11 @@ from abc import ABC, abstractmethod
 from dataclasses import dataclass
 from typing import Any
 
+import core.errors as e
 from core.const import (
     INTERNAL_COLUMN_TO_FORM_COLUMN_AND_SECTION,
     INTERNAL_TABLE_TO_FORM_TAB,
+    INTERNAL_TYPE_TO_USER_CENTERED_TYPE,
     PRETRANSFORMATION_FAILURE_UC_MESSAGE_BANK,
 )
 from core.util import group_by_first_element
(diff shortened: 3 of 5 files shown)


cyrusdobbs (changes requested):
Could you add unit tests for each case please? It should test that given a Failure object, its user centred components are as expected. As it's all so coupled to the input we just need to test realistic and expected scenarios. I think it's important because if we change any mappings later down the line it will affect these functions and if we don't have tests we won't notice.

core/validation/failures.py
```diff
@@ -147,10 +148,41 @@ def __str__(self):
...
+                f"Between columns {column} you entered {actual_type} when we expected {expected_type}. "
+                f"You must enter the required data in the correct format, for example, £5,588.13 or £238,"
+                f"062.50"
+            )
+
+        if sheet in ["Project Outputs", "Outcomes"]:
+            message = (
+                f"Between columns {column} you entered {actual_type} when we expected {expected_type}. "
+                f"You must enter data using the correct format, for example, 9 rather than 9m2. Only use numbers"
+            )
```
cyrusdobbs, line 183: My original comment was:

> Do we know what this case is or is it just an unknown path? If it's the latter I don't think we should let happen silently because it may lead to us returning potentially misleading messages.

I think using if and elif still makes sense. Just for the else we should raise an Exception. One option is to create a new exception like `UnimplementedUCFailure` and then we'll know if that occurs then we haven't accounted for all possibilities

Connor-OS: I think that is a good Idea, might be able to reuse UnimplementedUCFailure in other cases as well. Will we log user failures from the tool so that we can get this information?

core/validation/failures.py
```diff
@@ -147,10 +148,38 @@ def __str__(self):
...
+                message = (
+                    f"Between columns {column} you entered {actual_type} when we expected {expected_type} "
+                    f"enter data using the correct format, for example, 9 rather than 9m2. Only use numbers"
+                )
+            else:
+                message = (
+                    f"For column {column} you entered {actual_type} when we expected {expected_type} "
+                    f"enter data using the correct format, for example, 9 rather than 9m2. Only use numbers"
+                )
+
```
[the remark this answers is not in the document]
Connor-OS: I'm adding a section name: "Outcome Indicators (excluding footfall) and Footfall Indicator"
so the returnwill read

```
"Outcome Indicators (excluding footfall) and Footfall Indicator)": [
          "Between columns E-W you entered text when we expected a number enter data using the correct format, for example, 9 rather than 9m2. Only use numbers"
        ]
```

core/validation/failures.py
cyrusdobbs: So just to clarify I mean raising a run time exception rather than returning a different failure to the user.. If we raise an uncaught exception we'll return 500 to the frontend and have to handle that (i created a ticket for this this morning)

tests/validation_tests/test_failures.py
```diff
@@ -47,3 +48,100 @@ def test_serialise_user_centered_failures():
...
+        sheet="Private Investments",
+        column="Private Sector Funding Secured",
+        expected_type="float64",
+        actual_type="object",
+    )
+    failure6 = WrongTypeFailure(
+        sheet="Funding", column="Spend for Reporting Period", expected_type="float64", actual_type="object"
+    )
+    failure7 = WrongTypeFailure(sheet="Output_Data", column="Amount", expected_type="float64", actual_type="object")
+    failure8 = WrongTypeFailure(sheet="Outcome_Data", column="Amount", expected_type="float64", actual_type="object")
```
cyrusdobbs, line 82: I don't think duplicated failures like this would ever occur

tests/validation_tests/test_failures.py
```diff
@@ -47,3 +48,100 @@ def test_serialise_user_centered_failures():
...
+                "You must enter data using the correct format, for example, 9 rather than 9m2. Only use numbers"
+            ]
+        },
+        "Outcomes": {
+            "Outcome Indicators (excluding footfall) and Footfall Indicator": [
+                "Between columns F-O you entered text when we expected a number. "
+                "You must enter data using the correct format, for example, 9 rather than 9m2. Only use numbers"
+            ]
+        },
+    }
```
cyrusdobbs, line 147: I think it will be best to just have a single assertion on the full output. The initial granularity is a lot of repetition that I don't think makes it easier to debug.

core/validation/failures.py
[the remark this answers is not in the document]
cyrusdobbs:
```
column, section = "F-O", "Outcome Indicators (excluding footfall) and Footfall Indicator"
```
Columns F-O doesn't align with the spreadsheet for football.. could be missleading

tests/validation_tests/test_failures.py
cyrusdobbs: oh yeah I see mb!

core/validation/failures.py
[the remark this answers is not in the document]
Connor-OS: I've expanded it to include the entire range of cells an error could occur, D-O. Not very satisfying I know but maybe the best I can do without making changes to the failure logic itself

core/validation/failures.py
Connor-OS: I need a little bit more guidance here, currently getting errors arround circular imports when trying to raise UnimplementedUCException inside a ValidationError. Also not sure if we've nailed down the behaviour here. whether we still want to return an accordian of validation errors, or raise to a higher level error telling users to contact support

Connor-OS: I've pushed where I've got to with the UnimplementedUCException, I've also thought about a seperate option here

```
        else:
            message = (
                f"In one of the cells on this sheet you entered {actual_type} when we expected {expected_type}."
                f"You must enter data using the correct format, check all cells in which you entered data. If this "
                f"error persists contact support"
            )
```

giving a generic message that giving the option to contact support if the error persists. Probably more thinking required here form me.

core/exceptions.py
```diff
@@ -0,0 +1,14 @@
...
+        self.failure_message = "Unknown validation error, Please contact support"
+
+
+def unimplemented_uc_error_handler(error: UnimplementedUCException):
+    return {
+        "detail": "Uncaught workbook validation failure",
+        "validation_errors": error.failure_message,
+        "status": 500,
+        "title": "Bad Request",
+    }, 500
```
cyrusdobbs, line 14: Don't think we need `validation_errors` in this case, so we can remove that and the `__init__` from the exception I think. We can just handle a 500 error from the status code and then provide some static content.

core/const.py
```diff
@@ -825,4 +827,16 @@ class ITLRegion(StrEnum):
...
+    "Start Date": ("Start Date", "Projects Progress Summary"),
+    "Completion Date": ("Completion Date", "Projects Progress Summary"),
+    "Date of Most Important Upcoming Comms Milestone (e.g. Dec-22)": (
+        "Date of Most Important Upcoming Comms Milestone",
+        "Projects Progress Summary",
+    ),
+    "Private Sector Funding Required": ("Private Sector Funding Required", "Private Sector Investment"),
+    "Private Sector Funding Secured": ("Private Sector Funding Secured", "Private Sector Investment"),
+    "Spend for Reporting Period": ("G-Y", "Project Funding Profiles"),
+    "Amount": ("E-W", "Project Outputs"),
```
Joel-Pearce, line 839: I don't think "E-W" or "G-Y" are very useful. They're far too vague, and the letters in the spreadsheet to represent the X axis shouldn't be used interchangeably with the actual column names or else it gets confusing (i.e. "Column 'Unit of Measurement' is null" and "Columns 'E-W' are the wrong type" shouldn't really exist in the same validation payload as the use of 'Column' is different in each case). 

This honestly might be a question for design as it's very hard to capture the multiplicity of columns in one generic error message. For now something like, "In columns for Financial Years XXXX-XXX" or "Columns H1 (XXX-XXX), H2 (XXX-XXX) and TOTAL" adheres to our current usage of the word "Column". I admit this is also a suboptimal solution, hence getting design folk involved might be advisable.

Joel-Pearce (changes requested):
One comment on how we present columns to the user. Otherwise, looking good.

core/const.py
Connor-OS: I agree, I'll try and engage someone from design on this point, but for now will make the change you suggested

core/validation/failures.py
cyrusdobbs: Yeah I think this could be an option but I'd rather make sure that we notice the scenarios we haven't implemented during QA

core/validation/failures.py
```diff
@@ -147,10 +149,40 @@ def __str__(self):
...
+                f"You must enter the required data in the correct format, for example, £5,588.13 or £238,"
+                f"062.50"
+            )
+        elif sheet in ["Project Outputs", "Outcomes"]:
+            message = (
+                f"Between columns {column} you entered {actual_type} when we expected {expected_type}. "
+                f"You must enter data using the correct format, for example, 9 rather than 9m2. Only use numbers"
+            )
+        else:
+            raise UnimplementedUCException
```
cyrusdobbs, line 183: need a test case to cover this

core/exceptions.py
```diff
@@ -0,0 +1,10 @@
+class UnimplementedUCException(Exception):
+    """Returned when an uncaught exception is raised"""
+
+
+def unimplemented_uc_error_handler(error: UnimplementedUCException):
+    return {
+        "detail": "Uncaught workbook validation failure",
+        "status": 500,
+        "title": "Bad Request",
+    }, 500
```
cyrusdobbs, line 10: Can this be combined with `errors.py` - either file name works

cyrusdobbs: You can avoid the circular import if you avoid importing using `from`. 

Using `from core.exceptions import UnimplementedUCException` runs the entire `exceptions` module when the `failures` module is run whereas if we do `import core.exceptions as e` and then use `e.UnimplementedUCException` within the function then it lazily loads the errors module, avoiding the circular import.

core/errors.py
```diff
@@ -18,3 +18,15 @@ def validation_error_handler(error: ValidationError):
         "status": 400,
         "title": "Bad Request",
     }, 400
+
+
+class UnimplementedUCException(Exception):
+    """Returned when an uncaught exception is raised"""
```
cyrusdobbs, line 24: I don't think the docstring is quite right. Something like this is more accurate IMO:
```
"""Raised when a validation error occurs that is not supported as a user-centred error message."""
```

cyrusdobbs (approved):
LGTM 👍

Joel-Pearce (approved):
LGTM

PR #209, opened 2023-08-25
BAU: added JSON sucsessful upload response

Connor-OS:
Send response from ingest in JSON format to be in parity with the responses from bad ingests with validation errors.

Sucsessful response is:

```
{
          "detail": "Spreadsheet successfully uploaded",
          "status": 200,
          "title": "success",
}
```

### How to test

### DIFF
--- a/core/controllers/ingest.py
+++ b/core/controllers/ingest.py
@@ -3,7 +3,7 @@ from io import BytesIO
 
 import numpy as np
 import pandas as pd
-from flask import abort, current_app
+from flask import abort, current_app, jsonify
 from sqlalchemy import desc, func
 from werkzeug.datastructures import FileStorage
 
@@ -80,7 +80,13 @@ def ingest(body, excel_file):
 
     save_submission_file(excel_file, submission_id)
 
-    return "Success: Spreadsheet data ingested", 200
+    return jsonify(
+        {
+            "detail": "Spreadsheet successfully uploaded",
+            "status": 200,
+            "title": "success",
+        }
+    )
 
 
 def extract_data(excel_file: FileStorage) -> dict[str, pd.DataFrame]:
--- a/openapi/components.yml
+++ b/openapi/components.yml
@@ -20,11 +20,22 @@ components:
         }
 
     SpreadsheetIngested:
-      type: string
-
-      example:
-        type: string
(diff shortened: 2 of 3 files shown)


openapi/components.yml
```diff
@@ -20,11 +20,15 @@ components:
...
 
       example:
-        type: string
-        example: "Success: Spreadsheet data ingested"
+        type: object
+        example: {
+          "detail": "Spreadsheet successfully uploaded",
+          "status": 200,
+          "title": "success",
+         }
```
cyrusdobbs, line 31: We can define the response a bit more clearly using:
```
      properties:
        detail:
          type: string
        title:
          type: string
        status:
          type: integer
          format: int32
```

core/controllers/ingest.py
```diff
@@ -80,7 +80,14 @@ def ingest(body, excel_file):
 
     save_submission_file(excel_file, submission_id)
 
-    return "Success: Spreadsheet data ingested", 200
+    return jsonify(
+        {
+            "detail": "Spreadsheet successfully uploaded",
+            "status": 200,
+            "validation_errors": {},
```
cyrusdobbs, line 87: The current example doesn't include `validation_errors` and I don't think we need to include it as it will always be empty and we determine if the ingest is successful from the 200 status.

cyrusdobbs (changes requested):

core/controllers/ingest.py
Connor-OS: I've fixed both comments here

core/controllers/ingest.py
```diff
@@ -80,7 +80,13 @@ def ingest(body, excel_file):
...
     save_submission_file(excel_file, submission_id)
 
-    return "Success: Spreadsheet data ingested", 200
+    return jsonify(
+        {
+            "detail": "Spreadsheet successfully uploaded",
+            "status": 200,
+            "title": "success",
+        }
+    )
```
cyrusdobbs, line 89: Could you modify the test at `test_ingest.test_ingest_endpoint` to assert against this

cyrusdobbs (changes requested):
Small test change then should be good to go

cyrusdobbs (approved):
LGTM 👍

PR #211, opened 2023-08-29
[FMD-24] Optimisation - switch main serialisation function to generator

nsnape:
https://dluhcdigital.atlassian.net/jira/software/c/projects/FMD/boards/139?modal=detail&selectedIssue=FMD-47

### Change description
Switching serialise_download_data to be a Generator rather than a function reduces the memory usage of Python when transforming query data into Excel sheets (as it now only "lazily loads" data from each table query as it is added in, rather than holding it all in memory at once)

Reduced specific test case of all R2 data only, from 40.8s (avg) to 32.4s (avg)

### How to test
Deploy to TEST, set some filters, download and check file

### DIFF
--- a/tests/serialisation_tests/test_serialisation.py
+++ b/tests/serialisation_tests/test_serialisation.py
@@ -98,13 +98,15 @@ def test_xlsx_serialization(download_data):
 def test_serialise_download_data_specific_tab(seeded_test_client, additional_test_data):
     """Test that serialiser func doesn't return all "sheets" of data if "sheets_required" passed as param."""
     base_query = download_data_base_query()
-    test_serialised_data = serialise_download_data(base_query, sheets_required=["ProgrammeRef"])
+    test_generator = serialise_download_data(base_query, sheets_required=["ProgrammeRef"])
+    test_serialised_data = {sheet: data for sheet, data in test_generator}
+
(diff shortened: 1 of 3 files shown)


tests/serialisation_tests/test_serialisation.py
```diff
@@ -91,12 +99,14 @@ def test_serialise_download_data_specific_tab(seeded_test_client, additional_tes
     """Test that serialiser func doesn't return all "sheets" of data if "sheets_required" passed as param."""
     base_query = download_data_base_query()
     test_serialised_data = serialise_download_data(base_query, sheets_required=["ProgrammeRef"])
-    assert test_serialised_data.keys() == {"ProgrammeRef"}
+    for x, y in test_serialised_data:
+        print(2)
+    # assert test_serialised_data.keys() == {"ProgrammeRef"}
```
cyrusdobbs, line 104: Is this to just test iterating though the generator? Do we need to print 2s?

cyrusdobbs (approved):
LGTM. nice use of a generator 💯

tests/serialisation_tests/test_serialisation.py
nsnape: haha, no that's just so I can stop the debugger, forgot to remove

PR #213, opened 2023-08-31
[SMD-65]: outputs and outcomes edgecase validation

Joel-Pearce:
Failure to select an output or outcome left the unit of measurement field blank. This resulted in it being picked up as a null failure, when in actuality the issue is an enum failure with the output or outcome. Changed output and outcomes validation so that the fields are treated as enums, and removed null checks for unit of measurement.

### DIFF
--- a/core/const.py
+++ b/core/const.py
@@ -837,6 +837,7 @@ INTERNAL_COLUMN_TO_FORM_COLUMN_AND_SECTION = {
     "Total Project Value": ("Total Project Value (£)", "Private Sector Investment"),
     "Townsfund Funding": ("Award From Townsfund (£)", "Private Sector Investment"),
     "Output": ("Indicator Name", "Project Outputs"),
+    # TODO: currently 'Unit of Measurement' can be left null because of an enum failure, but still picked up
     "Unit of Measurement": ("Unit of Measurement", "Outcome Indicators (excluding footfall) / Footfall Indicator"),
     "Outcome": ("Indicator Name", "Outcome Indicators (excluding footfall) / Footfall Indicator"),
     "Project Name": ("Project Name", "Project Details"),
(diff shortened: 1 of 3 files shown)


core/const.py
```diff
@@ -110,6 +112,229 @@ class ITLRegion(StrEnum):
     SouthWest = "TLK"
 
 
+class OutputCategories(StrEnum):
+    NUMBER_OF_ADDITIONAL_ENTERPRISES_WITH_BROADBAND_ACCESS_OF_AT_LEAST_30MBPS = (
```
nsnape, line 116: I may be missing something, but these are not actually Outcome Categories, but Output names?

core/const.py
```diff
@@ -110,6 +112,229 @@ class ITLRegion(StrEnum):
...
+    NUMBER_OF_TEMPORARY_FT_JOBS_SUPPORTED = "# of temporary FT jobs supported"
+    NUMBER_OF_FULLTIME_EQUIVALENT_FTE_PERMANENT_JOBS_CREATED_THROUGH_THE_PROJECT = (
+        "# of full-time equivalent (FTE) permanent jobs created through the project"
+    )
+    NUMBER_OF_FULLTIME_EQUIVALENT_FTE_PERMANENT_JOBS_SAFEGUARDED_THROUGH_THE_PROJECT = (
+        "# of full-time equivalent (FTE) permanent jobs safeguarded through the project"
+    )
+
+
+class OutcomeCategories(StrEnum):
```
nsnape, line 211: As per comment on outputs, these aren't categories?

core/const.py
```diff
@@ -110,6 +112,229 @@ class ITLRegion(StrEnum):
     SouthWest = "TLK"
 
 
+class OutputCategories(StrEnum):
```
nsnape, line 115: I would be extremely wary of putting anything like this into hard-coded constants/globals in the code-base as:

1. the data is meant to be dynamic, and could theoretically change / be amended and we don't want to have to "support" that / be tied to Towns Fund specific fields in our main repo
2. Other funds will have completely different outcomes, so hard-coding is not a good solution
3. Any "custom" outputs/outcomes can be added, how are these handled in this case?

nsnape (changes requested):

nsnape (changes requested):

nsnape (approved):

PR #233, opened 2023-09-08
[SMD-120] Remove Unneeded FSHF Data from future ingests

Connor-OS:
Deal with the conditional formating in the funding profiles tab of the TF spreedsheet. There are a number of cells on this tab that are covered by a grey box dependant on a number of conditions on page. This PR is to drop these cells from the data extract going forward with round 3 and 4 data extracts.

There are 3 main considerations:
1. all cells for before 2020 are greyed out except for "Other funding sources"
2. if fund type is Town Deal and Programme only is selected the question: "How much of your CDEL forcast is contractually commited" is greyed out
3. if fund type is Future High Street Fund:
        a. Section B is not required
        b. "5% CDEL pre-payment", "RDEL payment", and "How much of your RDEL is commited" are not required
        c. Payment for reporting period is only required forcasted half a year ahead

See images bellow

All of this logic has been added at the end of extract_funding_questions() and extract_funding_data() to drop the relevant cells.

### Change description

- [ ] Unit tests and other appropriate tests added or updated (TODO)

### How to test

### Screenshots of UI changes (if applicable)
basic

1.
before_2020

2.
programme_only

3. a)
section_b

3. b,c)
FHSF

### DIFF
--- a/core/extraction/towns_fund.py
+++ b/core/extraction/towns_fund.py
@@ -422,6 +422,10 @@ def extract_funding_questions(df_input: pd.DataFrame, programme_id: str) -> pd.D
     :param programme_id: Programme id for this ingest.
     :return: A new DataFrame containing the extracted funding questions.
     """
+    if programme_id.split("-")[0] == "HS":
+        # return empty dataframe if fund_type is Future Hidh Street Fund
+        return pd.DataFrame(columns=["Question", "Guidance Notes", "Indicator", "Response", "Programme ID"])
+
     df_input = df_input.iloc[12:19, 2:13].dropna(axis=1, how="all")
     df_input.reset_index(drop=True, inplace=True)
 
@@ -492,6 +496,9 @@ def extract_funding_data(df_input: pd.DataFrame, project_lookup: dict) -> pd.Dat
     :param project_lookup: Dict of project_name / project_id mappings for this ingest.
     :return: A new DataFrame containing the extracted funding data.
     """
+
+    fund_type = next(iter(project_lookup.values())).split("-")[0]
+    check_programme_only = df_input.iloc[17, 4] == "Programme only"
     df_input = df_input.iloc[31:, 2:25]
 
     header_prefix = ["Funding Source Name", "Funding Source Type", "Secured"]
@@ -560,6 +567,51 @@ def extract_funding_data(df_input: pd.DataFrame, project_lookup: dict) -> pd.Dat
 
     df_funding = convert_financial_halves(df_funding, "Reporting Period")
     df_funding.reset_index(drop=True, inplace=True)
+
+    # drop always unused cells for funding secured before 2020 and beyond 2026
+    unused_mask = df_funding.loc[
+        (
+            # TODO: refactor Funding Source Type == Towns Fund outside of masks to avoid repeated logic
+            (df_funding["Funding Source Type"] == "Towns Fund")
+            & (df_funding["Start_Date"].isna() | (df_funding["End_Date"].isna()))
+        )
+    ]
+    df_funding.drop(unused_mask.index, inplace=True)
+
+    if fund_type == "HS":
+        unused_fhsf_mask = df_funding.loc[
+            # drop unused FHSF Questions
+            (
+                (df_funding["Funding Source Type"] == "Towns Fund")
+                & (
+                    df_funding["Funding Source Name"].isin(
+                        [
+                            "Town Deals 5% CDEL Pre-Payment",
+                            "Towns Fund RDEL Payment which is being utilised on TF project related activity",
+                            "How much of your RDEL forecast is contractually committed?",
+                        ]
+                    )
+                )
+            )
+            |
+            # drop unused FHSF forcast cells
+            ((df_funding["Funding Source Type"] == "Towns Fund") & (df_funding["Start_Date"] > datetime(2023, 10, 1)))
+            # TODO: Review logic for future forcast in reporting round 5
+        ]
+        df_funding.drop(unused_fhsf_mask.index, inplace=True)
+
+    if fund_type == "TD" and check_programme_only:
+        unused_td_mask = df_funding.loc[
+            # drop unused TD Questions in the case of Programme Only
+            (
+                (df_funding["Funding Source Type"] == "Towns Fund")
+                & df_funding["Funding Source Name"].isin(["Town Deals 5% CDEL Pre-Payment"])
+            )
+        ]
+        df_funding.drop(unused_td_mask.index, inplace=True)
+
+    df_funding.reset_index(drop=True, inplace=True)
+
     return df_funding
 
 
--- a/core/validation_schema.py
+++ b/core/validation_schema.py
@@ -1,5 +1,336 @@
 import core.const as enums
 
+ROUND_FOUR_TF_SCHEMA = {
+    "Submission_Ref": {
+        "columns": {
+            "Submission Date": "datetime",
+            "Reporting Period Start": "datetime",
+            "Reporting Period End": "datetime",
+            "Reporting Round": "int",
+        },
+        "non-nullable": ["Reporting Period Start", "Reporting Period End", "Reporting Round"],
+    },
+    "Organisation_Ref": {
+        "columns": {
+            "Organisation": "str",
+            "Geography": "str",
+        },
+        "uniques": ["Organisation"],
+        "non-nullable": ["Organisation"],
+    },
+    "Programme_Ref": {
+        "columns": {
+            "Programme ID": "str",
+            "Programme Name": "str",
+            "FundType_ID": "str",
+            "Organisation": "str",
+        },
+        "uniques": ["Programme ID"],
+        "foreign_keys": {"Organisation": {"parent_table": "Organisation_Ref", "parent_pk": "Organisation"}},
+        "enums": {"FundType_ID": enums.FundTypeIdEnum},
+        "non-nullable": ["Programme ID", "Programme Name", "FundType_ID", "Organisation"],
+    },
+    "Programme Progress": {
+        "columns": {
+            "Programme ID": "str",
+            "Question": "str",
+            "Answer": "str",
+        },
+        "foreign_keys": {
+            "Programme ID": {"parent_table": "Programme_Ref", "parent_pk": "Programme ID"},
+        },
+        # TODO: Round 4 Answer added because this all programme progress are required
+        "non-nullable": ["Programme ID", "Question", "Answer"],
+    },
+    "Place Details": {
+        "columns": {
+            "Programme ID": "str",
+            "Question": "str",
+            "Answer": "str",
+            "Indicator": "str",
+        },
+        "foreign_keys": {
+            "Programme ID": {"parent_table": "Programme_Ref", "parent_pk": "Programme ID"},
+        },
+        "composite_key": (
+            "Programme ID",
+            "Question",
+            "Indicator",
+        ),
+        # TODO: Round 4 Answer added because place details data is required
+        "non-nullable": ["Programme ID", "Question", "Indicator", "Answer"],
+    },
+    "Funding Questions": {
+        "table_nullable": True,
+        "columns": {
+            "Programme ID": "str",
+            "Question": "str",
+            "Indicator": "str",
+            "Response": "str",
+            "Guidance Notes": "str",
+        },
+        "foreign_keys": {
+            "Programme ID": {"parent_table": "Programme_Ref", "parent_pk": "Programme ID"},
+        },
+        "composite_key": (
+            "Programme ID",
+            "Question",
+            "Indicator",
+        ),
+        "non-nullable": ["Programme ID", "Question"],
+    },
+    "Project Details": {
+        "columns": {
+            "Project ID": "str",
+            "Programme ID": "str",
+            "Project Name": "str",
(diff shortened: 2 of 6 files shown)


core/extraction/towns_fund.py
```diff
@@ -437,6 +437,8 @@ def extract_funding_questions(df_input: pd.DataFrame, programme_id: str) -> pd.D
     fund_questions_df["Response"].replace("< Select >", "", inplace=True)
     fund_questions_df["Programme ID"] = programme_id
     fund_questions_df = fund_questions_df.reset_index(drop=True)
+    if programme_id.split("-")[0] == "HS":
+        fund_questions_df.drop(fund_questions_df.index, inplace=True)
```
Joel-Pearce, line 441: If a user is uploading for a FHSF we don't even need to run the function that extracts and transforms funding question data. We can skip this entire line of code:

```
    towns_fund_extracted["Funding Questions"] = extract_funding_questions(
        df_ingest["4a - Funding Profiles"],
        programme_id,
    )
```

core/extraction/towns_fund.py
```diff
@@ -547,6 +549,26 @@ def extract_funding_data(df_input: pd.DataFrame, project_lookup: dict) -> pd.Dat
...
+                    ]
+                )
+            )
+            |
+            # drop unused FHSF cells
+            (df_funding["Secured"].isna())
+            & ((df_funding["Start_Date"] > datetime(2023, 10, 1)) | (df_funding["Start_Date"].isna()))
+        ]
+        # TODO: Remove hard coded date
+        df_funding = df_funding.drop(unused_fhsf_mask.index)
```
Joel-Pearce, line 571: A user only uploads for either TD or FHSF. In the case where they have selected their fund_type as FHSF, cells pertaining to this financial information are greyed out. Rather than dropping this information after collecting it, it might make more sense to not collect it in the first place?

Joel-Pearce (changes requested):
Couple of comments about the need to drop data when we can just not collect it in the first place, let me know if you have any questions

core/extraction/towns_fund.py
Joel-Pearce: In this case, that might be more difficult than what you've done, or more expensive; if that is the case then dropping might be easier :)

core/validation_schema.py
```diff
@@ -844,6 +844,7 @@
         "non-nullable": ["Programme ID", "Question", "Indicator", "Answer"],
     },
     "Funding Questions": {
+        "table_nullable": True,
```
cyrusdobbs, line 847: This can be added for the Round 3 schema as well

core/extraction/towns_fund.py
Connor-OS: Yeah, I agree with this. Only issue is, correct me if I'm wrong, the validator expects an empty sheet with the correct columns to be there so we would need to return a dummy sheet and this was the simplest way I could implement that behaviour

core/extraction/towns_fund.py
Connor-OS: I did think about this. I think might be quite a bit harder, since we use iloc to grab the information of the page in the first instance, but we need to drop data based on some very specific rules, that would be hard to apply there.

core/extraction/towns_fund.py
```diff
@@ -547,6 +553,38 @@ def extract_funding_data(df_input: pd.DataFrame, project_lookup: dict) -> pd.Dat
...
+                    [
+                        "Town Deals 5% CDEL Pre-Payment",
+                        "Towns Fund RDEL Payment which is being utilised on TF project related activity",
+                        "How much of your RDEL forecast is contractually committed?",
+                    ]
+                )
+            )
+            |
+            # drop unused FHSF cells
+            (df_funding["Secured"].isna()) & (df_funding["Start_Date"] > end_of_reporting_period + timedelta(days=1))
```
nsnape, line 575: Looking at the form, this should only happen for the "mandatory" funding sources, not the custom ones. Need to adda bit of logic here, ie just remove these dates if the "funding source name" is one of the remaining 2 valid mandatory funding source names. Does that make sense?

core/extraction/towns_fund.py
```diff
@@ -547,6 +553,38 @@ def extract_funding_data(df_input: pd.DataFrame, project_lookup: dict) -> pd.Dat
...
+        ]
+        df_funding.drop(unused_fhsf_mask.index, inplace=True)
+
+    if list(project_lookup.values())[0].split("-")[0] == "TD" and check_programme_only:
```
nsnape, line 579: this line sort of gets done twice, might be cleaner to just do it once at the start of the function and set to a variable, ie `fund_type = project_lookup.values())[0].split("-")[0]`

core/extraction/towns_fund.py
```diff
@@ -547,6 +553,38 @@ def extract_funding_data(df_input: pd.DataFrame, project_lookup: dict) -> pd.Dat
...
+        unused_td_mask = df_funding.loc[
+            # drop unused TD Questions in the case of Programme Only
+            (df_funding["Funding Source Name"].isin(["Town Deals 5% CDEL Pre-Payment"]))
```
nsnape, line 582: The only problem with this approach is that users could theoretically legitimately add, for example, "Town Deals 5% CDEL Pre-Payment" as a "custom" funding source name, and in this case should be able to enter these fields without them being dropped? I guess we could just not worry about that very unlikely edge-case? Might be worth checking with a BA?

core/extraction/towns_fund.py
nsnape: you could potentially add an optional param to the funding question extraction function, defaulted to False, something like `return_empty_sheet=False`, then add a line right at the start of the function that just returns an empty df with the correct columns? 

This would save on unnecessary computation.

tests/extraction_tests/resources/mock_sheet_data/start_page_mock.csv
```diff
@@ -3,7 +3,7 @@
 ,,,,,,,,,
 ,,,,,,,,,
 ,Reporting Period: ,,,,,,,,
-0,1 October 2022 to 31 March 2023,,,,,,,,
+0,1 April 2023 to 30 September 2023,,,,,,,,
```
nsnape, line 6: does this mean the tests have effectively been changed from round 3 to round 4?

nsnape: We should probably leave the data as is (testing round 3) and extend as necessary, on import, for further subsequent rounds. That way keep all the historical testing intact.

tests/extraction_tests/resources/mock_sheet_data/funding_profiles_mock.csv
```diff
@@ -19,7 +19,7 @@ All the totals are automatically calculated.  If you need a table with more rows
 
 ",,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,
```
nsnape, line 20: I think if we go with this approach of have different mock data sources for HS and FHSF, we should at least split them into different sub-folders to keep it clearer.

nsnape: same for assertion data

nsnape (changes requested):

core/extraction/towns_fund.py
```diff
@@ -547,6 +553,38 @@ def extract_funding_data(df_input: pd.DataFrame, project_lookup: dict) -> pd.Dat
 
     df_funding = convert_financial_halves(df_funding, "Reporting Period")
     df_funding.reset_index(drop=True, inplace=True)
+
+    # drop always unused cells TF funding before 2020
+    unused_mask = df_funding.loc[(df_funding["Secured"].isna()) & (df_funding["Start_Date"].isna())]
+    df_funding.drop(unused_mask.index, inplace=True)
```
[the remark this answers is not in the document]
Connor-OS: The Yes/No is a drop down. So ideally users will get a dropdown error if they don't enter that field. I get that it is a bit of a hack/loose. Maybe there is a better way though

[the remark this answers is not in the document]
Connor-OS: Good spot about 25/26. I would have missed that

core/extraction/towns_fund.py
nsnape: Have you treid the Pndas .isempty method?
`assert df.empty`

nsnape: Or (apparently) even faster:
`assert len(df.index) > 0:`

tests/extraction_tests/resources/mock_sheet_data/start_page_mock.csv
Connor-OS: reverted, this in my latest push 👍

tests/extraction_tests/resources/mock_sheet_data/funding_profiles_mock.csv
Connor-OS: I have now changed mock data to a single source of truth

core/extraction/towns_fund.py
```diff
@@ -492,6 +496,9 @@ def extract_funding_data(df_input: pd.DataFrame, project_lookup: dict) -> pd.Dat
     :param project_lookup: Dict of project_name / project_id mappings for this ingest.
     :return: A new DataFrame containing the extracted funding data.
     """
+
+    fund_type = list(project_lookup.values())[0].split("-")[0]
```
nsnape, line 500: Although probably not a problem with a dict this small, a more efficient (better practice) way to do this is
`next(iter(project_lookup.values())).split("-")[0]`
as it doesn't then do the relatively expensive transaction of casting to a full list of which you don't actually need any value other than the first

tests/extraction_tests/test_round_three_extraction.py
```diff
@@ -314,6 +329,40 @@ def test_extract_funding_data(mock_funding_sheet, mock_project_lookup):
...
 
+def test_extract_funding_data_programme_only(mock_funding_sheet, mock_project_lookup):
+    """Test project level funding data extract does not include excluded data when programme only is selected"""
+    mock_funding_sheet.iloc[17, 4] = "Programme only"
+    extracted_funding_data = tf.extract_funding_data(mock_funding_sheet, mock_project_lookup)
+    assert not extracted_funding_data["Funding Source Name"].isin(["Town Deals 5% CDEL Pre-Payment"]).any()
+
+
+def test_extract_funding_data_fhsf(mock_funding_sheet, mock_project_lookup):
+    """Test project level funding data extract does not include excluded questions or excluded forcasts. when
```
nsnape, line 340: typo in forecast

tests/extraction_tests/test_round_three_extraction.py
```diff
@@ -314,6 +329,40 @@ def test_extract_funding_data(mock_funding_sheet, mock_project_lookup):
...
+                "How much of your RDEL forecast is contractually committed?",
+            ]
+        )
+        .any()
+    )
+    assert (
+        len(
+            extracted_funding_data[
+                (extracted_funding_data["Funding Source Type"] == "Towns Fund")
+                & (extracted_funding_data["Start_Date"] > datetime(2023, 10, 1))
```
nsnape, line 359: maybe could add a quick assertion to check that some rows are indeed extracted too, ie if start date is less than datetime(2023, 10, 1). Good to check the expected data as well as what has been removed

core/validation_schema.py
```diff
@@ -371,6 +371,7 @@
             "Programme ID": {"parent_table": "Programme_Ref", "parent_pk": "Programme ID"},
         },
         "non-nullable": ["Programme ID", "Question"],
+        "table_nullable": True,
```
nsnape, line 374: I think you may have added this to Round 2 instead of Round 3

nsnape: Which also makes me think, whilst you're here could you re-order the schema's to appear in numerical order in the file? Makes more sense than the random order they're in now

core/extraction/towns_fund.py
nsnape: did we just decide that we'd leave this as is?

nsnape (changes requested):

core/extraction/towns_fund.py
Connor-OS: I've added some conditional logic to ensure: df_funding["Funding Source Type"] == "Towns Fund" which should ensure that "Other funding sources" are unaffected

core/extraction/towns_fund.py
Connor-OS: Good to know, and added. I can see how this eis more effecient

core/validation_schema.py
nsnape: haha that's a hard change to check, as the schemas are quite similar, git has only detected partial changes :D It's too clever sometimes...

tests/extraction_tests/test_round_three_extraction.py
Connor-OS: Changed line 336 and 345 to assert on the expected length of each extracted dataframe 200, and 104 respectively

nsnape (commented):
Ok, looking good. Let's test in in TEST env before we merge in though, I can approve when that's done and we have a strategy to update the data in place. Ta

nsnape (approved):

PR #237, opened 2023-09-12
[SMD-144] TF Round 4 validation - Funding Source dropdown validation

cyrusdobbs:
https://dluhcdigital.atlassian.net/jira/software/c/projects/SMD/boards/140?selectedIssue=SMD-144
as part of story:
https://dluhcdigital.atlassian.net/jira/software/c/projects/SMD/boards/140?selectedIssue=SMD-69

### Change description
 - Funding Source wasn't being validated against previously. 
 - These changes enable enum validation on the Funding Source column found on the Funding tab for Round 3 and 4.
 - This has been implemented as a context specific validation rather than using the schema driven validation code because there are hardcoded sections of the spreadsheet that use "Towns Fund" as a "Funding Source Type" that is forbidden in the "Funding Source" column in the "Other Funding Sources" section. This means that if it was implemented using the schema then users could incorrectly enter "Towns Fund" as a Funding Source for the "Other Funding Sources" because all the data is combined into a single "Funding" table during transformation.

### How to test

### DIFF
--- a/core/validation/failures.py
+++ b/core/validation/failures.py
@@ -16,7 +16,7 @@ from core.const import (
     INTERNAL_TYPE_TO_MESSAGE_FORMAT,
     PRETRANSFORMATION_FAILURE_MESSAGE_BANK,
 )
-from core.util import group_by_first_element
+from core.util import get_project_number, group_by_first_element
 from core.validation.exceptions import UnimplementedErrorMessageException
 
 
@@ -159,7 +159,7 @@ class NonUniqueCompositeKeyFailure(ValidationFailure):
 
         if sheet == "Funding Profiles":
             row_str = ", ".join(self.row[1:4])
-            project_number = int(self.row[0].split("-")[2])
+            project_number = get_project_number(self.row[0])
             section = f"Funding Profiles - Project {project_number}"
             message = (
                 f"You have repeated funding information. You must use a new row for each project, "
@@ -167,7 +167,7 @@ class NonUniqueCompositeKeyFailure(ValidationFailure):
                 f' repeat entries for "{row_str}"'
             )
         elif sheet == "Project Outputs":
-            project_number = int(self.row[0].split("-")[2])
+            project_number = get_project_number(self.row[0])
             section = f"Project Outputs - Project {project_number}"
             message = (
                 f'You have entered the indicator "{self.row[1]}" repeatedly. Only enter an indicator once per project'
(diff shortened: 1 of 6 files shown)


tests/validation_tests/test_tf_r4_specific_validations.py
```diff
@@ -192,3 +197,101 @@ def test_validate_project_admin_gis_provided_returns_no_failure():
...
+    workbook = {"Funding": funding_df}
+
+    failures = validate_funding_profiles_funding_source(workbook)
+
+    assert failures == [
+        TownsFundRoundFourValidationFailure(
+            tab="Funding Profiles",
+            section="Project Funding Profiles - Project 1",
+            message='For column "Funding Source", you have entered "Invalid Funding Source Type" which isn\'t '
+            "correct. You must select an option from the list provided",
```
Joel-Pearce, line 228: I'm not entirely sure what I think about "which isn't correct" - it's more that it isn't a valid. I might just be splitting hairs though. What do you think?

cyrusdobbs: No, fair point. I've just taken it straight from the document by content design, like the other enum/dropdown messages. 

It states a generic dropdown error message could be:
`You have entered [number] which isn't correct.  You must select an option from the list provided`.

cyrusdobbs: I imagine we just go with the guidance provided for now and if design decide they want it written differently after testing etc then we can change down the line.

Joel-Pearce: Yeh, I think the validation messages will be tested with users at some point and we can go from there.

Joel-Pearce (approved):
LGTM

core/validation/failures.py
```diff
@@ -159,15 +159,15 @@ def to_message(self) -> tuple[str, str, str]:
 
         if sheet == "Funding Profiles":
             row_str = ", ".join(self.row[1:4])
-            project_number = int(self.row[0].split("-")[2])
+            project_number = get_project_number(self.row[0])
```
Connor-OS, line 162: Didn't we figure out a way to make sure this project number always lines up with the number on the sheet when there are missing/deactivated projects? like on the Blackfriars sheet

cyrusdobbs: Yet to be implemented but captured here https://dluhcdigital.atlassian.net/browse/SMD-147

PR #244, opened 2023-09-14
[SMD-156]: backend authorisation

Joel-Pearce:
A user can only submit for the places within their local authority's remit. When a user submits their returns, a tuple of place names is sent to the data store. The pre-transformation checks inspect the place name of the submission form and compare it to the tuple of place names the user is authorised to submit for. If the user is not authorised to submit for the place name, a PreTransFormationFailure is raised. 

### DIFF
--- a/core/validation/initial_check.py
+++ b/core/validation/initial_check.py
@@ -1,5 +1,3 @@
-from typing import Any
-
 import pandas as pd
 
 import core.validation.failures as vf
@@ -13,7 +11,8 @@ from core.const import (
 def extract_submission_details(
     workbook: dict[str, pd.DataFrame],
     reporting_round: int,
-) -> dict[str, list[str]] | dict[str, dict[Any, Any]]:
+    place_names: tuple[str] | None,
+) -> dict[str, tuple[str]]:
     """
     Extract submission details from the given workbook.
 
@@ -26,6 +25,7 @@ def extract_submission_details(
 
     :param workbook: A dictionary where keys are sheet names and values are pandas DataFrames.
     :param reporting_round: Integer representing the round being ingested.
+    :param place_names: A tuple of place names a given user can submit for.
     :return: A dictionary containing inputs outside expected values for the cell, or
     a list of missing or invalid sheets.
     """
@@ -63,6 +63,15 @@ def extract_submission_details(
     if invalid_sheets:
         return {"Invalid Sheets": missing_sheets}
 
+    # if None, then user is submitting via API
+    if reporting_round == 4 and place_names is not None:
+        # check if the place name is valid but unauthorised so the error is not mistaken for a WrongInputFailure
+        if (
+            wrong_input_checks["Place Name"][0] not in place_names
+            and wrong_input_checks["Place Name"][0] in TF_PLACE_NAMES_TO_ORGANISATIONS.keys()
+        ):
+            return {"Unauthorised Place Name": (wrong_input_checks["Place Name"][0], place_names)}
+
     return wrong_input_checks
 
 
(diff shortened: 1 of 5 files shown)


tests/validation_tests/test_pre_checks.py
```diff
@@ -336,3 +342,10 @@ def test_place_name_is_valid(valid_submission_details):
     failures = pre_transformation_check(valid_submission_details)
     assert len(failures) == 1
     assert isinstance(failures[0], vf.WrongInputFailure)
+
+
+def test_unauthorised_submission():
+    unauthorised_failure_dict = {"Unauthorised Place Name": ("Newark", ("Wigan",))}
+    failures = pre_transformation_check(unauthorised_failure_dict)
+    assert len(failures) == 1
+    assert isinstance(failures[0], vf.UnauthorisedSubmissionFailure)
```
cyrusdobbs, line 351: can we also have a test for an authorized submission

core/validation/initial_check.py
```diff
@@ -13,7 +13,13 @@
 def extract_submission_details(
     workbook: dict[str, pd.DataFrame],
     reporting_round: int,
-) -> dict[str, list[str]] | dict[str, dict[Any, Any]]:
+    place_names: tuple[str] | None,
+) -> (
+    dict[str, list[str]]
+    | dict[str, list[str]]
+    | dict[str, tuple[Any, set[Any]]]
+    | dict[str, tuple[Any, set[Any]] | tuple[Any, set[str]]]
```
cyrusdobbs, line 21: This feels a bit confusing and not very useful - can it be simplified?

cyrusdobbs (changes requested):
2 minor comments - have you tested it end to end?

tests/validation_tests/test_pre_checks.py
Joel-Pearce: That's effectively what this is:

```
def test_extract_round_four_submission_details(valid_workbook_round_four):
    details_dict = extract_submission_details(valid_workbook_round_four, 4, place_names=("Newark",))

    assert details_dict["Form Version"] == (
        "Town Deals and Future High Streets Fund Reporting Template (v4.0)",
        {"Town Deals and Future High Streets Fund Reporting Template (v4.0)"},
    )
    assert details_dict["Reporting Period"] == (
        "1 April 2023 to 30 September 2023",
        {"1 April 2023 to 30 September 2023"},
    )
    assert details_dict["Fund Type"] == (
        "Town_Deal",
        {"Town_Deal", "Future_High_Street_Fund"},
    )
    assert details_dict["Place Name"] == (
        "Newark",
        set(TF_PLACE_NAMES_TO_ORGANISATIONS.keys()),
    )
 ```
 
 places_names contains Newark, the worksheet has Newark selected as the place_name

File: core/validation/initial_check.py (hunk 2 above)
Comment (Joel-Pearce) [reply core/validation/initial_check.py:21]: This is what pycharm updated the returns to.

I think:
```
dict[str, tuple[str]]
```
captures all of the use cases apart from MissingSheets and InvalidSheets, which we don't expect the user to hit anyway.

Comment (cyrusdobbs) [reply core/validation/initial_check.py]: That looks more useful. This probably highlights that this function is doing too many things but not something to change as part of this PR.

File: core/validation/initial_check.py [hunk 3]
```diff
@@ -63,6 +63,15 @@ def extract_submission_details(
     if invalid_sheets:
         return {"Invalid Sheets": missing_sheets}
 
+    # if None, then user is submitting via API
```
gidsg, line 66: I wonder if rather than relying on this implicit check maybe we can set a user agent string or other header, to indicate the request is coming from the frontend?

cyrusdobbs: The data-store doesn't actually care if it's coming from the front-end. This is just implied because the front-end will never send a request to data-store with no place names because the users must be authorized to submit if via the frontend. Unless you think this is worth the extra effort right now I'd be tempted to get this in and tested ASAP and add this to the backlog.

gidsg (approved):
👍

cyrusdobbs (approved):
LGTM

PR #251, opened 2023-09-19
[SMD-71] bring single/multiple location validation outside of schema …

cyrusdobbs:
…validation

https://dluhcdigital.atlassian.net/browse/SMD-71

### Change description
Fixes validation for the location columns on the Project Admin tab so that if a user has not correctly selected Single or Multiple location, then the location columns for those rows are not validated against. This is because the messages could be misleading if we made an assumption as to which sets of rows to validate.

This is done separately from the schema validation so that we can skip validation on these columns for rows where
the user has not selected a Single or Multiple location correctly.

### How to test

### DIFF
--- a/tests/validation_tests/round_four_specific_validations/test_tf_r4_specific_validations.py
+++ b/tests/validation_tests/round_four_specific_validations/test_tf_r4_specific_validations.py
@@ -1,13 +1,13 @@
 import pandas as pd
 import pytest
 
-from core.const import PRE_DEFINED_FUNDING_SOURCES, StatusEnum
+from core.const import PRE_DEFINED_FUNDING_SOURCES, StatusEnum, YesNoEnum
 from core.validation.specific_validations.towns_fund_round_four import (
     TownsFundRoundFourValidationFailure,
     validate,
     validate_funding_profiles_funding_source,
+    validate_locations,
     validate_programme_risks,
-    validate_project_admin_gis_provided,
     validate_project_risks,
     validate_psi_funding_gap,
     validate_sign_off,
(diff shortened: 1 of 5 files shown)


cyrusdobbs:
Tests still to be written. Pushing this so Rishi can test.

tests/validation_tests/round_four_specific_validations/test_tf_r4_specific_validations.py
```diff
@@ -421,3 +382,110 @@ def test_validate_psi_funding_gap_success():
...
+            # this should not produce failures because the Single / Multiple value is invalid (this is picked up during
+            # schema validation)
+            {"Single or Multiple Locations": "Invalid Value", "GIS Provided": pd.NA, "Locations": "", "Lat/Long": ""},
+        ]
+    )
+    workbook = {"Project Details": project_details_df}
+
+    failures = validate_locations(workbook)
+
+    assert failures == []
```
Connor-OS, line 415: Why doesn't this produce a failure like: "There are blank cells in column: Single or Multiple Locations Use the space provided to tell us the relevant information"

Connor-OS (commented):
I think I am missing some context of how this PR came about but my understanding of how this should work is:

- User failes to enter "Single or Multiple Locations" -> They get a standared Non-nullable error, but all validation is skiped on Postcode and GIS.
- They  resubmit with the correction for locations -> Now they recieve any additional validation on the postcode and GIS provied columns.

This breaks the general idea of presenting all validation messages on a single page to the user at once and instead creates a two step process for addressing errors but in this case I think it is justified.

Have I got this correct and is this the functionality implemented?

tests/validation_tests/round_four_specific_validations/test_tf_r4_specific_validations.py
cyrusdobbs: That case is covered by the validation schema enum validation as usual.

Connor-OS: Ah, okay. Got it this is a seperate validation

Connor-OS (approved):

PR #257, opened 2023-09-22
[SMD-67] risk ingest without risk name

cyrusdobbs:
https://dluhcdigital.atlassian.net/browse/SMD-67

### Change description
- modifies Round 4 ingestion to extract risk data regardless of if it has a null Risk Name
- adds validation support to catch this scenario

### How to test

### DIFF
--- a/core/extraction/towns_fund_round_three.py
+++ b/core/extraction/towns_fund_round_three.py
@@ -300,7 +300,7 @@ def extract_project(df_project: pd.DataFrame, project_lookup: dict, programme_id
     # apply header to df with top rows stripped
     df_project = pd.DataFrame(df_project.values[2:], columns=header_row_combined)
 
-    df_project = drop_empty_rows(df_project, "Project Name")
+    df_project = drop_empty_rows(df_project, ["Project Name"])
     df_project = df_project.reset_index(drop=True)
 
     # rename some long column headers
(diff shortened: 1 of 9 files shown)


core/extraction/towns_fund_round_three.py
```diff
@@ -549,7 +549,7 @@ def extract_funding_data(df_input: pd.DataFrame, project_lookup: dict) -> pd.Dat
 
     # TODO: Check we should drop rows with no source name. Or should we use another rule?
```
nsnape, line 550: we could probably also remove this TODO now, I feel like this has been addressed?

nsnape (changes requested):

core/extraction/towns_fund_round_three.py
```diff
@@ -683,8 +681,16 @@ def extract_risks(df_risk: pd.DataFrame, project_lookup: dict, programme_id: str
         "Proximity",
         "RiskOwnerRole",
     ]
-    # TODO: check this drop behaviour. Should we only drop if whole row is empty / null?
-    df_risk_all = drop_empty_rows(df_risk_all, "RiskName")
+    df_risk_all.drop(["Pre-mitigated Raw Total Score", "Post-mitigated Raw Total Score"], axis=1, inplace=True)
+    df_risk_all.columns = risk_columns
+    if not round_four:
```
Connor-OS, line 686: Makes me think that I have added round 3 (suboptimal behaviour to this file in the past, and that I should go back and add this kind of logic.

It does make me a little wary though that this file might become bloated with round specific logic, if rounds start to diverge dramatically 🤔

cyrusdobbs: 1. re your first point - maybe best to go through and check it.

2, I think that is a legitimate concern. We may need larger architectural change if we keep scaling in this way. I think for now this is better than duplicating the whole function with a slight change but agree it won't scale well. We _should_ be moving away from spreadsheets / this template soon.

nsnape (approved):

PR #258, opened 2023-09-22
[SMD-188] Fix failing round 3 file ingests

Connor-OS:
Allow full re-ingest of Round 3 data

Draft PR because we are down to just 4 validation errors

This PR is more like several small bug fixes in one.
Changes:
- towns_fund_round_3.py 292: Some LA's added comments outside the area for inputing data, this created an extra column error so I tightened up the extraction for this sheet.
- towns_fund_round_3.py 883: Some round 3 data is missing projects assigned to outcomes. These rows are dropped (We should not carry this behaviour over to round 4) one test is ignored because and will be deleted.
- round 3 data has a large number of composite key validation errors where data has been previously ingested therefore this constraint is completely relaxed for round 3

Issues remaining:
- One LA has a Dropdown error: "Events programme" is in the cell when the expected is "Events Programme (LA_name)"
- Two LA's have left blank cells instead of inputing postcodes

These issues cannot be solved by relaxing validation rules in the service, therfore we need another way to reingest. The missing data is in the production database so could be downloaded and reingested from there 

### Change description

### How to test

### DIFF
--- a/core/extraction/towns_fund_round_three.py
+++ b/core/extraction/towns_fund_round_three.py
@@ -289,7 +289,7 @@ def extract_project(df_project: pd.DataFrame, project_lookup: dict, programme_id
     """
 
     # strip out everything except Section B
-    df_project = df_project.iloc[23:45, 4:]
+    df_project = df_project.iloc[23:45, 4:12]
 
     # in first header row, replace empty strings with preceding value.
     header_row_1 = [x := y if y is not np.nan else x for y in df_project.iloc[0]]  # noqa: F841,F821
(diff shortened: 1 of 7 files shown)


tests/extraction_tests/test_round_three_extraction.py
```diff
@@ -383,17 +383,6 @@ def test_extract_outcomes_with_invalid_project(mock_outcomes_sheet, mock_project
     )
 
 
-def test_extract_outcomes_with_null_project(mock_outcomes_sheet, mock_project_lookup, mock_programme_lookup):
```
Connor-OS, line 386: This test needs to come back for round 4 it is deleted because we are preserving old functionality

core/extraction/towns_fund_round_three.py
```diff
@@ -885,6 +885,7 @@ def extract_outcomes(df_input: pd.DataFrame, project_lookup: dict, programme_id:
     outcomes_df = pd.concat([outcomes_df, pd.DataFrame(df_input.values[27:37], columns=header_row_combined)])
     # If indicator name or project ref is empty - drop row (cannot map to DB table)
     outcomes_df = drop_empty_rows(outcomes_df, ["Indicator Name"])
+    outcomes_df = drop_empty_rows(outcomes_df, ["Relevant project(s)"])
```
cyrusdobbs, line 888: From my understanding this only needs to be executed for Round 3 right? Can you add a param and preserve the Round 4 functionality as part of this PR

tests/extraction_tests/test_round_three_extraction.py
cyrusdobbs: Can the above change be implemented for Round 4 and then this test be moved to the Round 4 test module

core/extraction/towns_fund_round_three.py
```diff
@@ -289,7 +289,7 @@ def extract_project(df_project: pd.DataFrame, project_lookup: dict, programme_id
     """
 
     # strip out everything except Section B
-    df_project = df_project.iloc[23:45, 4:]
+    df_project = df_project.iloc[23:45, 4:12]
```
cyrusdobbs, line 292: Out of interest what is this change for?

cyrusdobbs (changes requested):

core/extraction/towns_fund_round_three.py
Connor-OS: One of the local authorities wrote comments outside the tables specified for input, the open ended range was reading their comments into the extract and creating ExtraColumn errors

cyrusdobbs: good stuff

cyrusdobbs (approved):
LGTM

PR #259, opened 2023-09-22
[SMD-169] update Round 4 validation schema inline with new template

cyrusdobbs:
https://dluhcdigital.atlassian.net/browse/SMD-169

### Change description
- updated validation schema and failure message support for new columns
- specific validation for the "Leading Factor of Delay" column

Validation behaviour:
Project Delivery Status - always required.
Leading factor of Delay - required if Project Delivery Status is "1. Not yet started" or "3. Ongoing - delayed".
If Leading factor of Delay is present but not required due to the above constraint, then pass validation.

### How to test


tests/validation_tests/test_failures.py
```diff
@@ -255,6 +269,18 @@ def test_invalid_enum_messages():
         'For column "Proximity", you have entered "Value" which isn\'t correct. You '
         "must select an option from the list provided",
     )
+    assert failure13.to_message() == (
```
nsnape, line 272: I think this assertion naming clashes with the previous PR I just approved, which also has a 13

nsnape (approved):
LGTM

PR #260, opened 2023-09-22
BAU: fix more descriptive error messages by casting to string

Connor-OS:
Fixed an edge case where if users entered a numeric character as a funding source name it would be cast to NA when string methods were applied to it by first converting the colum with astype(str)

Seperate refactor of a small method to cast to a string and join with ', ' in failures.py as  was recomended a few days ago

### Change description

### How to test

### DIFF
--- a/core/extraction/utils.py
+++ b/core/extraction/utils.py
@@ -1,5 +1,6 @@
 """Module for reusable DataFrame transformation functions."""
 import re
+from typing import Sequence
 
 import numpy as np
 import pandas as pd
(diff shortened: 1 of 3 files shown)


core/extraction/utils.py
```diff
@@ -98,3 +99,7 @@ def extract_postcodes(s: str | float) -> list[str]:
     else:
         postcode_area_matches = re.findall(POSTCODE_REGEX, str(s))
     return postcode_area_matches
+
+
+def join_as_string(s: Sequence) -> str:
```
nsnape, line 104: can we try and avoid using single letter var names where possible (I know there's a few throughout the Extraction code) but we can try and not have any moving forwards. Ta

nsnape: also, please add a quick docstring

nsnape (changes requested):

PR #262, opened 2023-09-25
[SMD-147] Fix project numbers in validation messages not being in the…

Connor-OS:
Extract a list of active projects during extraction, then use this to reorder the projects to represent their position on the sheet in validation messages.

Initally the problem was that if only projects 1, 2, 4, 7 were active they would fill slots 1,2,3,4 on page so we need a way to map the first list to the second

### How to test

### DIFF
--- a/core/exceptions.py
+++ b/core/exceptions.py
@@ -1,5 +1,8 @@
 class ValidationError(RuntimeError):
     """Validation error."""
 
-    def __init__(self, validation_failures):
+    def __init__(self, validation_failures: list, active_project_ids=None):
+        if active_project_ids is None:
+            active_project_ids = []
+        self.active_project_ids = active_project_ids
         self.validation_failures = validation_failures
--- a/core/util.py
+++ b/core/util.py
@@ -58,10 +58,11 @@ def group_by_first_element(tuples: list[tuple]) -> dict[str, list[tuple | Any]]:
     return nested
 
 
-def get_project_number(project_id):
-    """Extracts the project number from a project ID.
+def get_project_number(project_id: str, active_project_ids: list[str]) -> int:
+    """Map project ID to it's on page position using position in list of active projects
 
     :param project_id: A project ID code
+    :param active_project_ids: A list of project ID's in the sheet in order
     :return: project number
     """
-    return int(project_id.split("-")[2])
+    return active_project_ids.index(project_id) + 1
--- a/core/handlers.py
+++ b/core/handlers.py
@@ -6,7 +6,9 @@ from core.validation.failures import failures_to_messages
 
 
 def handle_validation_error(validation_error: ValidationError):
-    validation_messages = failures_to_messages(validation_error.validation_failures)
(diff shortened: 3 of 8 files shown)


core/util.py
```diff
@@ -59,9 +61,9 @@ def group_by_first_element(tuples: list[tuple]) -> dict[str, list[tuple | Any]]:
...
 
 def get_project_number(project_id):
-    """Extracts the project number from a project ID.
+    """Map project ID to it's on page position using the active projects in g
 
     :param project_id: A project ID code
     :return: project number
     """
-    return int(project_id.split("-")[2])
+    return g.active_project_numbers.index(project_id) + 1
```
nsnape, line 69: I don't think this should be used like this here. It's making some test fail due to application context

nsnape (changes requested):

core/util.py
nsnape: have you removed the old functionality of get_project_number? This seems quite different, do you definitely want to switch it out?

nsnape (changes requested):

core/util.py
Connor-OS: The new functionality is very similar, there is just a significant refactor here.

tests/controller_tests/test_ingest.py
```diff
@@ -411,7 +411,7 @@ def test_ingest_endpoint_returns_uncaught_ingest_error(test_client, example_data
             )
 
     # logs an error
-    assert "ERROR    Sample API:errors.py:32 Uncaught ingest exception." in caplog.text
+    assert "ERROR    Sample API:errors.py:34 Uncaught ingest exception." in caplog.text
```
cyrusdobbs, line 414: ah we can probably change this to:
```
assert "ERROR" in caplog.text
assert "Uncaught ingest exception." in caplog.text
```
so it isn't dependent to the code location

cyrusdobbs (changes requested):
This one looked fun 😄

core/util.py
```diff
@@ -58,10 +58,10 @@ def group_by_first_element(tuples: list[tuple]) -> dict[str, list[tuple | Any]]:
     return nested
 
 
-def get_project_number(project_id):
-    """Extracts the project number from a project ID.
+def get_project_number(project_id, active_project_ids):
```
nsnape, line 61: please can you add type hints to signature, and update docstring to match.

nsnape (changes requested):

cyrusdobbs:
@Connor-OS if this is rebased we should be able to approve it

core/exceptions.py
```diff
@@ -1,5 +1,8 @@
 class ValidationError(RuntimeError):
     """Validation error."""
 
-    def __init__(self, validation_failures):
+    def __init__(self, validation_failures: list, active_project_ids=None):
```
cyrusdobbs, line 4: type hint for `active_project_ids`

cyrusdobbs (approved):

PR #266, opened 2023-09-29
[SMD-193] Pull through excell cell indexes into validation messages

Connor-OS:
This is the result of a spike to see if we could pull through excel cell indexes into the validation messages. The strategey is to add a column to the pretransformed data representing the excel row index. Post transformation I set the table index to this column. Now all rows are indexed by there original excel row number. This can be picked up in validation to display the row number. Column Letters are not considered yet in this PR.

This is a proof of concept:
It only works for Tab: funding profiles, Enum failure: column secured.

Behaviour in other areas of the sheet may be buggy/unpredictable so only test manually for this case

expected output:

Screenshot 2023-09-29 at 15-40-24 Swagger UI

### How to test

### DIFF
--- a/core/controllers/ingest.py
+++ b/core/controllers/ingest.py
@@ -21,7 +21,8 @@ from core.validation.initial_check import (
     extract_submission_details,
     pre_transformation_check,
 )
-from core.validation.specific_validations import towns_fund_round_four as tf_round_4
+
+# from core.validation.specific_validations import towns_fund_round_four as tf_round_4
 from core.validation.validate import validate
 
 REPORTING_ROUND = {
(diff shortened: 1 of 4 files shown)


Joel-Pearce:
Just a quick thought, is it not possible to add "original_index" as a column before any transformations occur? For instance, if I were to write it in pseudocode something like:

```
for df in df_dict:
     df["original_index"] = df.index
```

and then just plonk that value into the error message? 

However, I seem to remember that some excel sheets, when read into a dataframe, result in a jumble of data that don't accurately map to the cell coordinate of the original sheet. Are you finding that the index values accurately map to the y-axis of the original excel sheet before any transformations occur?

Connor-OS:
> Just a quick thought, is it not possible to add "original_index" as a column before any transformations occur? For instance, if I were to write it in pseudocode something like:
> 
> ```
> for df in df_dict:
>      df["original_index"] = df.index
> ```
> 
> and then just plonk that value into the error message?
> 
> However, I seem to remember that some excel sheets, when read into a dataframe, result in a jumble of data that don't accurately map to the cell coordinate of the original sheet. Are you finding that the index values accurately map to the y-axis of the original excel sheet before any transformations occur?

Yep, defo good shout. I think we are all in agreement that this has to be the most effective approach. We just have to be aware of:

- how is this column effected when we de-pivot
- make sure we always pull this column through the full transforamtion baring in mind we use iloc a lot

I've found that I can accuratly map the Y-axis using this approach for the case I have tried here

PR #269, opened 2023-10-03
BAU: fix sign off validation

Joel-Pearce:
Fixes a bug pertaining to sign-off validation.

Sign-off validation requires a pre-transformed worksheet, and therefore cannot happen in conjunction with post-transformation validations specific to round 4. This PR changes sign-off validation to function alongside other pre-transformation checks, but to be specific for round 4.


tests/validation_tests/test_pre_checks.py
```diff
@@ -105,11 +109,47 @@ def valid_workbook():
         ),
         "8 - Review & Sign-Off": pd.DataFrame(
             {
-                0: ["test", "", "", "", "", "", "", "", ""],
-                1: ["", "", "", "", "", "", "", "", ""],
-                2: ["", "", "", "", "", "", "", "", ""],
-                3: ["", "", "", "", "", "", "", "", ""],
-                4: ["", "", "", "", "", "", "", "", ""],
+                0: ["", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", ""],
```
nsnape, line 112: Not that you necessarily need to implement this here, but did you know an equivalent way to write a list of repeating values in python is, for example: `[""] * n`

Joel-Pearce: Ah yes, I did, but it completely slipped my mind whilst mocking these. Will be a lot easier writing as above in the future, thanks

core/validation/initial_check.py
```diff
@@ -150,3 +154,48 @@ def check_missing_sheets(expected_sheets: list[str], workbook: dict[str, pd.Data
...
+
+def validate_sign_off(workbook: dict[str, pd.DataFrame]) -> list[vf.NullInputPreTransformationFailure] | None:
+    """Validates Name, Role, and Date for the Review & Sign-Off Section
+
+    :param workbook: A dictionary where keys are sheet names and values are pandas
+                     DataFrames representing each sheet in the Round 4 submission.
+    :return: ValidationErrors
+    """
+    sheet = workbook.get("8 - Review & Sign-Off")
+    sheet.replace(r"", np.nan, inplace=True)
```
cyrusdobbs, line 167:
```
sheet = workbook.get("8 - Review & Sign-Off")
```
should be
```
sheet = workbook["8 - Review & Sign-Off"]
```
as we don't handle the case where the get returns None.

tests/validation_tests/round_four_specific_validations/test_tf_r4_specific_validations.py
```diff
@@ -11,14 +11,9 @@
     validate_programme_risks,
     validate_project_risks,
     validate_psi_funding_gap,
-    validate_sign_off,
 )
 
 # flake8: noqa
```
cyrusdobbs, line 16: can remove this as well

tests/validation_tests/test_pre_checks.py
```diff
@@ -209,6 +249,138 @@ def valid_workbook_round_four():
...
         "8 - Review & Sign-Off": pd.DataFrame(
+            {
+                0: ["", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", ""],
+                1: [
+                    "",
+                    "",
+                    "",
+                    "",
+                    "",
+                    "",
```
cyrusdobbs, line 260: Do we want these to be pulled out into a `mock_data.py` module again? This clutters the test module quite a lot

core/validation/failures.py
```diff
@@ -465,6 +465,21 @@ def to_message(self) -> tuple[str | None, str | None, str]:
...
+
+    tab: str
+    section: str
+    message: str
+
+    def __str__(self):
+        pass
+
+    def to_message(self) -> tuple[str, str, str]:
+        return self.tab, self.section, self.message
```
cyrusdobbs, line 480: I think we can renamed this to `GenericPreTransformationFailure` or similar as it's not specific to that failure in anyway other than by name

core/validation/initial_check.py
```diff
@@ -22,6 +23,9 @@ def validate_before_transformation(workbook: dict[str, pd.DataFrame], reporting_
         workbook=workbook, reporting_round=reporting_round, place_names=place_names
     )
     file_validation_failures = pre_transformation_check(pre_transformation_details)
+    if reporting_round == 4:
+        if sign_off_failures := validate_sign_off(workbook):
+            file_validation_failures.extend(sign_off_failures)
```
cyrusdobbs, line 28: Bit nit picky but..

Could this be rewritten in this style:
```
if reporting_round == 4:
    sign_off_failures = validate_sign_off(workbook):
    file_validation_failures = [*file_validation_failures , *sign_off_failures]
```
for consistency with:
https://github.com/communitiesuk/funding-service-design-post-award-data-store/blob/045ab1caec8cb92a895d8db8b7d12f710f4d44eb/core/validation/__init__.py#L22-L24

cyrusdobbs (changes requested):

core/validation/failures.py
Joel-Pearce: Reflecting on it, PreTransFormationFailures don't make use of the tab or section, but for a sign-off failure we need both for the validate message to be more specific to the user. We need to collect information about the tab, section, missing_value, and sign_off_officer, and pass these to the message string, which probably means this particular failure isn't reusable for other generic pre-transformation failures.

I've renamed the failure SignOffFailure, and refactored it such that it will utilise the tab and section but still be treated as a PreTransFormationFailure.

core/validation/initial_check.py
Joel-Pearce: Yes, provided validate_sign_off() returns an empty list rather than None is the case where no error is raised. I've refactored the function so this is the case.

core/validation/failures.py
```diff
@@ -465,6 +465,27 @@ def to_message(self) -> tuple[str | None, str | None, str]:
...
+    sign_off_officer: str
+
+    def __str__(self):
+        pass
+
+    def to_message(self) -> tuple[None, None, str]:
+        message = (
+            f"In the tab '{self.tab}' you must fill out the "
+            f"'{self.missing_value}' for '{self.section}'. "
+            f"You need to get sign off from {self.sign_off_officer}"
```
cyrusdobbs, line 484: Probably worth getting design's (content) input on this message.

cyrusdobbs (approved):
Looks good. Might be best to get design to approve the message and also get this QA'd before merging?

PR #270, opened 2023-10-03
[SMD-171] ingest Round 4 template as R4 into the db

cyrusdobbs:
https://dluhcdigital.atlassian.net/browse/SMD-171

### Change description
- modifies the submission details extraction function to be dynamic

This was previously hardcoded to Round 3. We can now trust the submission contents to determine the Pre transformation validation validates that the submission is the correct version and has the correct reporting period selected.

### How to test


core/extraction/towns_fund_round_four.py
```diff
@@ -23,8 +24,7 @@ def ingest_round_four_data_towns_fund(df_ingest: pd.DataFrame) -> dict[str, pd.D
     """
 
     towns_fund_extracted = dict()
-    # TODO: SMD-171 - update submission details extraction to support Round 4, currently hardcoded as "Round 3"
-    towns_fund_extracted["Submission_Ref"] = r3.extract_submission_details(df_ingest["1 - Start Here"].iloc[4, 1])
+    towns_fund_extracted["Submission_Ref"] = common.get_submission_details(4)
```
nsnape, line 27: not sure if it's strictly necessary, but it might be useful to invoke this param with it's kwarg? So it's more clear from glancing at it why/what it is hard-coded for?
any thoughts?

core/extraction/towns_fund_round_three.py
```diff
@@ -34,7 +34,7 @@ def ingest_round_three_data_towns_fund(df_ingest: pd.DataFrame) -> Dict[str, pd.
     """
 
     towns_fund_extracted = dict()
-    towns_fund_extracted["Submission_Ref"] = extract_submission_details(df_ingest["1 - Start Here"].iloc[4, 1])
+    towns_fund_extracted["Submission_Ref"] = common.get_submission_details(3)
```
nsnape, line 37: as per comment in round 4

core/extraction/common.py
```diff
@@ -0,0 +1,40 @@
+from datetime import datetime
```
nsnape, line 1: Just wondering, does it make sense to use this opportunity to move ALL common extraction methods from Round 3 to here? Since Round 4 imports a lot of R3 methods, but really lots of them are round agnostic? I think we chatted about this before in a similar bit of work, but it might make more sense to do it now?

nsnape (changes requested):

core/extraction/common.py
cyrusdobbs: Yeah I see what you mean.. They're actually all common in that sense. The only difference is the top level function that calls them with the particular parameters... I think this should be a separate ticket to be honest.

nsnape (approved):

PR #272, opened 2023-10-04
[SMD-220]: retain original row indexes

Joel-Pearce:
Retains the original indexes of the DataFrames in df_ingest. Upon ingest, the indexes are decremented by two, and so when we relay the original_index information to validation messages it will have to be incremented by two. 

Points of note:
- Footfall outcomes are more difficult to retrieve specific row numbers for as their values span multiple rows in the original spreadsheet. They are therefore not accounted for in this PR.
- The original index is not retained for Project Details, but is instead added on after the transformation occurs. We know that Project 1 will always begin at line 27 of the spreadsheet, and that each project is incremented by 1, and can set the index accordingly. 

### DIFF
--- a/core/extraction/towns_fund_round_four.py
+++ b/core/extraction/towns_fund_round_four.py
@@ -15,7 +15,7 @@ import core.extraction.towns_fund_round_three as r3
 from core.extraction import common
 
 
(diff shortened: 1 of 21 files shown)


nsnape (changes requested):
Looking good. I think we can remove all the old index drop rows that are now commented out. I commented on a couple but there are some in most extraction functions. Ta

nsnape (approved):

PR #273, opened 2023-10-04
[SMD-189] set up s3 client

cyrusdobbs:
https://dluhcdigital.atlassian.net/browse/SMD-189

### Change description
Sets up S3 for use throughout the application.

### How to test


tests/db_tests/test_aws_access.py
```diff
@@ -28,7 +20,7 @@ def test_list_buckets():
 
 @pytest.mark.skip(reason="Test should only be run locally in dev environment")
```
nsnape, line 21: is this an actual local test we want to run as standard? Or just some debugging tool? If the former we could not skip it, but check the env and only run if local dev? Otherwise just pass. Is that feasible?

config/envs/default.py
```diff
@@ -15,3 +15,9 @@ class DefaultConfig(object):
         FLASK_ROOT / "tests" / "controller_tests" / "resources" / "Post_transform_EXAMPLE_data.xlsx"
     )
     ENABLE_PROFILER = os.getenv("ENABLE_PROFILER")
+
+    AWS_REGION = os.getenv("AWS_REGION")
+    AWS_ACCESS_KEY_ID = os.getenv("AWS_ACCESS_KEY_ID")
+    AWS_SECRET_ACCESS_KEY = os.getenv("AWS_SECRET_ACCESS_KEY")
+    AWS_ENDPOINT_OVERRIDE = os.getenv("AWS_ENDPOINT_OVERRIDE")
+    AWS_S3_BUCKET_FAILED_FILES = os.getenv("AWS_S3_BUCKET_FAILED_FILES")
```
gidsg, line 23: A lot of these won't be used on AWS, can we move them into `development.py` env config instead?

core/aws.py
```diff
@@ -0,0 +1,36 @@
...
+from config import Config
+
+if Config.AWS_ACCESS_KEY_ID and Config.AWS_SECRET_ACCESS_KEY:
+    _S3_CLIENT = client(
+        "s3",
+        aws_access_key_id=Config.AWS_ACCESS_KEY_ID,
+        aws_secret_access_key=Config.AWS_SECRET_ACCESS_KEY,
+        region_name=Config.AWS_REGION,
+        endpoint_url=Config.AWS_ENDPOINT_OVERRIDE,
+    )
```
gidsg, line 16: Can we remove this block? We should never be passing access keys and secret keys in on AWS.

pytest.ini
```diff
@@ -1,3 +1,8 @@
 [pytest]
 env =
     FLASK_ENV=unit_test
+    AWS_REGION=eu-central-1
+    AWS_ACCESS_KEY_ID=test
+    AWS_SECRET_ACCESS_KEY=test
+    AWS_ENDPOINT_OVERRIDE=http://127.0.0.1:4566/
+    AWS_S3_BUCKET_FAILED_FILES=data-store-failed-files-dev
```
gidsg, line 8: Could we move this stuff into `unit_test.py` instead?

core/aws.py
```diff
@@ -0,0 +1,36 @@
...
+else:
+    # boto gets access keys from the environment directly in AWS
+    _S3_CLIENT = client(
+        "s3",
+        region_name=Config.AWS_REGION,
+        endpoint_url=Config.AWS_ENDPOINT_OVERRIDE,
+    )
+
+
+def upload_file(file: FileIO, bucket: str, object_name: str):
```
gidsg, line 26: Can `BUCKET_NAME` be an env var? If there are multiple buckets they can be separate env vars.

core/aws.py
```diff
@@ -0,0 +1,36 @@
...
+    _S3_CLIENT = client(
+        "s3",
+        aws_access_key_id=Config.AWS_ACCESS_KEY_ID,
+        aws_secret_access_key=Config.AWS_SECRET_ACCESS_KEY,
+        region_name=Config.AWS_REGION,
+        endpoint_url=Config.AWS_ENDPOINT_OVERRIDE,
+    )
+else:
+    # boto gets access keys from the environment directly in AWS
+    _S3_CLIENT = client(
```
gidsg, line 19: Will this break things if people are running locally without the docker runner? Should we put a separate env var to enable this functionality?

core/aws.py
cyrusdobbs: It is already an envar currently. We have 2 buckets configured in AWS at the moment linked to the data-store

core/aws.py
cyrusdobbs: It won't break - an exception will just be caught and logged

tests/db_tests/test_aws_access.py
cyrusdobbs: More of a debugging tool I think. Otherwise we will have to have the localstack container running every time we run the tests locally in order to pass

gidsg (approved):

PR #277, opened 2023-10-05
Map table and column names to original cell letters

Connor-OS:
Add a mapping to map table and column to original excel column letter. This can be used insed to_message functions to get original letters to be displayed this should be combined with row indexes in a seperate PR, as well as add adequate test coverage for this.

### Change description

### How to test


Joel-Pearce:
There's a bit of a convoluted way to test this by looking at the position of a given column, taking its index, and converting it into the letter it should represent.

Should we add the column letters to the failure messages when we can't verify, test, or use them yet?

Otherwise, looking good, I can't see any mistakes in the mapping, I will manually check again.

core/const.py
```diff
@@ -797,3 +797,62 @@ class ITLRegion(StrEnum):
     "Pre-Payment)",
     "Towns Fund RDEL Payment which is being utilised on TF project related activity",
 ]
+
+TABLE_AND_COLUMN_TO_ORIGINAL_COLUMN_LETTER = {
```
nsnape, line 801: Can you add a comment saying this is Towns Fund, and  Round 4 onwards only (I assume that's what it is anyway as ProjectProgress has the new columns?)?

On that note, is this only implemented on R4? Just wondering what'd happen if we ever need to re-ingest Round 3again.

nsnape: would a more descriptive name for this be something like TOWNS_FUND_REPORTING_FORM_COLUMN_MAPPING ?

nsnape (changes requested):

Joel-Pearce:
> There's a bit of a convoluted way to test this by looking at the position of a given column, taking its index, and converting it into the letter it should represent.
> 
> Should we add the column letters to the failure messages when we can't verify, test, or use them yet?
> 
> Otherwise, looking good, I can't see any mistakes in the mapping, I will manually check again.

On second thought, there's no point testing the actual data structure, but we could separate out just putting in the mapping of values and changing of the validation messages to include their original column values (which will require changing the tests).

core/const.py
Connor-OS: If we tried to reingest round 3 the column letters in Project progress would be wrong. We should think if we want to cover user centric messaging for round 3 or not maybe

Joel-Pearce: We don't need to cover use centric messaging for round 3 because users will never upload for it, and if we have to re-upload the files (because of some data quality issue or something) then we'll do it ourselves

core/const.py
```diff
@@ -797,3 +797,62 @@ class ITLRegion(StrEnum):
...
+    "Place Details": {"Question": "C", "Indicator": "D", "Answer": "E"},
+    "Project Details": {
+        "Project name": "E",
+        "Primary Intervention Theme": "F",
+        "Single or Multiple Locations": "G",
+        "Locations": "H/K",
+        "Postcodes": "H/K",
+        "Lat/Long": "I/L",
```
Joel-Pearce, line 809: I feel like this might breed some confusion as when I first read "H/K" I read it as "H-K", but it's fairly minor. We can leave it as is for now and get some design input.

Connor-OS:
> ).

Good point, I think I will reveret some of this PR, just to contain the mapping and leave how we use the maaping for a future PR

core/const.py
Connor-OS: In that case maybe we should turn of user centered messaging for rounds prior to the current, because it's no use maintaining messaging for reporting rounds that are never going to be used again

Joel-Pearce: We've already updated R1, R2, and R3 multiple times, and there's no reason they should be uploaded again; in the case that they are, and that there are errors, then we will need to revert to the original validation messages as they are more useful for developer debugging.

Joel-Pearce (approved):
Nice one, approving

core/const.py
nsnape: remember we will have to re-import R1 - 3 at least once, since we have all that erroneous data still in from the conditional formatting.

nsnape (approved):

PR #281, opened 2023-10-06
Feature/smd 87 funding spent vs funding allocated validation

Connor-OS:
PR to add a r4 specific validation of funding spent vs funding allocated pulled from a seperate csv stored on s3

Still need to properly implement unit tests by mocking the s3 connection. so only a draft PR for now.

### Change description

### How to test

### DIFF
--- a/core/validation/specific_validations/towns_fund_round_four.py
+++ b/core/validation/specific_validations/towns_fund_round_four.py
@@ -3,9 +3,12 @@ from dataclasses import dataclass
 import numpy as np
 import pandas as pd
 
+from config.envs.default import DefaultConfig
+from core.aws import get_file
 from core.const import (
     INTERNAL_TABLE_TO_FORM_TAB,
     PRE_DEFINED_FUNDING_SOURCES,
+    TF_FUNDING_ALLOCATED_FILE,
     FundingSourceCategoryEnum,
     MultiplicityEnum,
     StatusEnum,
@@ -32,6 +35,7 @@ def validate(workbook: dict[str, pd.DataFrame]) -> list["TownsFundRoundFourValid
         validate_psi_funding_gap,
         validate_locations,
         validate_leading_factor_of_delay,
+        validate_funding_spent,
     )
 
     validation_failures = []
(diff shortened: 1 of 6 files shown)


core/validation/specific_validations/towns_fund_round_four.py
```diff
@@ -268,6 +270,86 @@ def validate_leading_factor_of_delay(
...
+            & (funding_df["Funding Source Type"] == "Towns Fund")
+            & ~(funding_df["Funding Source Name"].str.contains(r"contractually committed"))
+        ]
+        .sum()
+    )
+    # Business logic here is taken from spreadsheet 4a - Funding Profile Z45 for grand total expenditure
+    return funding_spent
+
+
 @dataclass
```
Connor-OS, line 353: Does a function like this need to go in utils? It is only used onece and put here because it is a massive piec of logic. Could add some tests just for this although they are covered in the parent function tests

core/validation/specific_validations/towns_fund_round_four.py
```diff
@@ -268,6 +270,86 @@ def validate_leading_factor_of_delay(
...
+    For Town deal funds this is done on a per-project basis whereas for future High Street Funds this is done
+    per-programme leading to slightly different logics for both.
+
+    :param workbook: A dictionary where keys are sheet names and values are pandas
+                     DataFrames representing each sheet in the Round 4 submission.
+    :return: ValidationErrors
+    """
+    # Pull programme and project indexes from the workbook
+    programme_id = workbook["Programme_Ref"]["Programme ID"][0]
+    fund_type = programme_id[:2]
```
cyrusdobbs, line 286: I think using `workbook["Programme_Ref"]["FundType_ID"]` makes more sense here

core/validation/specific_validations/towns_fund_round_four.py
```diff
@@ -268,6 +270,86 @@ def validate_leading_factor_of_delay(
...
+                funding_spent_failures.append(
+                    TownsFundRoundFourValidationFailure(
+                        tab="Funding Profiles",
+                        section=f"Project Funding Profiles - Project {get_project_number(idx)}",
+                        message=(
+                            f"[Template message] You spent {funding_spent[idx]} "
+                            f"but were only allocated {funding_allocated[idx]}"
+                        ),
+                    )
+                )
```
cyrusdobbs, line 327: There's a fair bit of duplication here. 

Can we set `ids_to_check` depending on `fund_type` and then reuse the same code to check and capture any failures?

I suppose section is depended on `fund_type` as well but we could also add that as an inline if else?

core/validation/specific_validations/towns_fund_round_four.py
cyrusdobbs: I think it's too specific to go in utils so is fine here. I agree that the parent function tests should cover this fine as well

core/validation/specific_validations/towns_fund_round_four.py
```diff
@@ -268,6 +270,86 @@ def validate_leading_factor_of_delay(
...
+    fund_type = programme_id[:2]
+    project_ids = workbook["Project Details"]["Project ID"]
+    funding_df = workbook["Funding"]
+
+    # pull funding spent for individual projects and programme wide total
+    funding_spent = {project: spend_per_project(funding_df, project) for project in project_ids}
+    funding_spent[programme_id] = sum(funding_spent.values())
+
+    allocated_file = s3.get_file("data-store-file-assets-dev", f"{fund_type}-grant-awarded.csv")
+    funding_allocated = pd.read_csv(allocated_file, index_col="Index Code")["Grant Awarded"]
```
cyrusdobbs, line 295: Can we just do this once during config instantiation and then use it here

cyrusdobbs (changes requested):
Looks good, just a few comments on the code. I want to double check the business logic but will do that soon.

core/validation/specific_validations/towns_fund_round_four.py
Connor-OS: We decided to load this on validation instead as we didn't want potentially dynamic functions in const.py

core/validation/specific_validations/towns_fund_round_four.py
```diff
@@ -296,6 +300,89 @@ def validate_leading_factor_of_delay(
...
+        validate_by = "project"
+        ids_to_check = project_ids.tolist()
+
+    for idx in ids_to_check:
+        if funding_spent[idx] > funding_allocated[idx]:
+            funding_spent_failures.append(
+                TownsFundRoundFourValidationFailure(
+                    sheet="Funding Profiles",
+                    section="Project Funding Profiles"
+                    + ("" if fund_type == "HS" else f" - Project {get_project_number(idx)}"),
```
cyrusdobbs, line 343: Niggly thing but I think it's more readable if we use `validate_by` here instead of HS? Think it's more intuitive to see the project number being added to the string if `validate_by` == "project" - not immediately clear why fund_type would decide that

Connor-OS: agree, that helps with readability a lot. I've added this now

core/validation/specific_validations/towns_fund_round_four.py
```diff
@@ -296,6 +300,89 @@ def validate_leading_factor_of_delay(
...
+        .sum()
+    )
+    # Business logic here is taken from spreadsheet 4a - Funding Profile Z45 for grand total expenditure
+    return funding_spent
+
+
+def get_allocated_funding():
+    return pd.read_csv(
+        get_file(DefaultConfig.AWS_S3_BUCKET_FILE_ASSETS, TF_FUNDING_ALLOCATED_FILE), index_col="Index Code"
+    )["Grant Awarded"]
```
cyrusdobbs, line 383: My issue with this was that it's then done every request. This almost certainly won't make a practical difference at our level of scale. It just seems like better practice to load this into memory once per application initialisation and then fetch from there instead of repeatedly doing this I/O. I feel like there has to be clean way to implement this.. 

What do you think @Connor-OS @nsnape?

cyrusdobbs: Could we just instantiate it as a constant at the top of this module from this function and use that within `validate_funding_spent`? The way we mock wouldn't have to change

Connor-OS: I I think the way I spoke about it with Noel earlier was that it can't be a constant if it is being dynamically read, it is more of a config prehaps, but that has other issues.

When do you think it will be reused? When users reupload after fixing errors? I guess that will only be 1 or 2 times on average.

cyrusdobbs: This way we will be reading the file from S3 every single request that is made. The reuse would be the lifetime of the application - so a fair bit. Perhaps constant isn't the correct term, I just mean a variable that lasts the lifetime of the application, it could be stored in this module for instance.

Connor-OS: We've looked into doing this as it would be more effecient to load this data just once on application initialisation. However it has made things to complex to mock this function call. We've deprioritised the issue for now and will add a ticket to the backlog to be picked up after delivery

cyrusdobbs (approved):
Good stuff - LGTM

PR #284, opened 2023-10-10
[BAU] upversion accepted Round 4 template

cyrusdobbs:
### Change description
- modifies the pre-validation check to accept V4.2 instead of V4 of the R4 template
- refactor of pre check mock data so that fixtures are accessed via pytest rather than importing like normal modules

### How to test


tests/validation_tests/test_pre_checks.py
```diff
@@ -1,27 +1,29 @@
...
+    mocker.patch(
+        "core.validation.initial_check.GET_FORM_VERSION_AND_REPORTING_PERIOD",
+        {
+            3: ("Town Deals and Future High Streets Fund Reporting Template (v3.0)", "1 October 2022 to 31 March 2023"),
+            4: (
+                "Town Deals and Future High Streets Fund Reporting Template (v4.0)",
+                "1 April 2023 to 30 September 2023",
+            ),
+        },
+    )
```
Joel-Pearce, line 26: Why is this being mocked? Can't the updated sheet version be added into the R4 valid workbook in mock_data?

core/const.py
```diff
@@ -784,7 +784,7 @@ class ITLRegion(StrEnum):
 # form version and reporting period for different rounds
 GET_FORM_VERSION_AND_REPORTING_PERIOD = {
     3: ("Town Deals and Future High Streets Fund Reporting Template (v3.0)", "1 October 2022 to 31 March 2023"),
-    4: ("Town Deals and Future High Streets Fund Reporting Template (v4.0)", "1 April 2023 to 30 September 2023"),
+    4: ("Town Deals and Future High Streets Fund Reporting Template (v4.2)", "1 April 2023 to 30 September 2023"),
```
Joel-Pearce, line 787: This also needs to be reflected in PRETRANSFORMATION_FAILURE_MESSAGE_BANK. The messages are currently specific to R4, and operate under the presumption of a static reporting period and form version.

Joel-Pearce (changes requested):

tests/validation_tests/test_pre_checks.py
cyrusdobbs: If we mock it here then we don't need to keep changing it if we modify the version. I don't think it adds to the quality of these tests to keep it coupled to the specifics of these string values

core/const.py
cyrusdobbs: Ah - thanks will change this

Joel-Pearce (approved):
LGTM

PR #286, opened 2023-10-10
[SMD-243] incorporate cell numbers into round 4 specific validations

Connor-OS:
incourperate cell numbers into all round 4 specific validations as well as making sure all validations pull through column information. 

I've also created a utils function that can construct a list of cell indexes from cell letter and numbers to be returned in the failure payload.

### Change description

### How to test


core/util.py
```diff
@@ -65,3 +65,20 @@ def get_project_number(project_id):
...
+    """Constructs the index of an error from the column and rows it occurred in.
+
+    :param section: the internal table name where the error occurred
+    :param column: the internal table name where the error occurred
+    :param rows: tuple or int of row indexes where the error occurred
+    :return: indexes tuple of constructed letter and number indexes
+    """
+
+    column_letter = TABLE_AND_COLUMN_TO_ORIGINAL_COLUMN_LETTER[section][column]
+    if type(rows) is int:
```
cyrusdobbs, line 80: it's more pythonic to use `isinstance(rows, int)`

core/util.py
```diff
@@ -65,3 +65,20 @@ def get_project_number(project_id):
...
+        indexes = column_letter + str(rows)
+    else:
+        indexes = ", ".join([column_letter + str(row) for row in rows])
+    return indexes
```
cyrusdobbs, line 84: Can we have some unit tests for this function

core/validation/specific_validations/towns_fund_round_four.py
```diff
@@ -124,17 +132,19 @@ def validate_funding_profiles_funding_source(
     if len(invalid_rows) > 0:
         return [
             TownsFundRoundFourValidationFailure(
-                tab="Funding Profiles",
+                sheet="Funding",
                 section=f"Project Funding Profiles - Project {get_project_number(row['Project ID'])}",
+                column="Funding Source Type",
                 message=f'For column "Funding Source", you have entered "{row["Funding Source Type"]}" which isn\'t '
                 f"correct. You must select an option from the list provided",
+                indexes=row.name,
```
cyrusdobbs, line 140: How come we use `name` for this?

cyrusdobbs (changes requested):

core/util.py
```diff
@@ -65,3 +65,20 @@ def get_project_number(project_id):
...
+    :param column: the internal table name where the error occurred
+    :param rows: tuple or int of row indexes where the error occurred
+    :return: indexes tuple of constructed letter and number indexes
+    """
+
+    column_letter = TABLE_AND_COLUMN_TO_ORIGINAL_COLUMN_LETTER[section][column]
+    if type(rows) is int:
+        indexes = column_letter + str(rows)
+    else:
+        indexes = ", ".join([column_letter + str(row) for row in rows])
```
Joel-Pearce, line 83: We shouldn't use arrays in order to store the indexes. These unit tests currently pass because all of the arrays only contain one value, but if they contain more than two the tests will fail. You can read more about it here: https://stackoverflow.com/questions/10062954/valueerror-the-truth-value-of-an-array-with-more-than-one-element-is-ambiguous

We can use tuples instead, which don't encounter this issue

Joel-Pearce (changes requested):
Looking good, but the testing on this probably needs to be beefed-up in some scenarios (i.e. when len(indexes) > 1) the tests will fail. Let's account for this by having scenarios where indexes will contain multiple values.

Also we need to test that the original index is retained, and not defaulted to a 0-index. To do this I think all we have to do is add arbitrary index numbers to the test data (that or just by default increment the original index by a certain number and check to make sure it has not somehow been lost by the time the failure is raised).

core/validation/specific_validations/towns_fund_round_four.py
Joel-Pearce: I think it's because row.index will not return an integer of the index but instead a series of index labels, but row.name returns whatever the value in the index is for that row. I think it should therefore also work on the index isn't 0-indexed, but perhaps we should test for when it is not as well in this case?

core/util.py
Connor-OS: To push back on this a little, I thought a good rule of thumb is to use a python list for homogeneous data and since the indexes will all be ints it seems to make sense.

Also I can add more than one value to the index lists and still have them pass. See updated test_validate_locations_failure.

The link you sent seems to be talking more about numpy arrays. Whereas python lists don't have the same behaviour I belive. Maybe needs more thought we should choose one, list or tuple and stick with that.

Joel-Pearce: I raised this because for the non-R4 specific validation tests, including indexes as a list was having tests fails with a ValueError when I tried to compare them. The issue seemed to be that my arrays were being treated as numpy arrays, and I thought this might end up being the case with all of our arrays of indexes. 

Investigating, it looks like this comes from lines of code like:
```
failed_indexes = sheet.index[(pd.isna(sheet[column]) | (sheet[column].astype("string") == ""))]
```
but thinking about it, the lists can be converted from numpy arrays to Python arrays with tolist(). 

Provided we've got more than one value for indexes (where we expect multiple) in the tests then I don't think it matters. This PR precedes mine so we can go with lists :)

Connor-OS: Okay, no problem. I'll go with lists then for this case and try and address the other comments here so that we can get thisone in soon

core/util.py
Connor-OS: added to test_failures for want of a better place to put them. As that seems like this function will be used the most

Joel-Pearce (approved):
LGTM, nice job!

cyrusdobbs (changes requested):
just a quick one

core/util.py
```diff
@@ -65,3 +65,17 @@ def get_project_number(project_id):
...
+
+    :param section: the internal table name where the error occurred
+    :param column: the internal table name where the error occurred
+    :param rows: tuple or int of row indexes where the error occurred
+    :return: indexes tuple of constructed letter and number indexes
+    """
+
+    column_letter = TABLE_AND_COLUMN_TO_ORIGINAL_COLUMN_LETTER[section][column]
+    indexes = ", ".join([column_letter + str(row) for row in rows])
+    return indexes
```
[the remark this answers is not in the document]
cyrusdobbs: One solution is to define the mapping as a set of strings to be formatted like:
```
TABLE_AND_COLUMN_TO_ORIGINAL_COLUMN_LETTER = {
    "Place Details": {"Question": "C{i}", "Indicator": "D{i}", "Answer": "E{i}"},
    "Project Details": {
        "Project name": "E{i}",
        "Primary Intervention Theme": "F{i}",
        "Single or Multiple Locations": "G{i}",
        "Locations": "H{i}/K{i}",
        "Postcodes": "H{i}/K{i}",
        "Lat/Long": "I{i}/L{i}",
        "GIS Provided": "J{i}-K{i}",  # I've changed this range, its fake to give an example of this
    },
```

where we can then do:
```
column_letter = TABLE_AND_COLUMN_TO_ORIGINAL_COLUMN_LETTER[section][column]
indexes = ", ".join([column_letter.format(i=row) for row in rows])
```

cyrusdobbs (changes requested):
What do you think?

core/util.py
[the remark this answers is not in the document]
Connor-OS: I think this is a definate improvement so will add this as well.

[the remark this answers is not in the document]
Connor-OS: I've added this now and updated the tests for it

core/util.py
```diff
@@ -65,3 +65,17 @@ def get_project_number(project_id):
...
+    """Constructs the index of an error from the column and rows it occurred in.
+
+    :param section: the internal table name where the error occurred
+    :param column: the internal table name where the error occurred
+    :param rows: tuple or int of row indexes where the error occurred
+    :return: indexes tuple of constructed letter and number indexes
+    """
+
+    column_letter = TABLE_AND_COLUMN_TO_ORIGINAL_COLUMN_LETTER[section][column]
+    indexes = ", ".join([column_letter.format(i=row + 2) for row in rows])
```
cyrusdobbs, line 80: Could the reason behind this increment of 2 be documented in the docstring and/or a comment please?

core/validation/specific_validations/towns_fund_round_four.py
```diff
@@ -153,19 +164,22 @@ def validate_locations(workbook: dict[str, pd.DataFrame]) -> list["TownsFundRoun
     single_rows = project_details_df[project_details_df["Single or Multiple Locations"] == MultiplicityEnum.SINGLE]
     multiple_rows = project_details_df[project_details_df["Single or Multiple Locations"] == MultiplicityEnum.MULTIPLE]
 
-    # tuples of (column_data, column_name) to validate empty cells on
+    # tuples of (column, column_data, column_name) to validate empty cells on
```
cyrusdobbs, line 167: can this naming be improved? what is `column` vs `column_name`? Do we need both? Seems convoluted

core/validation/specific_validations/towns_fund_round_four.py
```diff
@@ -272,12 +296,16 @@ def validate_leading_factor_of_delay(
 class TownsFundRoundFourValidationFailure(ValidationFailure):
     """Generic Towns Fund Round 4 Validation Failure."""
 
-    tab: str
+    sheet: str
     section: str
+    column: str
     message: str
+    indexes: list[int] | int
```
cyrusdobbs, line 303: I think `row_indexes` may be clearer - what do you think?

Connor-OS: I prefer row indexes, we should make sure to keep this name consistent accross generic and round 4 validations

core/validation/specific_validations/towns_fund_round_four.py
Connor-OS: I agree, the two here mean internal table column and form column. we don'thave an existing mapping that works perfectly for this use case but I can try to make it more clear

core/util.py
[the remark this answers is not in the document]
cyrusdobbs: nice one 👍

core/validation/specific_validations/towns_fund_round_four.py
cyrusdobbs: yep agreed

cyrusdobbs (approved):
LGTM 👍

PR #290, opened 2023-10-12
[SMD-236]: put indexes in validation messages for non-R4 specific errors

Joel-Pearce:
Relay the indexes of where an error has occurred to the validation failures for non R4-specific validation failures. Excludes WrongTypeFailure.

### DIFF
--- a/core/extraction/towns_fund_round_three.py
+++ b/core/extraction/towns_fund_round_three.py
@@ -862,9 +862,11 @@ def extract_outcomes(df_input: pd.DataFrame, project_lookup: dict, programme_id:
         raise ValidationError(
             [
                 vf.InvalidOutcomeProjectFailure(
-                    invalid_project=project, section="Outcome Indicators (excluding footfall)"
+                    invalid_project=row["Relevant project(s)"],
+                    section="Outcome Indicators (excluding footfall)",
+                    row_indexes=[idx],
                 )
-                for project in invalid_projects
+                for idx, row in outcomes_df.loc[outcomes_df["Relevant project(s)"].isin(invalid_projects)].iterrows()
             ]
         )
     # Drop rows with Section header selected as the outcome from dropdown on form - This is not a valid outcome option.
@@ -959,7 +961,8 @@ def extract_footfall_outcomes(df_input: pd.DataFrame, project_lookup: dict, prog
     if invalid_projects := relevant_projects - set(project_lookup.keys()):
         raise ValidationError(
             [
-                vf.InvalidOutcomeProjectFailure(invalid_project=project, section="Footfall Indicator")
+                vf.InvalidOutcomeProjectFailure(invalid_project=project, section="Footfall Indicator", row_indexes=None)
+                # TODO: add index for footfalls
                 for project in invalid_projects
             ]
         )
--- a/core/validation/failures.py
+++ b/core/validation/failures.py
@@ -135,6 +135,7 @@ class NonUniqueCompositeKeyFailure(ValidationFailure):
     sheet: str
     cols: tuple
     row: list
+    row_indexes: list[int]
 
     def __str__(self):
         """Method to get the string representation of the non-unique-composite_key failure."""
@@ -197,7 +198,7 @@ class WrongTypeFailure(ValidationFailure):
     column: str
     expected_type: str
     actual_type: str
-    index: list[int]
+    row_indexes: list[int]
 
     def __str__(self):
         """Method to get the string representation of the wrong type failure."""
@@ -259,7 +260,7 @@ class OrphanedRowFailure(ValidationFailure):
         """Method to get the string representation of the orphaned row failure."""
         return (
             f'Orphaned Row Failure: Sheet "{self.sheet}" Column "{self.foreign_key}" '
-            f"Row {self.row + 2} "  # +2 for Excel 1-index and table header row
+            f"Row {self.row} "
             f'Value "{self.fk_value}" not in parent table '
             f'"{self.parent_table}" where PK "{self.parent_pk}"'
         )
@@ -274,7 +275,7 @@ class InvalidEnumValueFailure(ValidationFailure):
 
     sheet: str
     column: str
-    row: int
+    row_indexes: list[int]
     row_values: tuple
     value: Any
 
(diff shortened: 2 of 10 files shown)


core/extraction/towns_fund_round_three.py
```diff
@@ -957,7 +959,8 @@ def extract_footfall_outcomes(df_input: pd.DataFrame, project_lookup: dict, prog
     if invalid_projects := relevant_projects - set(project_lookup.keys()):
         raise ValidationError(
             [
-                vf.InvalidOutcomeProjectFailure(invalid_project=project, section="Footfall Indicator")
+                vf.InvalidOutcomeProjectFailure(invalid_project=project, section="Footfall Indicator", index=None)
+                # TODO: add index for footfalls
```
nsnape, line 963: is this TODO addressed in any PR?

core/validation/validate.py
```diff
@@ -304,10 +308,13 @@ def validate_nullable(
```
nsnape, line 308: please can you add a docstring to this function

core/validation/validate.py
```diff
@@ -304,10 +308,13 @@ def validate_nullable(
 
     for column in sheet.columns:
         if column in non_nullable and (pd.isna(sheet[column]).any() | (sheet[column].astype("string") == "").any()):
+            failed_indexes = sheet.index[(pd.isna(sheet[column]) | (sheet[column].astype("string") == ""))]
+
             non_nullable_constraint_failure.append(
                 vf.NonNullableConstraintFailure(
                     sheet=sheet_name,
                     column=column,
+                    indexes=failed_indexes.asi8.tolist(),
```
nsnape, line 317: I think the .asi8 might be superfluous here, tolist() should handle casting to int without it?

nsnape (changes requested):

core/validation/validate.py
```diff
@@ -271,6 +272,7 @@ def validate_enums(
     :return: A list of InvalidEnumValueFailure objects for any rows with values outside
              the set of valid enum values.
     """
+    workbook[sheet_name]["original_index"] = workbook[sheet_name].index
     sheet = workbook[sheet_name].reset_index(drop=True)
```
Connor-OS, line 276: You can theoretically combine these on one line 
```sheet = workbook[sheet_name].reset_index()```
will insert the onld index as a column called Index

File: core/validation/failures.py [hunk 5]
```diff
@@ -421,6 +423,7 @@ class InvalidOutcomeProjectFailure(ValidationFailure):
 
     invalid_project: str
     section: str
+    index: list[int]
```
Connor-OS, line 426: Maybe just my preferance but prehaps we should keep the naming convention of these indexes the same throughout. Prehaps row_indexes is more descriptive. What do you think?

Connor-OS (changes requested):
Looking good, almost there

cyrusdobbs (changes requested):

core/validation/failures.py
Joel-Pearce: If they already have a standardised name from the R4-specific PR then I'm happy to change to that, I'll take a look in a second. If not, then row_indexes or indexes are both fine.

core/validation/validate.py
```diff
@@ -271,6 +272,7 @@ def validate_enums(
     :return: A list of InvalidEnumValueFailure objects for any rows with values outside
              the set of valid enum values.
     """
+    workbook[sheet_name]["original_index"] = workbook[sheet_name].index
```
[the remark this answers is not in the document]
Joel-Pearce: No, it will cause an error because for some of our DataFrames the indexes will be replicated (where, for instance, a melt has happened), and  ```invalid_value = sheet[column][row_idx]``` will no longer return an individual value, but instead a series of values.

tests/validation_tests/test_validate.py
```diff
@@ -68,6 +70,10 @@ def valid_workbook_and_schema():
         ),
     }
 
+    # Incrementing indexes by 5 for each DataFrame
+    for sheet_name, df in workbook.items():
+        workbook[sheet_name] = df.set_index(df.index + 5)
+
```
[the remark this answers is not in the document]
Joel-Pearce: To make sure the validation methods don't default the indexes to being 0-indexed and lose our original indexes (for example, in validate_enums we have to do some more index manipulation). This way we know for sure that whatever indexes are passed to validation, the same indexes will be returned.

core/validation/failures.py
```diff
@@ -273,15 +274,15 @@ class InvalidEnumValueFailure(ValidationFailure):
...
+    row: list[int]
     row_values: tuple
     value: Any
 
     def __str__(self):
         """Method to get the string representation of the invalid enum value failure."""
         return (
             f'Enum Value Failure: Sheet "{self.sheet}" Column "{self.column}" '
-            f"Row {self.row + 2} "  # +2 for Excel 1-index and table header row
+            f"Row {self.row[0] + 2} "  # +2 for Excel 1-index and table header row
```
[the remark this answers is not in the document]
Joel-Pearce: We don't have to do it everywhere else - apart from here (because this is a legacy message and doesn't follow our to_message() protocol), and in our tests (which will have to happen anyway as even incrementing indexes +2 at the end of transformation means testing validation functions require us to mimic the +2). We have a function that constructs the cell locations (from the R4-specific PR) from the index and sheet/col, and the incrementation happens in there, but either way that function has to be called in every to_message() call.

We can also remove it from that function and put it at the end of the R3 and R4 transformation pipelines, which might be more readable from a developer perspective. I'll rebase this branch onto main now that the R4-specific PR has been merged and refactor everything now that it is all in the same branch.

core/extraction/towns_fund_round_three.py
Joel-Pearce: It's a ticket in the backlog we will probably move to the upcoming sprint, we separated it out because footfall outcomes is a bit trickier than the rest to retain the indexes for.

core/validation/validate.py
Joel-Pearce: You're correct, .asi8 returns a np array, which I then had to cast to tolist() anyway

tests/validation_tests/test_validate.py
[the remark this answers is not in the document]
cyrusdobbs: Ok. Could this be made clear in a comment please?

nsnape (approved):
All my comments addressed. Ta

Connor-OS (approved):

cyrusdobbs (approved):
LGTM

PR #300, opened 2023-10-17
[SMD-147] Fix project numbers in the wrong place in validation

Connor-OS:
This PR fixes a bug where the project number in the section field for some of the validation errors displayed does not match their on page position, for background this is because the projects in the sheet could be for example:
TD-ABC-01 - project 1
TD-ABC-03 - project 2
TD-ABC-05 - project 3

so we can't simple read the final part of the ID to get project number as we have done so far.

To fix this I have implemented two functions
- get_project_number_by_cell()

This uses the row index to get project number

- get_project_number_by_id()

this uses the project ID and the list of active ids

Then in the code wherever we need project number I have used which ever function is best for a given use case.

In the future we hope to refactor the to_message() and validatiion functions to only need one of these functions as maintaining both options is overly complex. However for now we are keeping both.

### DIFF
--- a/core/util.py
+++ b/core/util.py
@@ -1,4 +1,5 @@
 import itertools
+import math
 import re
 from typing import Any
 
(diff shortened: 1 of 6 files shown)


core/util.py
```diff
@@ -58,13 +59,36 @@ def group_by_first_element(tuples: list[tuple]) -> dict[str, list[tuple | Any]]:
     return nested
 
 
-def get_project_number(project_id):
-    """Extracts the project number from a project ID.
+def get_project_number_by_cell(row_index: int, sheet: str):
```
cyrusdobbs, line 62: I think `get_project_number_by_position` is more accurate? Cell to me is both row and column indexes.

core/util.py
```diff
@@ -58,13 +59,36 @@ def group_by_first_element(tuples: list[tuple]) -> dict[str, list[tuple | Any]]:
...
+    """
+    match sheet:
+        case "Funding" | "Funding Comments":
+            project_number = math.ceil((row_index - 32) / 28)
+        case "Output_Data":
+            project_number = math.ceil((row_index - 17) / 38)
+        case "RiskRegister":
+            project_number = math.ceil((row_index - 19) / 8) if row_index < 44 else math.ceil((row_index - 20) / 8)
+        case _:
+            project_number = None
```
cyrusdobbs, line 77: Do we want this to fail silently? The other project number function will raise an exception. I think it's best to keep the behaviour consistent.

core/validation/specific_validations/towns_fund_round_four.py
```diff
@@ -138,7 +141,7 @@ def validate_funding_profiles_funding_source(
         return [
             TownsFundRoundFourValidationFailure(
                 sheet="Funding",
-                section=f"Project Funding Profiles - Project {get_project_number(row['Project ID'])}",
+                section=f"Project Funding Profiles - Project {get_project_number_by_cell(row.name, 'Funding')}",
```
cyrusdobbs, line 144: so just to confirm - this `row.name` doesn't need `+ 2`?

cyrusdobbs (changes requested):
Nice work on this - feel much better about this method

core/validation/specific_validations/towns_fund_round_four.py
Connor-OS: No, because Joel added a +2 on all the indexes before validation

cyrusdobbs (approved):
LGTM 👍

PR #304, opened 2023-10-18
[SMD-269] added function for combining validation messages

cyrusdobbs:
https://dluhcdigital.atlassian.net/browse/SMD-269

### Change description
- adds a function for combing validation messages that have identical sheet, section and description by concatenating the cell values together.
- this function is currently unused

### How to test

### DIFF
--- a/core/validation/failures.py
+++ b/core/validation/failures.py
@@ -526,3 +526,25 @@ def failures_to_messages(
         tab: group_by_first_element(failures) for tab, failures in failures_grouped_by_tab.items()
     }
     return {"TabErrors": failures_grouped_by_tab_and_section}
+
+
+def group_validation_messages(validation_messages: list[tuple[str, str, str, str]]) -> list[tuple[str, str, str, str]]:
+    """Groups validation messages by concatenating the cell indexes together on identical sheet, section and description
+
+    :param validation_messages: a list of tuples representing validation messages: sheet, section, description, cell
+    :return: grouped validation messages
+    """
+    grouped_dict = {}
+    for item in validation_messages:
+        key = item[:3]  # use the first three values as the key - sheet, section and description
+        value = item[3]  # use the cell index as the value
+        if key in grouped_dict:
+            grouped_dict[key].append(value)  # collect cells to concatenate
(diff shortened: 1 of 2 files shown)


Joel-Pearce:
Line 520 of failures.py:
```
error_messages = list(set(error_messages))
```
will remove messages that should otherwise be grouped. 

Do you know where this function is going to be called? I think we might need to refactor ```failures_to_messages``` alongside this to ensure the grouping works end-to-end.

cyrusdobbs:
> Line 520 of failures.py:
> 
> ```
> error_messages = list(set(error_messages))
> ```
> 
> will remove messages that should otherwise be grouped.
> 
> Do you know where this function is going to be called? I think we might need to refactor `failures_to_messages` alongside this to ensure the grouping works end-to-end.

That line will still only remove completely duplicate errors though - i.e. all fields, including Cell number would have to be identical. This shouldn't happen in practice any more I don't think. If it does then we still want to remove the duplicates.

Joel-Pearce (approved):

cyrusdobbs:
> > > Line 520 of failures.py:
> > > ```
> > > error_messages = list(set(error_messages))
> > > ```
> > > 
> > > 
> > > will remove messages that should otherwise be grouped.
> > > Do you know where this function is going to be called? I think we might need to refactor `failures_to_messages` alongside this to ensure the grouping works end-to-end.
> > 
> > 
> > That line will still only remove completely duplicate errors though - i.e. all fields, including Cell number would have to be identical. This shouldn't happen in practice any more I don't think. If it does then we still want to remove the duplicates.
> 
> I think I was confused by the wording. `error_messages = [failure.to_message() for failure in validation_failures]` returns a list of tuples with all of tab, section, message, and (now) cell, but error_messages makes it sound like it's only the messages. Probably can refactor along with the rest of to_message() in a separate PR.

Yeah I think this a wider point of `failures.py` becoming full of inconsistencies and legacy terms. We should definitely look to tidy it up after launch. I've already created a ticket in reference to this [here](https://dluhcdigital.atlassian.net/browse/SMD-268)

PR #305, opened 2023-10-18
[SMD-228]: db load concurrency bug

Joel-Pearce:
Fixes the issue whereby two users could not simultaneously submit a return. 

### DIFF
--- a/core/controllers/ingest.py
+++ b/core/controllers/ingest.py
@@ -4,13 +4,14 @@ from io import BytesIO
 import numpy as np
 import pandas as pd
 from flask import Response, abort, g, jsonify
-from sqlalchemy import desc, func
+from sqlalchemy import desc, exc, func
 from werkzeug.datastructures import FileStorage
 
 from core.const import EXCEL_MIMETYPE, EXCLUDED_TABLES_BY_ROUND, SUBMISSION_ID_FORMAT
 from core.controllers.mappings import INGEST_MAPPINGS, DataMapping
 from core.db import db
 from core.db.entities import Organisation, Programme, Project, Submission
+from core.db.utils import transaction_retry_wrapper
 from core.extraction import transform_data
 from core.validation import validate
 from core.validation.initial_check import validate_before_transformation
@@ -50,7 +51,7 @@ def ingest(body: dict, excel_file: FileStorage) -> Response:
     if reporting_round in [1, 2]:
         populate_db_historical_data(workbook, INGEST_MAPPINGS)
     else:
-        populate_db(workbook, INGEST_MAPPINGS)
+        populate_db(workbook=workbook, mappings=INGEST_MAPPINGS)
 
     # provisionally removing unreferenced entities caused by updates to ingest process
     # TODO: DELETE THIS WHEN R1 AND R3 RE-INGESTED, OR NO MORE DUPLICATE ORGS
@@ -128,6 +129,7 @@ def next_submission_id(reporting_round: int) -> str:
     return SUBMISSION_ID_FORMAT.format(reporting_round, incremented_submission_num)
 
 
+@transaction_retry_wrapper(max_retries=5, sleep_duration=0.6, error_type=exc.IntegrityError)
 def populate_db(workbook: dict[str, pd.DataFrame], mappings: tuple[DataMapping]) -> None:
     """Populate the database with the data from the specified workbook using the provided data mappings.
 
@@ -142,7 +144,6 @@ def populate_db(workbook: dict[str, pd.DataFrame], mappings: tuple[DataMapping])
 
     reporting_round = int(workbook["Submission_Ref"]["Reporting Round"].iloc[0])
     programme_id = workbook["Programme_Ref"]["Programme ID"].iloc[0]
-
     # if already added this round, this entity used to drop existing round data
     programme_exists_same_round = (
         Programme.query.join(Project)
--- a/core/db/utils.py
+++ b/core/db/utils.py
@@ -1,0 +1,44 @@
+import time
+from functools import wraps
+from typing import Type
+
+from flask import current_app
+
+from core.db import db
+
+
+def transaction_retry_wrapper(max_retries: int, sleep_duration: float, error_type: Type[Exception]):
+    """Execute a transaction with retries for a specified error type.
+
+    This function wraps the provided 'function' in a database transaction and allows for retries
+    in case of a specified error type. If an error of the specified type occurs during the
+    transaction, the operation is retried up to a specified maximum number of times with a
+    delay between retries.
+
+    :param max_retries: The maximum number of retries in case of an error.
+    :param sleep_duration: The duration to sleep between retry attempts.
+    :param error_type: The type of error to catch and retry.
+    :return: None
+    """
+
+    def decorator(func):
+        @wraps(func)
+        def wrapper(retries_left=max_retries, *args, **kwargs):
+            for retry in range(retries_left):
+                try:
+                    with db.session.begin():
+                        func(*args, **kwargs)
+                    break
+                except error_type as transaction_error:
+                    if retry < retries_left:
+                        retries_left -= 1
+                        time.sleep(sleep_duration)
+                        current_app.logger.error(f"{func.__name__} failed with {str(transaction_error)}.")
+                        current_app.logger.error(f"Retries remaining: {retries_left}.")
+                    else:
+                        current_app.logger.error(f"Number of max retries exceeded for function '{func.__name__}'")
+                        raise transaction_error
+
+        return wrapper
+
+    return decorator
--- a/tests/db_tests/test_queries.py
+++ b/tests/db_tests/test_queries.py
@@ -1,9 +1,11 @@
+import logging
 from datetime import datetime, timedelta
+from unittest.mock import patch
 
 import numpy as np
 import pandas as pd
 import pytest
-from sqlalchemy import case
+from sqlalchemy import case, exc
 
 from core.const import GeographyIndicatorEnum, ITLRegion
 from core.db import db
@@ -16,6 +18,7 @@ from core.db.entities import (
     Submission,
 )
 from core.db.queries import download_data_base_query, outcome_data_query, project_query
+from core.db.utils import transaction_retry_wrapper
 
 
 def expected_outcome_join(query):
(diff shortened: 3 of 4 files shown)


core/controllers/ingest.py
```diff
@@ -361,3 +362,30 @@ def remove_unreferenced_organisations():
...
+
+    This function wraps the 'populate_db' operation in a database transaction, allowing for
+    retries in case of IntegrityError or InvalidRequestError. If an error occurs during the
+    transaction, the operation is retried up to a specified maximum number of times with a
+    delay of 0.4 seconds between retries.
+
+    :param workbook: A dictionary containing DataFrames representing the workbook data.
+    :param mappings: A tuple of DataMapping instances defining the mapping of workbook data to the database.
+    :return: None
+    """
```
cyrusdobbs, line 378: We could make this a generic function that accepts a function to call and a set of kwargs to pass to it, as well as parametrising `max_retries` etc. 

Something like:
```
def with_transaction(function: callable, f_kwargs: dict, max_retries: int = 3):
    pass
```
``

core/controllers/ingest.py
```diff
@@ -361,3 +362,30 @@ def remove_unreferenced_organisations():
...
+    logger = logging.getLogger(__name__)
```
cyrusdobbs, line 379: best to use `current_app.logger` here

cyrusdobbs (commented):
Just a couple of thoughts I had..

core/controllers/ingest.py
Joel-Pearce: Good shout on parameterising the function call. We were thinking of setting the max_retries as an env var so that it won't slow down automated unit testing if we decide to increase it by more than three.

nsnape: second this. It would be good to keep context of `populate_db` completely out of the doctring / function scope. So the idea is each function is totally separable in functionality, can be reused for any transaction and tested in isolation from populate_db (although an integration test with both could be useful).

Also, this func catches the specific IntegretyError, would it be appropriate / clear to make the specific exception to be caught also a param?

core/controllers/ingest.py
```diff
@@ -361,3 +362,30 @@ def remove_unreferenced_organisations():
...
+        try:
+            with db.session.begin():
+                populate_db(workbook, mappings)
+            break
+        except exc.IntegrityError as e:
+            if retry < max_retries:
+                max_retries -= 1
+                time.sleep(0.4)  # populate_db() runs for 0.3 seconds
+            else:
+                logger.error(e, exc_info=True)
```
nsnape, line 391: should this error then be re-raised?

nsnape: I think it currently will log it then continue without falling over?

nsnape (changes requested):

core/controllers/ingest.py
Joel-Pearce: I think I need to investigate more before deciding what to do with the error. There's a possibility another error might arise when queuing up ```populate_db``` like this.

core/controllers/ingest.py
```diff
@@ -50,7 +52,9 @@ def ingest(body: dict, excel_file: FileStorage) -> Response:
     if reporting_round in [1, 2]:
         populate_db_historical_data(workbook, INGEST_MAPPINGS)
     else:
-        populate_db(workbook, INGEST_MAPPINGS)
+        with_transaction(
+            populate_db, 3, 0.4, error_type=exc.IntegrityError, workbook=workbook, mappings=INGEST_MAPPINGS
```
cyrusdobbs, line 56: think it will be clearer to use keyword arguments for the max retries and sleep duration here

tests/db_tests/test_queries.py
```diff
@@ -481,3 +484,27 @@ def test_project_if_no_outcomes(seeded_test_client, additional_test_data):
...
+    specified maximum number of times. It also checks whether the error is properly
+    logged after reaching the maximum number of retries.
+    """
+    max_retries = 3
+    sleep_duration = 0.1
+    error_type = exc.IntegrityError
+
+    mocked_function = mocker.Mock(side_effect=exc.IntegrityError("Mocked IntegrityError", None, None))
+    mocked_function.__name__ = "mocked_function"
+    with patch.object(current_app.logger, "error") as mock_logger:
```
cyrusdobbs, line 504: caplog looks like a cleaner way of asserting on log contents

see this example:
https://github.com/communitiesuk/funding-service-design-post-award-data-store/blob/9210b8ad2400322305a45c9eafc20a89259d70c2/tests/test_handlers.py#L10-L24

core/controllers/ingest.py
```diff
@@ -361,3 +364,34 @@ def remove_unreferenced_organisations():
             db.session.delete(organisation)
 
     db.session.commit()
+
+
+def with_transaction(
+    function: callable, max_retries: int, sleep_duration: float, error_type: Type[Exception], **kwargs: dict
+) -> None:
```
cyrusdobbs, line 371: is this better placed in the `db` package? perhaps in a `utils` module?

cyrusdobbs (commented):
looking good - I like this approach

core/controllers/ingest.py
```diff
@@ -361,3 +364,34 @@ def remove_unreferenced_organisations():
...
+    """Execute a transaction with retries for a specified error type.
+
+    This function wraps the provided 'function' in a database transaction and allows for retries
```
nsnape, line 374: Can we just re-word the first line of summary paragraph slightly, as "wrap" is a bit misleading in Python. So something like, "Run provided function within an atomic database transaction..."

nsnape: Which makes me think, could we implement this function as a decorator? Might be useful? Or is that overkill?

nsnape: It means every time we want to run a function within a transaction/retry loop, we just decorate it in place, rather than always nesting within another function. Any thoughts?

tests/db_tests/test_queries.py
```diff
@@ -481,3 +484,27 @@ def test_project_if_no_outcomes(seeded_test_client, additional_test_data):
...
+    """
+    Test the behavior of 'with_transaction' function in case of IntegrityError.
+
+    This test simulates a scenario where a function raises an IntegrityError
+    and ensures that the 'with_transaction' function retries the operation up to the
+    specified maximum number of times. It also checks whether the error is properly
+    logged after reaching the maximum number of retries.
+    """
+    max_retries = 3
+    sleep_duration = 0.1
```
nsnape, line 499: You should mock sleep, then assert it is called rather than actually running a sleep method on tests

core/controllers/ingest.py
```diff
@@ -361,3 +364,34 @@ def remove_unreferenced_organisations():
...
+    """
+    for retry in range(max_retries):
+        try:
+            with db.session.begin():
+                function(**kwargs)
+            break
+        except error_type as transaction_error:
+            if retry < max_retries:
+                max_retries -= 1
+                time.sleep(sleep_duration)
```
nsnape, line 394: Do we currently log anything other than errors? If so we could log the retry here?

tests/db_tests/test_queries.py
```diff
@@ -481,3 +484,27 @@ def test_project_if_no_outcomes(seeded_test_client, additional_test_data):
 
     test_df = pd.read_sql(test_query_proj.statement, con=db.engine.connect())
     assert list(test_df["project_name"])
+
+
+def test_with_transaction_max_retries(mocker, test_client):
```
nsnape, line 489: could we also add a test that retries, but then succeeds on a subsequent try? I think you can pass a list of exceptions to the param for side_effects to make a call to the same function do different behaviour each time. 

So basically, a test we can assert calls the "function" more than once, and whereby no exception is raised at the end (since on the 2nd or 3rd attempt it did not raise anything)

nsnape (changes requested):

core/db/utils.py
```diff
@@ -0,0 +1,44 @@
...
+
+    :param max_retries: The maximum number of retries in case of an error.
+    :param sleep_duration: The duration to sleep between retry attempts.
+    :param error_type: The type of error to catch and retry.
+    :return: None
+    """
+
+    def decorator(func):
+        @wraps(func)
+        def wrapper(retries_left=max_retries, *args, **kwargs):
```
gidsg, line 26: If it's in a transaction what's the benefit of re-trying, are the transactions not atomic?

Joel-Pearce: Because I don't think it will queue by default. When I've tested locally it falls over if you have a database transaction open and another user tries to enter the same loop. 

We want to retry based on the approximate duration of the function call, and a reasonable number attempts (e.g. 5 if we expect a maximum of 5 users to be concurrently submitting returns).

core/db/utils.py
```diff
@@ -0,0 +1,44 @@
+import time
+from functools import wraps
+from typing import Type
+
+from flask import current_app
+
+from core.db import db
+
+
+def database_transaction(max_retries: int, sleep_duration: float, error_type: Type[Exception]):
```
nsnape, line 10: could we rename this to something a bit more specific? Ie transaction_wrapper?

nsnape: or transaction_retry_wrapper?

tests/db_tests/test_queries.py
```diff
@@ -481,3 +484,53 @@ def test_project_if_no_outcomes(seeded_test_client, additional_test_data):
...
+
+def test_database_transaction_wrapper_successful_retry(mocker, test_client, caplog):
+    """
+    Test the behavior of 'database_transaction' wrapper succeeding after failing once.
+    """
+    max_retries = 3
+    sleep_duration = 0.1
+    error_type = exc.IntegrityError
+
+    def mocked_side_effect(*args, **kwargs):
```
nsnape, line 523: I think a simpler way to achieve this is to just pass the expected side_effects as a list (the param can take an iterable). If you enter None, then for that iteration it doesn't raise. Ie, remove this mocked_side_effect sub-function and just use the following when you define mocked_function:

```
    mocked_function = mocker.Mock(side_effect=[
        exc.IntegrityError("Mocked IntegrityError", None, None),
        None,
        None,
    ])
```

nsnape: I tested this on mine locally and it worked

nsnape (changes requested):

nsnape (approved):

PR #309, opened 2023-10-19
Override work_mem for queries known to be slow

mattpep:
Mostly these are queries which export a lot of data. Note that though this setting can be set on the whole cluster, there is plenty of advice not to set it at that scope and instead to use either transaction or database level. In this case we set it on the txn level.

See the section titled _Considerations_ in
https://aws.amazon.com/blogs/database/tune-sorting-operations-in-postgresql-with-work_mem/

Jira ticket: https://dluhcdigital.atlassian.net/browse/FMD-103

### Change description

### How to test

### DIFF
--- a/core/serialisation/data_serialiser.py
+++ b/core/serialisation/data_serialiser.py
@@ -3,7 +3,9 @@ from typing import Generator
 
 from marshmallow_sqlalchemy import SQLAlchemySchema, auto_field
 from sqlalchemy.orm import Query
+from sqlalchemy.sql import text
 
+from core.db import db
 from core.db.entities import (
     Funding,
     FundingComment,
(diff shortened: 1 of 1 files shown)


nsnape (approved):

core/serialisation/data_serialiser.py
```diff
@@ -78,6 +80,9 @@ def serialise_download_data(
     for sheet in sheets_required:
         query_extender, schema = table_queries[sheet]
         extended_query = query_extender(base_query)
+        # TODO: remove this conditional test on postgres once we strip out sqlite (see FMD-8)
+        if db.engine.name == "postgresql":
+            db.session.execute(text("SET LOCAL work_mem TO '128MB'"))
```
gidsg, line 85: Does this not need to be reset as per the linked blog post?
> For the targeted queries, set the optimal value at the session level, then make sure to reset it (RESET work_mem;) back to the instance level setting at the end of the request so that the memory is released back to system and used for other activities.

mattpep: Change made as per suggestions

nsnape: I'm not 100% sure that line in the blog is technically correct, I thought the session is handled as a context, so will be garbage collected anyway between each query? Otherwise we would definitely have run out of memory overall

nsnape: But maybe just worth double checking :)

core/serialisation/data_serialiser.py
```diff
@@ -78,8 +80,13 @@ def serialise_download_data(
...
         table_data = extended_query.all()
         table_serialised = schema(many=True).dump(table_data)
+        if db.engine.name == "postgresql":
```
nsnape, line 88: so if this is actually needed, surely we also need to handle it in the case that the query/serialise functionality raises an error, and hence the transaction does not complete atomically?

nsnape (changes requested):

core/serialisation/data_serialiser.py
mattpep: That setting applies on a session level. So if there is an exception, the process will terminate and the db connection will close. Subsequent connections (e.g. by new processes) will use the default setting which is currently 4MB. So no need to revert.

nsnape: hmm, actually probably not since the whole session will end

nsnape (approved):

PR #313, opened 2023-10-23
fix place names null bug

Joel-Pearce:
Fixes bug whereby empty place_name returns a 500. .strip() is no longer required as no place_names in the sheet contain extraneous white spaces.

### DIFF
--- a/core/validation/initial_check.py
+++ b/core/validation/initial_check.py
@@ -81,7 +81,7 @@ def extract_submission_details(
     try:
         wrong_input_checks["Fund Type"] = (sheet_a2.iloc[5][4], {"Town_Deal", "Future_High_Street_Fund"})
         wrong_input_checks["Place Name"] = (
-            sheet_a2.iloc[6][4].strip(),
+            str(sheet_a2.iloc[6][4]).strip(),
             set(TF_PLACE_NAMES_TO_ORGANISATIONS.keys()),
         )
     except IndexError:
--- a/tests/validation_tests/test_pre_checks.py
+++ b/tests/validation_tests/test_pre_checks.py
@@ -70,6 +70,17 @@ def test_extract_round_four_submission_details(valid_workbook_round_four, mocker
     )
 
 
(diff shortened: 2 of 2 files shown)


core/validation/initial_check.py
```diff
@@ -81,7 +81,7 @@ def extract_submission_details(
     try:
         wrong_input_checks["Fund Type"] = (sheet_a2.iloc[5][4], {"Town_Deal", "Future_High_Street_Fund"})
         wrong_input_checks["Place Name"] = (
-            sheet_a2.iloc[6][4].strip(),
```
nsnape, line 84: Is it possible that fields with whitespace could still find their way in, as they did for example in R3 with some kind of document version control issue? Or would that scenario be handled by something else now? 

If not, then maybe we should just leave it in? Happy yo go with removing if you're happy though

Joel-Pearce: The ```strip()``` was causing a 500 if users copy and pasting something into the cell not recognised as a string. We have an updated sheet with updated place names, and none of them have trailing white spaces. I've checked that once and will check again. The alternative is casting the value to a string I suppose, which might be more reliable.

nsnape: Will a non-string in there be caught as a validation error anyway? Does the parser/mapping not automatically cast that to string? 

If not, we probably should just cast to string and keep the strip. I know the current data looks good, but it was so easy for that bug to come back in R3 (someone replaced one of the ingest files with a whitespace project, was all it took). Since we're not actually 100% control of that data we should probably take a defensive stance, to at least catch the edge-case problems we've already encountered

Joel-Pearce: It won't be caught as a validation error because this extraction of cell information occurs before any of the transformations or validation. 

Casting to string and stripping removes the bug, militates against the possibility of trailing white spaces, so I've updated and pushed.

nsnape (approved):

PR #323, opened 2023-10-24
[SMD-315]: float type in rag ratings bug fix

Joel-Pearce:
Fixes issue where putting a float into project progress rag ratings columns causes a 500.

### DIFF
--- a/core/extraction/towns_fund_round_three.py
+++ b/core/extraction/towns_fund_round_three.py
@@ -361,18 +361,6 @@ def extract_project_progress(df_data: pd.DataFrame, project_lookup: dict, round_
     df_data = df_data.drop(["Project Name"], axis=1)
     df_data = df_data.replace("< Select >", "")
 
-    # Fixes a validation edge case where a user enters a blank cell for a RAG. THis causes the column to be read as
-    # floats, because int type does not allow NAs. These floats are then checked against integer enum values, which fail
-    # validation. This fix works by forcing the type "Int64" (rather than int64), which allows NA integer values.
-    nullable_int_cols = ("Delivery (RAG)", "Spend (RAG)", "Risk (RAG)")
-    for col in nullable_int_cols:
-        try:
-            # use nullable "Int64" type to allow for empty cells in the submission whilst retaining data as ints
-            df_data[col] = df_data[col].astype("Int64")
-        except ValueError:
-            # cannot convert type due to non-numerical data in the cells so leave as object type
-            pass
-
     return df_data
 
 
--- a/tests/extraction_tests/test_round_four_extraction.py
+++ b/tests/extraction_tests/test_round_four_extraction.py
@@ -92,11 +92,6 @@ def test_extract_project_progress(mock_progress_sheet, mock_project_lookup):
         resources_assertions / "project_progress_expected.csv", index_col=0, dtype=str
     )
 
-    # set expected RAG columns to Int64 type in line with transformation logic
-    expected_project_progress[["Delivery (RAG)", "Spend (RAG)", "Risk (RAG)"]] = expected_project_progress[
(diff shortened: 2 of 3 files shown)


core/extraction/towns_fund_round_three.py
```diff
@@ -369,7 +369,7 @@ def extract_project_progress(df_data: pd.DataFrame, project_lookup: dict, round_
         try:
             # use nullable "Int64" type to allow for empty cells in the submission whilst retaining data as ints
             df_data[col] = df_data[col].astype("Int64")
-        except ValueError:
+        except (ValueError, TypeError):
```
nsnape, line 372: Are we definitely wanting to let non integer values through other than null? I may be inferring incorrectly here, but would it not be better to explicitly set conditions, rather than using a vague try/except catch-all?

So something like:
if cell is empty do blah..
if cell type is not in [x, y] then blah

nsnape: Is there an actual use case for having RAG as a float?

nsnape: if there is, then it should be explicitly set, as per the int, rather than doing it this way

nsnape: handling/catching exceptions in this way is often described as an Anti-pattern as all sorts of other unexpected errors can then slip through. In this case it is unlikely but is leaves the code vulnerable to bugs if the part within the try/except is ever extended.

nsnape (changes requested):

core/extraction/towns_fund_round_three.py
Joel-Pearce: If a user copy & pastes a float value into a RAG rating column it will fall over because the try-except block that was already present in the code only looked for ValueErrors and not TypeErrors. We want it let it through transformation in order to be picked up as an enum validation failure, otherwise here it will just 500.

I feel like we want to let all TypeErrors through so they can be caught as enums failures and send back to the user rather than failing on this loop of code.

nsnape: ok got ya, so it does kind of follow the workflow as we deliberately let the error pass silently and then "kind of" re catch the error later in validation. 

So I had a look through, I can't see where it gets validated against an enum (validation schema round 3 has it as str). Am I missing something.

For me, this also raises the question, why do we cast it to int64 here at all? Why not just pass whatever value it has through (including null) and validate that against enum?

core/extraction/towns_fund_round_three.py
```diff
@@ -369,7 +369,7 @@ def extract_project_progress(df_data: pd.DataFrame, project_lookup: dict, round_
...
             # cannot convert type due to non-numerical data in the cells so leave as object type
```
nsnape, line 373: This comment not strictly speaking true now as float is not non-numerical

nsnape (changes requested):

core/extraction/towns_fund_round_three.py
Joel-Pearce: ```validate_enums``` used to work column-by-column, and so anything that could be converted to int64 was. Now it works cell by cell in order to a) retain the original row index, and b) because it's actually a more granular and accurate way of checking datatypes. This function needs to be refactored to take that into account. I think we can probably dispense with the anti-pattern of attempting to cast entire columns in a dtype in a try/catch block.

The RAG rating columns are all validated against the RagEnum enum:

```
"Delivery (RAG)": enums.RagEnum,
 "Spend (RAG)": enums.RagEnum,
 "Risk (RAG)": enums.RagEnum,
```

Joel-Pearce: All of the redundant code has been removed.

nsnape (approved):

PR #330, opened 2023-10-26
convert postcodes to non nullable field

bobbi-hbrown:
https://dluhcdigital.atlassian.net/jira/software/c/projects/SMD/boards/140?quickFilter=815&selectedIssue=SMD-284

### Change description
Adds postcode as a non-nullable field to the Project Details (under Project Admin) sheet, and makes the lat/long column values nullable.

This can be tested manually by uploading different data returns with either completed or null values in columns H, I, K and L in the Project Admin tab

Note - I'm not sure if the "Lat/Long": pd.NA, values should be removed entirely from the test cases, as they should no longer cause a failure due to null values.

### DIFF
--- a/core/const.py
+++ b/core/const.py
@@ -765,6 +765,7 @@ INTERNAL_COLUMN_TO_FORM_COLUMN_AND_SECTION = {
     "Project Name": ("Project Name", "Project Details"),
     "Primary Intervention Theme": ("Primary Intervention Theme", "Project Details"),
     "Locations": ("Project Location(s) - Post Code (e.g. SW1P 4DF)", "Project Details"),
+    "Postcodes": ("Project Location - Post Code (e.g. SW1P 4DF)", "Project Locations - Post Code (e.g. SW1P 4DF) "),
     "Lat/Long": ("Project Location - Lat/Long Coordinates (3.d.p e.g. 51.496, -0.129)", "Project Details"),
     "Private Sector Funding Required": ("Private Sector Funding Required", "Private Sector Investment"),
     "Private Sector Funding Secured": ("Private Sector Funding Secured", "Private Sector Investment"),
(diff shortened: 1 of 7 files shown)


core/validation/specific_validations/towns_fund_round_four.py
```diff
@@ -252,17 +252,7 @@ def validate_locations(workbook: dict[str, pd.DataFrame]) -> list["TownsFundRoun
...
-            "Lat/Long",
-            "Single location | Project Location - Lat/Long Coordinates (3.d.p e.g. 51.496, -0.129)",
-            single_rows["Lat/Long"],
-        ),
         ("Locations", "Multiple locations | Project Locations - Post Code (e.g. SW1P 4DF)", multiple_rows["Locations"]),
-        (
-            "Lat/Long",
-            "Multiple locations | Project Locations - Lat/Long Coordinates (3.d.p e.g. 51.496, -0.129)",
-            multiple_rows["Lat/Long"],
-        ),
```
cyrusdobbs, line 265: Can you update the function docstring to reflect this change please

core/validation_schema.py
```diff
@@ -103,6 +103,7 @@
             "Primary Intervention Theme": enums.PrimaryInterventionThemeEnum,
         },
         "non-nullable": [
+            "Postcodes",
```
cyrusdobbs, line 106: We actually don't want to add this here because we carry out this validation as part of `round_4_specific_validations.validate_locations` - the docstring for that function explains why. Otherwise we will validate against Postcodes when a user hasn't entered single/multiple, which we want to avoid.

core/const.py
```diff
@@ -765,6 +765,7 @@ class ITLRegion(StrEnum):
     "Project Name": ("Project Name", "Project Details"),
     "Primary Intervention Theme": ("Primary Intervention Theme", "Project Details"),
     "Locations": ("Project Location(s) - Post Code (e.g. SW1P 4DF)", "Project Details"),
+    "Postcodes": ("Project Location - Post Code (e.g. SW1P 4DF)", "Project Locations - Post Code (e.g. SW1P 4DF) "),
```
cyrusdobbs, line 768: This can be removed

cyrusdobbs (changes requested):
I didn't remember the existing function when I wrote up the ticket - good job for finding it

cyrusdobbs (approved):
LGTM - please merge with squash

PR #332, opened 2023-10-26
[SMD-256]: cells for footfall outcomes

Joel-Pearce:
Present cells of errors from footfall outcomes to users.

### DIFF
--- a/core/extraction/towns_fund_round_three.py
+++ b/core/extraction/towns_fund_round_three.py
@@ -947,6 +947,8 @@ def extract_footfall_outcomes(df_input: pd.DataFrame, project_lookup: dict, prog
             footfall_instance = pd.concat([footfall_instance, df_input.iloc[(year_idx + 6), 2:-1]])
 
         footfall_instance = pd.DataFrame(footfall_instance).T
+        # footfall_idx + 6 for the index to match the index of the row for "Footfall Indicator" in original spreadsheet
+        footfall_instance.index = [df_input.index[footfall_idx + 6]]
         footfall_instance.columns = header
         footfall_df = pd.concat([footfall_df, footfall_instance])
 
@@ -954,11 +956,13 @@ def extract_footfall_outcomes(df_input: pd.DataFrame, project_lookup: dict, prog
     relevant_projects = set(footfall_df["Relevant Project(s)"])
     relevant_projects.discard("Multiple")
     if invalid_projects := relevant_projects - set(project_lookup.keys()):
+        # validation error raised here as post-transformation the project lookup information is lost
         raise ValidationError(
             [
-                vf.InvalidOutcomeProjectFailure(invalid_project=project, section="Footfall Indicator", row_indexes=None)
-                # TODO: add index for footfalls
-                for project in invalid_projects
+                vf.InvalidOutcomeProjectFailure(
+                    invalid_project=row["Relevant Project(s)"], section="Footfall Indicator", row_indexes=[idx + 2 + 5]
+                )  # +2 to match original spreadsheet index, extra +5 for Outcome -> Project row
+                for idx, row in footfall_df.loc[footfall_df["Relevant Project(s)"].isin(invalid_projects)].iterrows()
             ]
         )
 
--- a/core/validation/failures.py
+++ b/core/validation/failures.py
@@ -266,6 +266,11 @@ class InvalidEnumValueFailure(ValidationFailure):
         # additional logic for outcomes to differentiate between footfall and non-footfall
         if sheet == "Outcomes" and self.row_values[4] == "Year-on-year % change in monthly footfall":
             section = "Footfall Indicator"
+            # +5 as GeographyIndicator is 5 rows below Footfall Indicator
+            if column == "Geography Indicator":
+                actual_index = self.row_indexes[0] + 5
+                cell = f"C{actual_index}"
+                return sheet, section, cell, message
 
         # additional logic for risk location
         if sheet == "Risk Register":
@@ -363,8 +368,9 @@ class InvalidOutcomeProjectFailure(ValidationFailure):
     def to_message(self) -> tuple[str, str, str, str]:
         sheet = "Outcomes"
         section = self.section
-        if self.section == "Footfall Indicator":
-            cell_index = "Cell unavailable for this section"
+        if section == "Footfall Indicator":
+            row = self.row_indexes[0]
+            cell_index = f"B{row}"  # different col for footfall indicator
         else:
             cell_index = construct_cell_index(
                 table="Outcome_Data", column="Relevant project(s)", row_indexes=self.row_indexes
--- a/tests/extraction_tests/resources/assertion_data/round_three/outcomes_data_expected.csv
+++ b/tests/extraction_tests/resources/assertion_data/round_three/outcomes_data_expected.csv
@@ -1,195 +1,195 @@
 ,Higher Frequency,Project ID,Programme ID,Outcome,UnitofMeasurement,GeographyIndicator,Amount,Actual/Forecast,Start_Date,End_Date
-21,Half yearly,TD-FAK-01,,Patronage of the public transport system in the area of interest (for public transport schemes),Number of passengers,Town,,Actual,04/01/20,31/03/2021
-22,,TD-FAK-01,,Estimated carbon dioxide equivalent reductions as a result of support,Tonnes,Other / custom geography,,Actual,04/01/20,31/03/2021
-41,Monthly or higher,TD-FAK-01,,test custom outcome,test unit,Locations provided in 'Project Admin' tab,,Actual,04/01/20,31/03/2021
-21,Half yearly,TD-FAK-01,,Patronage of the public transport system in the area of interest (for public transport schemes),Number of passengers,Town,55,Actual,04/01/21,31/03/2022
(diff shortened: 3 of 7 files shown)


core/extraction/towns_fund_round_three.py
```diff
@@ -956,9 +957,10 @@ def extract_footfall_outcomes(df_input: pd.DataFrame, project_lookup: dict, prog
     if invalid_projects := relevant_projects - set(project_lookup.keys()):
         raise ValidationError(
             [
-                vf.InvalidOutcomeProjectFailure(invalid_project=project, section="Footfall Indicator", row_indexes=None)
-                # TODO: add index for footfalls
-                for project in invalid_projects
+                vf.InvalidOutcomeProjectFailure(
+                    invalid_project=row["Relevant Project(s)"], section="Footfall Indicator", row_indexes=[idx + 7]
+                )  # extra +5 for Outcome -> Project row
+                for idx, row in footfall_df.loc[footfall_df["Relevant Project(s)"].isin(invalid_projects)].iterrows()
```
cyrusdobbs, line 963: unclear as to why it says `+5` in the comment but `+7` in the code?

core/validation/failures.py
```diff
@@ -267,6 +267,11 @@ def to_message(self) -> tuple[str, str, str, str]:
         # additional logic for outcomes to differentiate between footfall and non-footfall
         if sheet == "Outcomes" and self.row_values[4] == "Year-on-year % change in monthly footfall":
             section = "Footfall Indicator"
+            # +5 as GeographyIndicator is 5 rows below Footfall Indicator
+            if column == "Geography Indicator":
+                actual_index = self.row_indexes[0] + 5
+                cell = "C" + str(actual_index)
```
cyrusdobbs, line 273: nit picky but I prefer an fstring here? I think it means we can do away with the `str` function

core/validation/failures.py
```diff
@@ -364,8 +369,8 @@ class InvalidOutcomeProjectFailure(ValidationFailure):
     def to_message(self) -> tuple[str, str, str, str]:
         sheet = "Outcomes"
         section = self.section
-        if self.section == "Footfall Indicator":
-            cell_index = "Cell unavailable for this section"
+        if section == "Footfall Indicator":
+            cell_index = "B" + str(self.row_indexes[0])  # different col for footfall indicator
```
cyrusdobbs, line 373: It's annoying that these cases don't fit into our `construct_cell_index` pattern.. I wonder if there is a way of managing it.. Will think again tomorrow on this

core/extraction/towns_fund_round_three.py
Joel-Pearce: Because the +5 represents "Footfall Indicator" and "Relevant Project(s)" being 5 rows apart, and the +2 is just generic for the spreadsheet. I should have made that clearer, will amend.

core/validation/failures.py
Joel-Pearce: Passing some kind of flag for footfall outcomes? I believe it's the only case where a column in the spreadsheet is different to a column in our tables

cyrusdobbs: A flag feels really coupling. This is a more dynamic way of doing it but could lead to the mappings becoming less interpretable.. what do you think?

see `additional_lookups`
```
def construct_cell_index(table: str, column: str, row_indexes: list[int], additional_lookups: tuple = ()) -> str:
    """Constructs the original Excel cell index of an error from the column and rows it occurred in.

    :param table: the internal table name where the error occurred
    :param column: the internal column name where the error occurred
    :param row_indexes: list of row indexes where the error occurred
    :param additional_lookups: additional lookups
    :return: indexes tuple of constructed letter and number indexes
    """
    column_letter = TABLE_AND_COLUMN_TO_ORIGINAL_COLUMN_LETTER[table][column]
    for lookup in additional_lookups:
        column_letter = column_letter[lookup]
    # remove duplicate row numbers to stop multiple identical indexes being constructed whilst retaining order
    row_indexes = list(dict.fromkeys(row_indexes))
    indexes = ", ".join([column_letter.format(i=index) for index in row_indexes])
    return indexes
```

this would come with a mapping change like:

```
    "Outcome_Data": {
        "Outcome": "B{i}",
        "UnitofMeasurement": "C{i}",
        "Relevant project(s)": {"Outcome Indicators (excluding footfall)": "D{i}",  # outcome section
                                "Footfall Indicator": "B{i}"},  # footfall outcome section
        "GeographyIndicator": "E{i}",
        "Amount": "F{i}-O{i}",
        "Higher Frequency": "P{i}",
    },
```

core/validation/failures.py
cyrusdobbs: Then as per my other comment this would be modified to:

```
@dataclass
class InvalidEnumValueFailure(ValidationFailure):
    """Class representing an invalid enum value failure."""

    sheet: str
    column: str
    row_indexes: list[int]
    row_values: tuple
    value: Any

    def to_message(self) -> tuple[str, str, str, str]:
        sheet = INTERNAL_TABLE_TO_FORM_TAB[self.sheet]
        column, section = INTERNAL_COLUMN_TO_FORM_COLUMN_AND_SECTION[self.column]
        message = msgs.DROPDOWN
        
        cell_index = construct_cell_index(table=self.sheet, column=self.column, row_indexes=self.row_indexes)

        # additional logic for outcomes to differentiate between footfall and non-footfall
        if sheet == "Outcomes" and self.row_values[4] == "Year-on-year % change in monthly footfall":
            cell_index = construct_cell_index(
                table=self.sheet,
                column=self.column,
                row_indexes=self.row_indexes,
                additional_lookups=("Footfall Indicator",),
            )

        # additional logic for risk location
        if sheet == "Risk Register":
            project_id = self.row_values[1]
            section = risk_register_section(project_id, self.row_indexes[0], self.sheet)

        if section == "Project Funding Profiles":
            project_number = get_project_number_by_position(self.row_indexes[0], self.sheet)
            section = f"Project Funding Profiles - Project {project_number}"

        return sheet, section, cell_index, message
```

cyrusdobbs: Not convinced this is the right solution.. or even worth making this additional complexity for. It all stems from having two different tables in the spreadsheet feed into the same dataframe.. one way to fix this would be to have separate dataframes for footfall and non-footfall

cyrusdobbs: This would actually be quite simple to do because we'd just create a new mapping for the footfall table that maps to the same db model:

```
DataMapping(
    worksheet_name="Outcome_Data",  # THIS WOULD BE e.g. Footfall_Outcome_Data
    model=ents.OutcomeData,
    column_mapping={
        "Submission ID": "submission_id",
        "Project ID": "project_id",
        "Programme ID": "programme_id",
        "Outcome": "outcome",
        "Start_Date": "start_date",
        "End_Date": "end_date",
        "UnitofMeasurement": "unit_of_measurement",
        "GeographyIndicator": "geography_indicator",
        "Amount": "amount",
        "Actual/Forecast": "state",
        "Higher Frequency": "higher_frequency",
    },
    fk_relations=[
        ("submission_id", ents.Submission, "submission_id", "submission_id"),
        ("programme_id", ents.Programme, "programme_id", "programme_id"),
        ("project_id", ents.Project, "project_id", "project_id"),
        ("outcome_name", ents.OutcomeDim, "outcome_id", "outcome"),
    ],
),
```

cyrusdobbs: Really thinking out loud now... but another example of where this could be useful is for `Other Funding Sources`.. it preserves more context of the original format which would then allow us to use the standard validation in more cases as well. i.e. The "Funding Source" and "Secured" dropdowns require validation for Other Funding Sources but not for the standard TF Funding Sources

cyrusdobbs:
I'm going to pull the conclusion of my comments out into a new ticket. For now having this additional logic to determine the footfall cells is fine.

core/extraction/towns_fund_round_three.py
cyrusdobbs: Can we make it even more explicit by doing `idx + 2 + 5` and reference why we do the `+2` here as well?

Could you also just add a general comment about why we're checking this and raising a Failure so it's super clear? It deviates from our usual validation so worth explaining in a comment I think otherwise it could be hard for someone without the context to know what's going on here.

core/validation/failures.py
cyrusdobbs, line 373: fstring here as well

core/extraction/towns_fund_round_three.py
```diff
@@ -947,6 +947,7 @@ def extract_footfall_outcomes(df_input: pd.DataFrame, project_lookup: dict, prog
             footfall_instance = pd.concat([footfall_instance, df_input.iloc[(year_idx + 6), 2:-1]])
 
         footfall_instance = pd.DataFrame(footfall_instance).T
+        footfall_instance.index = [df_input.index[footfall_idx + 6]]  # retain index of row of "Footfall Indicator"
```
cyrusdobbs, line 950: same with this `+6` can we make this more explicit?

cyrusdobbs (changes requested):
Just a few minor changes

core/extraction/towns_fund_round_three.py
cyrusdobbs: Think it's still missing a general comment about raising these Validation Failures

cyrusdobbs: specifically, why we raise one here instead of during validation

cyrusdobbs (approved):
LGTM 👍

PR #333, opened 2023-10-26
[SMD-325] add postcode validation for round 4

Connor-OS:
Validate that postcodes in cells H/K on project Admin tab are valid using regex.

The implementation her may be overkill as a regex is already used in the extraction (towns_fund_round_three.py lin 288) to create extract the poscodes that are in this column.

It may also be an usefull idea to swap out this validation for an ONS Postcode Directory lookup a sthis would then also catch postcodes of the correct format but not valid UK postcodes. But this should be included as a seperate ticket/PR if we decide to go down that route.

### Change description

### How to test


core/validation/messages.py
```diff
@@ -39,3 +39,6 @@
 )
 # PRE-TRANSFORMATION
 BLANK_SIGN_OFF = "In tab '8 - Review & Sign-Off', cell {cell} is blank but is required."
+
+# POSTCODE
+POSTCODE = "You entered an incorrect postcode. You must enter a postcode like SW1A 2AA"
```
cyrusdobbs, line 44: Has this gone past design/content?

tests/validation_tests/round_four_specific_validations/test_tf_r4_specific_validations.py
```diff
@@ -803,3 +805,59 @@ def test_validate_funding_profiles_funding_secured_not_null():
...
+
+    assert failures == [
+        TownsFundRoundFourValidationFailure(
+            sheet="Project Details",
+            section="Project Details",
+            column="Postcodes",
+            message="You entered an incorrect postcode. You must enter a postcode like SW1A 2AA",
+            row_indexes=[12, 13],
+        )
+    ]
```
cyrusdobbs, line 863: Could a case be added for when Locations has a value but Postcodes is an empty string

cyrusdobbs (changes requested):
Looks good - in reality the Postcodes column should only ever contain a comma separated list of valid postcodes, or be an empty string. This is because we do the extraction of postcodes and formatting of this column during extraction - I guess you saw that. So a lot of these test cases won't happen in practice but its fine to have them here anyway.

Did you test via `/ingest` to see how this behaves when you enter non comma separated lists of postcodes into the location boxes in the spreadsheet?

core/validation/messages.py
Connor-OS: Yep, this came from design through Gemma 👍

cyrusdobbs: Can this be updated to match the new content: `You entered an invalid postcode. Enter a full UK postcode, for example SW1A 2AA.`

cyrusdobbs (approved):
LGTM

PR #335, opened 2023-10-27
[SMD-332]: validate fund type vs place name

Joel-Pearce:
Validates that the place name has the correct fund type.


core/validation/initial_check.py
```diff
@@ -120,6 +125,20 @@ def pre_transformation_check(submission_details: dict[str, dict[str, dict]]) ->
...
+        for failure in failures
+        if not (
+            failure.value_descriptor == "Place Name vs Fund Type"
+            and any(
+                other_failure.value_descriptor in ["Fund Type", "Place Name"]
+                for other_failure in failures
+                if other_failure != failure
+            )
+        )
+    ]
```
Connor-OS, line 140: I think the other_failure != failure clause is unnecessary. It would work the same as
```
    failures = [
        failure
        for failure in failures
        if not (
            failure.value_descriptor == "Place Name vs Fund Type"
            and any(
                other_failure .value_descriptor in ["Fund Type", "Place Name"]
                for other_failure in failures
            )
        )
    ]
```
I also think there might be a more succinct refactor here as well

Connor-OS: we know that failure.value_descriptor == "Place Name vs Fund Type" so it dosen't == "Fund Type" or "Place Name"

Connor-OS (changes requested):

core/validation/initial_check.py
```diff
@@ -120,6 +125,20 @@ def pre_transformation_check(submission_details: dict[str, dict[str, dict]]) ->
...
+
```
cyrusdobbs, line 141: This is an improvement in time complexity and readability I think:

```
invalid_fund_or_place = [failure for failure in failures if failure.value_descriptor in ("Fund Type", "Place  Name")]
if invalid_fund_or_place:
   # remove the place vs fund failure
   failures = [failure for failure in failures if failure.value_descriptor != "Place Name vs Fund Type"]
````

File: core/validation/messages.py [hunk 3]
```diff
@@ -52,4 +52,6 @@
     "option from the dropdown list provided.",
     "Form Version": "The selected file must be the Town Deals and Future High Streets Fund Reporting Template "
     f"({TF_ROUND_4_TEMPLATE_VERSION}).",
+    "Place Name vs Fund Type": "We do not recognise the combination of fund type and place name in cells E7 and E8 in "
+    "“project admin”. Check the data is correct.",
```
cyrusdobbs, line 56: Has this gone via Content?

cyrusdobbs (changes requested):
Couple of comments 👍 

I think we need to refactor this module going forwards. We should be able to use a similar schema-based method to the standard validation.

core/const.py
```diff
@@ -870,3 +870,171 @@ class ITLRegion(StrEnum):
         "RiskOwnerRole": "P{i}",
     },
 }
+
+# map place to fund type associated with it
+PLACE_TO_FUND_TYPE = {
+    "Heanor": {"Future_High_Street_Fund"},
```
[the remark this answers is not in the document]
Joel-Pearce: Pre-transformation validation for place_name/fund_type works by looking at an ```entered_value``` and seeing if it is in a set of ```expected_values``` because we don't want and there is no reason for these expected_values to be duplicated so this dictionary uses a set as well in order to be validated by the same function without altering the datatype.

Connor-OS (approved):
My changes have been addressed, looks good

cyrusdobbs (approved):
LGTM

PR #336, opened 2023-10-27
[FMD-53] Remove superfluous joins

mattpep:

### Change description

### How to test


core/controllers/download.py
```diff
@@ -43,7 +43,12 @@ def download():
         itl_regions,
         outcome_categories,
     )
-    data_generator = serialise_download_data(base_query, outcome_categories)
+
+    if outcome_categories:
```
nsnape, line 47: I wonder, rather than have lots of if/else's around with different params in namespace, could we just rename the original `base_query` var as `query`, then we don't have `base_query` and `query` in namespace, and we don't need the else block to re-assign one to the other. i.e. only extend if we need to and overwrite the original var?

nsnape (changes requested):

core/controllers/download.py
mattpep: Good idea - updated in a later commit in this PR

nsnape (approved):

nsnape (changes requested):
Test properly in dev befor merging

core/controllers/download.py
```diff
@@ -35,15 +35,18 @@ def download():
...
         rp_end_datetime,
         organisation_ids,
         fund_ids,
         itl_regions,
-        outcome_categories,
     )
-    data_generator = serialise_download_data(base_query)
+
+    if outcome_categories:
+        query = query_extend_with_outcome_filter(query)
```
nsnape, line 47: This needs to be passed the outcome_categories param, this is why the outcomes filter is not working. Also, because the param defaults to None, we didn't notice it (when we test this stuff we did pass the param).

core/db/queries.py
```diff
@@ -26,13 +26,35 @@ def filter_on_regions(itl_regions: set[str]) -> list[UUID]:
     return updated_results
 
 
+def query_extend_with_outcome_filter(base_query, outcome_categories: list[str] | None = None):
```
nsnape, line 29: Can we have a docstring please?

core/db/queries.py
```diff
@@ -218,6 +226,18 @@ def outcome_data_query(base_query: Query) -> Query:
         else_=ents.Organisation.organisation_name,
     )
 
+    if join_outcome_info:
```
nsnape, line 229: This code is a duplicate of that in `query_extend_with_outcome_filter` minus a filter. But that func if not passed an outcome_categories param will default to None not apply the filter, so we can probably just re-use it here.

core/db/queries.py
```diff
@@ -238,13 +258,23 @@ def outcome_data_query(base_query: Query) -> Query:
...
+def outcome_dim_query(base_query: Query, join_outcome_info=False) -> Query:
     """
     Extend base query to select specified columns for OutcomeDim.
 
     :param base_query: SQLAlchemy Query of core tables with filters applied.
+    :param join_outcome_info: boolean of whether to join OutcomeData and OutcomeDim
     :return: updated query.
     """
+    if join_outcome_info:
+        base_query = base_query.join(  # left outer join: Outcomes is child of Project and hence optional
```
nsnape, line 270: in-line comment now needs removing

core/db/queries.py
```diff
@@ -238,13 +258,23 @@ def outcome_data_query(base_query: Query) -> Query:
...
+            ents.OutcomeData,
+            or_(
+                ents.Project.id == ents.OutcomeData.project_id,
```
nsnape, line 273: I think this _might_ now need to match the join code in `query_extend_with_outcome_filter `. I think this is actually a mistake in the code that was already here before this ticket, and we just never noticed.

But I'm not 100%, might be worth updating/replacing with the function call and check if everything still works

nsnape (changes requested):

core/db/queries.py
```diff
@@ -26,13 +26,43 @@ def filter_on_regions(itl_regions: set[str]) -> list[UUID]:
     return updated_results
 
 
+def query_extend_with_outcome_filter(base_query, outcome_categories: list[str] | None = None):
```
nsnape, line 29: Can you add Query as type hint here please? Can do as per `download_data_base_query`

nsnape: for both the 1st param, and the return type. Ta

core/db/queries.py
```diff
@@ -26,13 +26,43 @@ def filter_on_regions(itl_regions: set[str]) -> list[UUID]:
...
+    """
+    Extend base query to select additional columns for OutcomeDim / OutcomeData.
```
nsnape, line 31: I'm not sure this description is totally accurate, can it say something more like: 

"""
Extend base query to include join to OutcomeDim / OutcomeData.

Conditionally apply a filter on OutcomeDim catergory field
...
"""

nsnape (changes requested):

nsnape (approved):

PR #343, opened 2023-10-30
[SMD-343] add ingest param to disable load

cyrusdobbs:
https://dluhcdigital.atlassian.net/browse/SMD-343

### Change description
- adds param to disable load for an ingest request

### How to test

### DIFF
--- a/core/controllers/ingest.py
+++ b/core/controllers/ingest.py
@@ -39,6 +39,7 @@ def ingest(body: dict, excel_file: FileStorage) -> Response:
         "reporting_round"
     )  # optional, if None then file contents is expected to be round 3 in data model format
     place_names = body.get("place_names")  # optional, restrict ingest to submission of these places only
+    do_load = body.get("do_load", True)  # defaults to True, if False then do not load to database
     original_workbook = extract_data(excel_file=excel_file)
 
     if reporting_round:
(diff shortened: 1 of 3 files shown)


core/controllers/ingest.py
```diff
@@ -51,28 +52,39 @@ def ingest(body: dict, excel_file: FileStorage) -> Response:
     validate(workbook, original_workbook, reporting_round)
 
     clean_data(workbook)
+
+    if not do_not_load:
+        load_data(workbook, excel_file, reporting_round)
+
+    success_payload = {
+        "detail": f"Spreadsheet successfully validated{' and ingested' if not do_not_load else ''}",
```
Joel-Pearce, line 60: Do we not need to make it clear to the user that the spreadsheet has definitely not been ingested? Maybe saying it has been "validated" is a little bit ambiguous. I suppose we need to change the frontend screen to make this clear.

Joel-Pearce (changes requested):
Couple small comments

core/controllers/ingest.py
```diff
@@ -51,28 +52,39 @@ def ingest(body: dict, excel_file: FileStorage) -> Response:
     validate(workbook, original_workbook, reporting_round)
 
     clean_data(workbook)
+
+    if not do_not_load:
```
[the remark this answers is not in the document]
cyrusdobbs: The alternative is to set the param as `do_load` and set the default to `True`... maybe this is best tbh

core/controllers/ingest.py
cyrusdobbs: The user will never see this - this is just the API response but yeah it's a fair point for even this message

cyrusdobbs: Added the `loaded` attribute to the JSON response for further clarity:

https://github.com/communitiesuk/funding-service-design-post-award-data-store/blob/e8c78b9c70f5d83d1056bd365804d05f4038aa2b/core/controllers/ingest.py#L59-L65

Joel-Pearce (approved):
LGTM

PR #344, opened 2023-10-31
[SMD-336]: accurate cell mappings for outcomes amount col

Joel-Pearce:
Implements functionality where by errors in the 'Amount' column for Outcomes are accurately mapped back to the original cell location in which that error has occurred.

### DIFF
--- a/core/const.py
+++ b/core/const.py
@@ -1037,3 +1037,34 @@ PLACE_TO_FUND_TYPE = {
 }
 
 FUND_TYPE_TO_TD_OR_HS = {"Future_High_Street_Fund": "HS", "Town_Deal": "TD"}
+
+
+# maps the financial year's start year back to original col letter for non-footfall outcomes
+FINANCIAL_YEAR_TO_ORIGINAL_COLUMN_LETTER_FOR_NON_FOOTFALL_OUTCOMES = {
+    2020: "F{i}",
+    2021: "G{i}",
+    2022: "H{i}",
+    2023: "I{i}",
+    2024: "J{i}",
+    2025: "K{i}",
+    2026: "L{i}",
+    2027: "M{i}",
+    2028: "N{i}",
+    2029: "O{i}",
+}
+
+# maps the month of the financial year's start year back to the original column for footfall outcomes
+MONTH_TO_ORIGINAL_COLUMN_LETTER_FOR_FOOTFALL_OUTCOMES = {
+    4: "D{i}",
+    5: "E{i}",
+    6: "F{i}",
+    7: "G{i}",
+    8: "H{i}",
+    9: "I{i}",
+    10: "J{i}",
+    11: "K{i}",
+    12: "L{i}",
+    1: "M{i}",
+    2: "N{i}",
+    3: "O{i}",
+}
--- a/core/validation/utils.py
+++ b/core/validation/utils.py
@@ -1,4 +1,11 @@
+from datetime import datetime
+
 import pandas as pd
+
+from core.const import (
+    FINANCIAL_YEAR_TO_ORIGINAL_COLUMN_LETTER_FOR_NON_FOOTFALL_OUTCOMES,
+    MONTH_TO_ORIGINAL_COLUMN_LETTER_FOR_FOOTFALL_OUTCOMES,
+)
 
 
 def remove_duplicate_indexes(df: pd.DataFrame):
@@ -12,3 +19,47 @@ def remove_duplicate_indexes(df: pd.DataFrame):
     :return: the DataFrame without duplicate indexes
     """
     return df[~df.index.duplicated(keep="first")]
+
+
+def get_cell_indexes_for_outcomes(failed_row: pd.Series) -> str:
+    """
+    Constructs cell indexes for outcomes based on the provided failed row.
+
+    The function determines whether the error occurred in footfall outcomes (starting from row 60)
+    or non-footfall outcomes. If the index is greater than or equal to 60, it calculates the cell
+    index for footfall outcomes based on the provided 'Start_Date'. If the index is less than 60,
+    it calculates the cell index for non-footfall outcomes.
+
+    For footfall outcomes, the cell index is determined by adding a row index gap calculated from
+    the start year of the UK financial year as each year represents an additional 5 rows in the
+    original spreadsheet from the row where the entry for a given footfall outcome's data begins.
+
+    :param failed_row: A pandas Series representing a row where an error has occurred.
+    :return: A string containing the constructed cell index for outcomes.
+    """
+    start_date = failed_row["Start_Date"]
+    financial_year = get_uk_financial_year_start(start_date)
+    index = failed_row.name
+
+    # footfall outcomes starts from row 60
+    if index >= 60:
+        # row for 'Amount' column is end number of start year of financial year * 5 + 'Footfall Indicator' index
+        row_index_gap = int(str(financial_year)[-1]) * 5
+        index = int(index) + row_index_gap
+        cell_index = MONTH_TO_ORIGINAL_COLUMN_LETTER_FOR_FOOTFALL_OUTCOMES[start_date.month].format(i=index)
+    else:
+        cell_index = FINANCIAL_YEAR_TO_ORIGINAL_COLUMN_LETTER_FOR_NON_FOOTFALL_OUTCOMES[financial_year].format(i=index)
+
+    return cell_index
+
+
+def get_uk_financial_year_start(start_date: datetime) -> int:
+    """
+    Gets the start year of the UK financial year based on the provided start date.
+
+    :param start_date: A datetime in the format '%Y-%m-%d %H:%M:%S'.
+    :return: An integer representing the start year of the UK financial year.
+    """
+
+    financial_year = start_date.year if start_date.month >= 4 else start_date.year - 1
+    return financial_year
--- a/core/validation/validate.py
+++ b/core/validation/validate.py
@@ -155,6 +155,7 @@ def validate_types(
                         expected_type=exp_type,
                         actual_type=got_type,
                         row_indexes=[index],
+                        failed_row=row,
                     )
                 )
 
@@ -335,14 +336,19 @@ def validate_nullable(
    """
     for column in sheet.columns:
         if column in non_nullable and (pd.isna(sheet[column]).any() | (sheet[column].astype("string") == "").any()):
-            failed_indexes = sheet.index[(pd.isna(sheet[column]) | (sheet[column].astype("string") == ""))]
+            invalid_rows = sheet[(pd.isna(sheet[column]) | (sheet[column].astype("string") == ""))]
 
-            non_nullable_constraint_failure.append(
-                vf.NonNullableConstraintFailure(
-                    sheet=sheet_name,
-                    column=column,
-                    row_indexes=failed_indexes.tolist(),
+            if len(invalid_rows) > 0:
+                non_nullable_constraint_failure.extend(
+                    [
+                        vf.NonNullableConstraintFailure(
+                            sheet=sheet_name,
+                            column=column,
+                            row_indexes=[index],
+                            failed_row=row,
+                        )
+                        for index, row in invalid_rows.iterrows()
+                    ]
                 )
-            )
 
     return non_nullable_constraint_failure
--- a/core/validation/failures.py
+++ b/core/validation/failures.py
@@ -20,6 +20,7 @@ from core.const import (
 from core.extraction.utils import join_as_string
 from core.util import get_project_number_by_position
 from core.validation.exceptions import UnimplementedErrorMessageException
+from core.validation.utils import get_cell_indexes_for_outcomes
 
 
 class ValidationFailure(ABC):
@@ -191,6 +192,7 @@ class WrongTypeFailure(ValidationFailure):
     expected_type: str
     actual_type: str
     row_indexes: list[int]
+    failed_row: pd.Series | None
 
     def __str__(self):
         """Method to get the string representation of the wrong type failure."""
@@ -203,10 +205,13 @@ class WrongTypeFailure(ValidationFailure):
         sheet = INTERNAL_TABLE_TO_FORM_TAB[self.sheet]
         _, section = INTERNAL_COLUMN_TO_FORM_COLUMN_AND_SECTION[self.column]
         actual_type = INTERNAL_TYPE_TO_MESSAGE_FORMAT[self.actual_type]
+        cell_index = construct_cell_index(table=self.sheet, column=self.column, row_indexes=self.row_indexes)
+
         if sheet == "Outcomes":
             _, section = "Financial Year 2022/21 - Financial Year 2029/30", (
                 "Outcome Indicators (excluding " "footfall) and Footfall Indicator"
             )
+            cell_index = get_cell_indexes_for_outcomes(self.failed_row)
 
         if self.expected_type == "datetime64[ns]":
             message = msgs.WRONG_TYPE_DATE.format(wrong_type=actual_type)
@@ -218,8 +223,6 @@ class WrongTypeFailure(ValidationFailure):
             message = msgs.WRONG_TYPE_NUMERICAL
         else:
             message = msgs.WRONG_TYPE_UNKNOWN
-
-        cell_index = construct_cell_index(table=self.sheet, column=self.column, row_indexes=self.row_indexes)
 
         return sheet, section, cell_index, message
 
@@ -293,6 +296,7 @@ class NonNullableConstraintFailure(ValidationFailure):
     sheet: str
     column: str
     row_indexes: list[int]
+    failed_row: pd.Series | None
 
     def to_message(self) -> tuple[str, str, str, str]:
         """Generate error message components for NonNullableConstraintFailure.
@@ -320,6 +324,7 @@ class NonNullableConstraintFailure(ValidationFailure):
             if column == "Financial Year 2022/21 - Financial Year 2025/26":
                 section = "Outcome Indicators (excluding footfall) / Footfall Indicator"
                 message = msgs.BLANK_ZERO
+                cell_index = get_cell_indexes_for_outcomes(self.failed_row)
         elif sheet == "Funding Profiles":
             message = msgs.BLANK_ZERO
         elif section == "Programme-Wide Progress Summary":
--- a/tests/controller_tests/test_ingest.py
+++ b/tests/controller_tests/test_ingest.py
@@ -409,7 +409,9 @@ def test_ingest_endpoint_returns_validation_errors(test_client, example_data_mod
         "core.controllers.ingest.validate",
         side_effect=ValidationError(
             validation_failures=[
-                NonNullableConstraintFailure(sheet="Project Progress", column="Start Date", row_indexes=[5])
+                NonNullableConstraintFailure(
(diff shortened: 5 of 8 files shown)


core/validation/utils.py
```diff
@@ -12,3 +16,94 @@ def remove_duplicate_indexes(df: pd.DataFrame):
...
+def get_month(start_date: str) -> str:
+    """
+    Gets the month name based on the provided start date.
+
+    :param start_date: A string representing the start date in the format '%Y-%m-%d %H:%M:%S'.
+    :return: A string representing the month name.
+    """
+    start_date = datetime.strptime(start_date, "%Y-%m-%d %H:%M:%S")
+
+    return start_date.strftime("%B")
```
cyrusdobbs, line 85: This can just be `start_date.month`

cyrusdobbs (changes requested):

core/validation/utils.py
```diff
@@ -12,3 +16,94 @@ def remove_duplicate_indexes(df: pd.DataFrame):
...
+    """
+    cell_indexes = []
+
+    for index, start_date in failed_rows:
+        # footfall outcomes starts from row 60
+        if index >= 60:
+            financial_year = get_uk_financial_year_start(start_date)
+            # row for 'Amount' column is end number of start year of financial year * 5 + 'Footfall Indicator' index
+            row_index_gap = int(financial_year[-1]) * 5
+            index = str(int(index) + row_index_gap)
```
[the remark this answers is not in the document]
Joel-Pearce: image

There's six rows for each footfall outcome, and each of those are 5 rows apart. The rows are based on the financial year the data is reporting for. The end number of the starting year of that financial year * 5 is the gap between the first row's index and the index of the row where the failure has occurred.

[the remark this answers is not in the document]
cyrusdobbs: How come the index set during transformation isn't already aligned to this?

core/validation/utils.py
cyrusdobbs: I would argue that `str(start_date.month)` is clearer than `start_date.strftime("%B")`

It's not immediately obvious what `%B` means

core/validation/utils.py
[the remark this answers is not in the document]
Joel-Pearce: It's not possible to align as a row in our table comes from multiple rows in the original spreadsheet. In effect, different columns in footfall outcomes will have different original row indexes. By default, the row index is the first line - where Indicator Name is - but as soon as we want to map back to where "Amount" or "Relevant Project(s)" came from we have to figure out the gap between where the row where that information comes from and the row where the Indicator Name is for that outcome.

core/validation/utils.py
Joel-Pearce: ```str(start_date.month)``` returns a string of '5' if the month is May, and I think the string representation of the month's name is clearer.

We can do ```month_name = calendar.month_name[start_date.month]```, which is perhaps a little clearer as to the intention of the code.

cyrusdobbs: Ah I see what you mean now sorry. That looks good 👍

core/validation/utils.py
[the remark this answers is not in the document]
cyrusdobbs: With you now thanks

core/validation/utils.py
```diff
@@ -12,3 +17,95 @@ def remove_duplicate_indexes(df: pd.DataFrame):
...
+    """
+    Retrieves the 'Start_Date' column values for the specified indexes.
+
+    :param failed_rows: A pandas DataFrame containing the data of rows that have failed validation.
+    :return: A list of tuples, where each tuple contains an index and the corresponding 'Start_Date' value.
+    """
+    start_dates_series = failed_rows["Start_Date"] if "Start_Date" in failed_rows.columns else None
+    if start_dates_series is not None:
+        test = [(index, str(start_date)) for index, start_date in start_dates_series.iteritems()]
+        return test
```
Connor-OS, line 32: Why is this variable called test?

core/const.py
```diff
@@ -870,3 +870,31 @@ class ITLRegion(StrEnum):
...
+    "July": "G{i}",
+    "August": "H{i}",
+    "September": "I{i}",
+    "October": "J{i}",
+    "November": "K{i}",
+    "December": "L{i}",
+    "January": "M{i}",
+    "February": "N{i}",
+    "March": "O{i}",
+}
```
cyrusdobbs, line 900: Does it make sense to have all of these dates in the same dictionary - could it be clearer to have two separate dictionaries with more accurate names? 

This gives the sense that it's generic but is actually very specific. 

There is a TODO comment to add non-outcome tables but I think that will mean this is more unclear?

core/validation/utils.py
```diff
@@ -12,3 +17,95 @@ def remove_duplicate_indexes(df: pd.DataFrame):
...
+
+def get_start_date_column_for_failures(failed_rows: pd.DataFrame) -> list[tuple] | None:
+    """
+    Retrieves the 'Start_Date' column values for the specified indexes.
+
+    :param failed_rows: A pandas DataFrame containing the data of rows that have failed validation.
+    :return: A list of tuples, where each tuple contains an index and the corresponding 'Start_Date' value.
+    """
+    start_dates_series = failed_rows["Start_Date"] if "Start_Date" in failed_rows.columns else None
+    if start_dates_series is not None:
```
Connor-OS, line 30: Can we simplify ```if start_dates_series is not None:``` to ```if start_dates_series:``` in this case

core/const.py
```diff
@@ -870,3 +870,31 @@ class ITLRegion(StrEnum):
         "RiskOwnerRole": "P{i}",
     },
 }
+
+# maps the financial year's start tear, or a month, back to original col letter
+# non-footfall outcomes are mapped by the financial year's start year, but footfall outcomes just by the month
+# TODO: add non-outcome tables to this
```
Connor-OS, line 876: Will it be possible to use the same dict for Finances and Outputs? or will we have duplicate keys? If we can't could rename this dict to specify it's only for Outcomes

core/validation/validate.py
```diff
@@ -335,14 +336,18 @@ def validate_nullable(
...
+                    vf.NonNullableConstraintFailure(
+                        sheet=sheet_name,
+                        column=column,
+                        row_indexes=[index],
+                        failed_row=pd.DataFrame(dict(row), index=[index]),
+                    )
+                    for index, row in sheet[
+                        (pd.isna(sheet[column]) | (sheet[column].astype("string") == ""))
+                    ].iterrows()
+                ]
```
cyrusdobbs, line 350: I think this may be more clear to split it into two lines: 

Step 1: Filter dataframe to offending rows

Step 2: iterate the rows, returning a Failure for each

Can we also just return the row as a series instead of creating a df with a single row?

core/validation/validate.py
```diff
@@ -155,6 +155,7 @@ def validate_types(
                         expected_type=exp_type,
                         actual_type=got_type,
                         row_indexes=[index],
+                        failed_row=pd.DataFrame(dict(row), index=[index]),
```
cyrusdobbs, line 158: Can we not just return the row as a series?

core/validation/utils.py
```diff
@@ -12,3 +17,95 @@ def remove_duplicate_indexes(df: pd.DataFrame):
...
+        test = [(index, str(start_date)) for index, start_date in start_dates_series.iteritems()]
+        return test
+
+
+def get_cell_indexes_for_outcomes(failed_rows: pd.DataFrame) -> str:
```
cyrusdobbs, line 35: If we then return series instead then these functions need to be modified to expect a series

cyrusdobbs (changes requested):
More good progress - left a few more comments

Connor-OS (changes requested):

core/validation/utils.py
Joel-Pearce: Because I forgot to rename it after writing it as a test, will revert now!

core/validation/utils.py
Joel-Pearce: Unfortunately not as you cannot assert a truth value about a DataFrame or a Series.

core/const.py
Joel-Pearce: Possibly not, will have to investigate that. If not, will rename to make it clear it is exclusively a mapping for outcomes.

core/validation/validate.py
Joel-Pearce: The problem with this is that we need the index of the row to pass to the function that gets the cell index. We can now add row_indexes as an additional variable to work this out. I didn't want to do this before because with multiple indexes and multiples Series there is always the possibility of them getting decoupled and somehow rearranged in an incorrect order (with, for instance, some unforeseen code change down the line). 

I don't think I see any issues if we always expect both to have just one value. It will create difficulties if we move away from that model, but it looks like we're moving to streamline to just return one index per failure anyway.

core/const.py
Joel-Pearce: I think it might make more sense to have a dictionary mapping per table that needs this extra layer of mapping as per Connor's comment, but I'll do a quick investigation to see if there's a reasonable way we can get all of the tables requiring this mapping to only use one mapping or not.

cyrusdobbs: Worth considering if it's clear enough to have them in one table, even if it's feasible with unique keys

core/validation/validate.py
cyrusdobbs: Why do we need to pass more than one index? Think will always return a single row per Failure now I think?

Joel-Pearce: Yes, now it has been rewritten that is the case, but row_indexes still implies there will be multiple indexes, and it's still an array rather than an integer, so we probably need to refactor all of the functions to return row_index as a single integer to make this clearer in the upcoming refactor of validate.py. 

Now the function has been rewritten to only return a single index I will revert failed_row back to a series

core/validation/utils.py
```diff
@@ -12,3 +17,88 @@ def remove_duplicate_indexes(df: pd.DataFrame):
...
+    :param start_date: A string representing the start date in the format '%Y-%m-%d %H:%M:%S'.
+    :return: A string representing the start year of the UK financial year.
+    """
+    start_date = datetime.strptime(start_date, "%Y-%m-%d %H:%M:%S")
+
+    financial_year_start = datetime(start_date.year, 4, 1)
+    if start_date < financial_year_start:
+        financial_year_start = datetime(start_date.year - 1, 4, 1)
+
+    return f"{str(financial_year_start.year)}"
```
cyrusdobbs, line 68: we pass around datetimes instead of strings this function can just be this I think?
```
financial_year = date.year if date.month >= 4 else date.year - 1
return str(financial_year)
```

core/validation/utils.py
```diff
@@ -12,3 +17,88 @@ def remove_duplicate_indexes(df: pd.DataFrame):
...
+
+
+def footfall_outcome_amount_to_col(start_date: str) -> str:
+    """
+    Maps the month to the original column letter.
+
+    :param start_date: A string representing the start date in the format '%Y-%m-%d %H:%M:%S'.
+    :return: A string representing the original column letter corresponding to the month.
+    """
+    to_map = get_month(start_date)
```
cyrusdobbs, line 102: do we need a function for this? I think we can just do `calendar.month_name[date.month]`

core/const.py
```diff
@@ -870,3 +870,30 @@ class ITLRegion(StrEnum):
...
+    "July": "G{i}",
+    "August": "H{i}",
+    "September": "I{i}",
+    "October": "J{i}",
+    "November": "K{i}",
+    "December": "L{i}",
+    "January": "M{i}",
+    "February": "N{i}",
+    "March": "O{i}",
+}
```
cyrusdobbs, line 899: Can we split this out into two dictionaries, keeping the months and years as integers. 

i.e.
`YEAR_TO_OUTCOME_COLUMN` and `MONTH_TO_FOOTFALL_OUTCOME_COLUMN`

core/validation/utils.py
cyrusdobbs: If we split into two dicts, with ints as keys then we can just do 
`MONTH_TO_FOOTFALL_OUTCOME_COLUMN[date.month]`

core/validation/utils.py
```diff
@@ -12,3 +17,88 @@ def remove_duplicate_indexes(df: pd.DataFrame):
...
+
+    :param start_date: A string representing the start date in the format '%Y-%m-%d %H:%M:%S'.
+    :return: A string representing the month name.
+    """
+    start_date = datetime.strptime(start_date, "%Y-%m-%d %H:%M:%S")
+
+    return calendar.month_name[start_date.month]
+
+
+def non_footfall_outcome_amount_to_col(start_date: str) -> str:
```
cyrusdobbs, line 83: Can these functions be renamed `X_amount_to_col` -> `X_date_to_column`

cyrusdobbs (changes requested):
I think this can be simplified a fair bit

core/validation/utils.py
```diff
@@ -12,3 +19,48 @@ def remove_duplicate_indexes(df: pd.DataFrame):
...
+
+    # footfall outcomes starts from row 60
+    if index >= 60:
+        # row for 'Amount' column is end number of start year of financial year * 5 + 'Footfall Indicator' index
+        row_index_gap = int(str(financial_year)[-1]) * 5
+        index = str(int(index) + row_index_gap)
+        cell_index = FINANCIAL_MONTH_TO_ORIGINAL_COLUMN_LETTER_FOR_FOOTFALL_OUTCOMES[start_date.month].format(i=index)
+    else:
+        index = str(index)
+        cell_index = FINANCIAL_MONTH_TO_ORIGINAL_COLUMN_LETTER_FOR_NON_FOOTFALL_OUTCOMES[financial_year].format(i=index)
```
cyrusdobbs, line 52: do we have to convert to string or does the format method do this implicitly?

core/validation/utils.py
```diff
@@ -12,3 +19,48 @@ def remove_duplicate_indexes(df: pd.DataFrame):
...
+    """
+    start_date = failed_row["Start_Date"]
+    financial_year = get_uk_financial_year_start(start_date)
+
+    # footfall outcomes starts from row 60
+    if index >= 60:
+        # row for 'Amount' column is end number of start year of financial year * 5 + 'Footfall Indicator' index
+        row_index_gap = int(str(financial_year)[-1]) * 5
+        index = str(int(index) + row_index_gap)
+        cell_index = FINANCIAL_MONTH_TO_ORIGINAL_COLUMN_LETTER_FOR_FOOTFALL_OUTCOMES[start_date.month].format(i=index)
```
cyrusdobbs, line 49: same here

core/validation/utils.py
```diff
@@ -12,3 +19,48 @@ def remove_duplicate_indexes(df: pd.DataFrame):
     :return: the DataFrame without duplicate indexes
     """
     return df[~df.index.duplicated(keep="first")]
+
+
+def get_cell_indexes_for_outcomes(failed_row: pd.Series, index: int) -> str:
```
cyrusdobbs, line 24: Do we need to pass the series and the index? We can obtain the index from the series I think?

cyrusdobbs (changes requested):

core/validation/utils.py
```diff
@@ -12,3 +19,48 @@ def remove_duplicate_indexes(df: pd.DataFrame):
...
+    For footfall outcomes, the cell index is determined by adding a row index gap calculated from
+    the start year of the UK financial year as each year represents an additional 5 rows in the
+    original spreadsheet from the row where the entry for a given footfall outcome's data begins.
+
+    :param failed_row: A pandas Series representing a row where an error has occurred.
+    :param index: The index indicating the position of the error in the data.
+    :return: A string containing the constructed cell index for outcomes.
+    """
+    start_date = failed_row["Start_Date"]
+    financial_year = get_uk_financial_year_start(start_date)
```
cyrusdobbs, line 42: I thought `start_date` was a string? Or now this is a series does it return a datetime?

core/validation/utils.py
Joel-Pearce: The value in the dictionary is a string so there's no need

core/validation/utils.py
Joel-Pearce: When we iterate through the DataFrame, we iterate through the index and row as two separate values.

```
for index, row in workbook[sheet_name].iterrows():
```

The row does not contain the index by default. The index of the Series is the name of the columns. We could bodge 'original_index' in as an extra column but this feels simpler

Joel-Pearce: Although come to think of it, we have accessed the index before with Series.name. Let me investigate if that will work here

core/validation/utils.py
Joel-Pearce: It's a datetime, converted to a string before as I believed that would be easier for manipulating it. Now we can just leave as it is

core/validation/utils.py
cyrusdobbs: I mean this row
```
index = str(index)
```
and then this `.format(i=index)`

core/validation/utils.py
Joel-Pearce: Ah ok, we can access the index with the .name attribute, but it wouldn't work with unit tests unless .name is explicitly set after instantiating the Series

core/validation/utils.py
Joel-Pearce: It's month for footfall outcomes, but thinking about it 'financial month' doesn't make sense as there's no such thing. I've changed to INTEGER_OF_MONTH_TO_ORIGINAL_COLUMN_LETTER_FOR_FOOTFALL_OUTCOMES

core/validation/utils.py
cyrusdobbs: You can set name at the same time as instantiating the Series e.g.

```
pd.Series([1, 2, 3], dtype=np.int64, name='Numbers')
```

core/validation/utils.py
cyrusdobbs: I'm not sure ` INTEGER_OF_ ` is needed can we shorten to `MONTH_TO_COLUMN_LETTER_FOR_FOOTFALL_OUTCOMES`?

Connor-OS (approved):
This is much easier to read now 👍 Are we planning to do a similar thing for funding and outputs tabs next?

Joel-Pearce:
> This is much easier to read now 👍 Are we planning to do a similar thing for funding and outputs tabs next?

IMO, we should do this for all tabs that have date ranges as it's much easier for the user. Alongside a full refactor of how validation works, or else it might start getting very messy!

core/validation/utils.py
```diff
@@ -12,3 +19,48 @@ def remove_duplicate_indexes(df: pd.DataFrame):
...
+    index = failed_row.name
+
+    # footfall outcomes starts from row 60
+    if index >= 60:
+        # row for 'Amount' column is end number of start year of financial year * 5 + 'Footfall Indicator' index
+        row_index_gap = int(str(financial_year)[-1]) * 5
+        index = str(int(index) + row_index_gap)
+        cell_index = MONTH_TO_ORIGINAL_COLUMN_LETTER_FOR_FOOTFALL_OUTCOMES[start_date.month].format(i=index)
+    else:
+        index = str(index)
```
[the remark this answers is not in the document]
Joel-Pearce: Ah ok, I've removed the casting to string now

cyrusdobbs (approved):

PR #346, opened 2023-11-01
[SMD-344] add Funding Questions validation

cyrusdobbs:
https://dluhcdigital.atlassian.net/browse/SMD-344

### Change description
- adds validation for Funding Questions

### How to test


core/const.py
```diff
@@ -827,7 +833,12 @@ class ITLRegion(StrEnum):
         "Most Important Upcoming Comms Milestone": "N{i}",
         "Date of Most Important Upcoming Comms Milestone (e.g. Dec-22)": "O{i}",
     },
-    "Funding Questions": {"Question": "C{i}", "Guidance Notes": "M{i}", "Response": "E{i} to L{i}"},
+    "Funding Questions": {
+        "All Columns": "E{i} to L{i}",  # stretches across all 3 columns below
+        "TD 5% CDEL Pre-Payment\n(Towns Fund FAQs p.46 - 49)": "E{i}",
+        "TD RDEL Capacity Funding": "F{i} to H{i}",
+        "TD Accelerated Funding": "I{i} to L{i}",
+    },
```
cyrusdobbs, line 841: run this by design

cyrusdobbs: done

Connor-OS (commented):
LGTM

core/validation/utils.py
```diff
@@ -12,3 +15,50 @@ def remove_duplicate_indexes(df: pd.DataFrame):
...
+
+    :param value: value to check is numerical
+    :return: True if numerical, else False
+    """
+    try:
+        numerical_eval(value)
+        return True
+    except ValueError:
+        return False
+
```
Joel-Pearce, line 44: Can't we use ```isinstance(value, Number)``` here?

core/validation/specific_validations/towns_fund_round_four.py
```diff
@@ -508,7 +582,7 @@ class TownsFundRoundFourValidationFailure(ValidationFailure):
     section: str
     column: str
     message: str
-    row_indexes: list[int]
+    row_indexes: list[int | Hashable]
```
Joel-Pearce, line 585: Why can row_indexes now be a Hashable data type? Don't we want consistency across all row_indexes?

Joel-Pearce (changes requested):
A couple of small comments

core/validation/utils.py
cyrusdobbs: The value returned from the df is a string regardless of the value, because the `Response` column includes both text and numerical answers. For that reason this function just tries to to evaluate the `value` to a `float` or `int`, if it can't then it must not be a number and is rejected

core/validation/specific_validations/towns_fund_round_four.py
cyrusdobbs: Nice spot

I made this change without thinking because my IDE says that we're returning `Hashable` because `iterrows` documentation says that. An `int` is a `Hashable` though, the IDE just isn't that clever.

I'll revert

core/validation/utils.py
Joel-Pearce: In that case I think you can just do ```return value.isdecimal()``` as that will return True if the string can be a float or int, and removes the need for the try/catch block.

Joel-Pearce: Ah no, forget what I just wrote. That won't work on a float.

cyrusdobbs: hmmm that almost works.. Could use this instead though and just add a type hint that the function only works with strings. I think I prefer this


```
if string.replace(".", "").isnumeric():
    return True
else:
    return False
```

Joel-Pearce: Won't work for negative numbers though. If we don't expect any of them then I agree this is an improve over the try/catch!

cyrusdobbs: Ah good point. Maybe we'll just go with the casting to types option then.. I'll revisit monday

tests/validation_tests/test_utils.py
```diff
@@ -45,3 +51,43 @@ def test_get_project_number_by_id():
...
+    assert is_numeric("0")
+    assert is_numeric("1.1")
+    assert is_numeric("99999999999.9")
+    assert is_numeric("0.9")
+    assert not is_numeric("0-9")
+    assert not is_numeric("0,9")
+    assert not is_numeric("0'9")
+    assert not is_numeric("0@9")
+    assert not is_numeric("zero")
+    assert not is_numeric("nine")
```
Joel-Pearce, line 69: If we think negative numbers are possibility, add a case with negative numbers to this test

core/validation/utils.py
Joel-Pearce: Well, that or

```return string.replace(".", "").strip("-").isnumeric()```

Saves using two functions, and two try/catch blocks...but will validate "1000.000.000", which I suppose is a bigger concern? 

```return string.replace(".", "", 1).strip("-").isnumeric()```

Should work in all cases

Joel-Pearce: Although that will validate both "10000-" and "10000."

Maybe better to stick with the try/catch blocks, just surprised there's not a more elegant way of evaluating if a string can be a float or an int Python

cyrusdobbs: Yep going to stick with the try catch. I agree - it's surprising but I can't see anything online that looks better

tests/validation_tests/test_utils.py
cyrusdobbs: added 👍

Joel-Pearce (approved):
LGTM

cyrusdobbs:
Will merge after QA

PR #347, opened 2023-11-02
[SMD-386]: make cols nullable depending on if project is complete

Joel-Pearce:
When a project is complete, "Currently Project Delivery Stage" and columns pertaining to upcoming comms milestones are now nullable.

### DIFF
--- a/core/validation/const.py
+++ b/core/validation/const.py
@@ -1,0 +1,5 @@
+import numpy as np
+import pandas as pd
+
+# values we treat as null
+NA_VALUES = {"", np.nan, None, pd.NA}
--- a/core/validation/messages.py
+++ b/core/validation/messages.py
@@ -2,6 +2,7 @@ from core.const import TF_ROUND_4_TEMPLATE_VERSION
 
 # BLANK MESSAGES
 BLANK = "The cell is blank but is required."
+BLANK_IF_PROJECT_INCOMPLETE = "The cell is blank but is required for incomplete projects."
 BLANK_UNIT_OF_MEASUREMENT = (
     BLANK + "Check you’ve entered or generated a unit of measurement.\n"
     "If it’s for a custom indicator, enter the unit of measurement into the blank cell.\n"
--- a/core/validation/specific_validations/towns_fund_round_four.py
+++ b/core/validation/specific_validations/towns_fund_round_four.py
@@ -1,7 +1,6 @@
 import re
 from dataclasses import dataclass
 
-import numpy as np
 import pandas as pd
 
 import core.validation.messages as msgs
@@ -19,6 +18,7 @@ from core.extraction.utils import POSTCODE_REGEX
 from core.util import get_project_number_by_id, get_project_number_by_position
 from core.validation.failures import ValidationFailure, construct_cell_index
 from core.validation.utils import (
+    find_null_values,
     is_blank,
     is_from_dropdown,
     is_numeric,
@@ -44,12 +44,11 @@ def validate(workbook: dict[str, pd.DataFrame]) -> list["TownsFundRoundFourValid
         validate_funding_profiles_funding_secured_not_null,
         validate_psi_funding_gap,
         validate_locations,
-        validate_leading_factor_of_delay,
         validate_funding_spent,
         validate_funding_profiles_funding_secured_not_null,
         validate_psi_funding_not_negative,
         validate_postcodes,
-        validate_funding_questions,
+        validate_project_progress,
     )
 
     validation_failures = []
@@ -333,40 +332,6 @@ def validate_psi_funding_gap(workbook: dict[str, pd.DataFrame]) -> list["TownsFu
         ]
 
 
-def validate_leading_factor_of_delay(
-    workbook: dict[str, pd.DataFrame]
-) -> list["TownsFundRoundFourValidationFailure"] | None:
-    """Validates that Leading Factor of Delay is present for delayed projects.
-
-    This means rows where Project Delivery Status is "1. Not yet started" or "3. Ongoing - delayed", there must be a
-    valid value for Leading Factor of Delay.
-
-    If Leading Factor of Delay is present but not required due to the above constraint, then do not fail.
-
-    :param workbook: A dictionary where keys are sheet names and values are pandas
-                     DataFrames representing each sheet in the Round 4 submission.
-    :return: ValidationErrors
-    """
-    project_progress_df = workbook["Project Progress"]
-    delayed_mask = project_progress_df["Project Delivery Status"].isin(
-        {StatusEnum.NOT_YET_STARTED, StatusEnum.ONGOING_DELAYED}
-    )
-
-    na_values = {"", "< Select >", np.nan, None}
-    delayed_rows = project_progress_df[delayed_mask]
-    invalid_indexes = delayed_rows.index[delayed_rows["Leading Factor of Delay"].isin(na_values)].tolist()
-    if invalid_indexes:
-        return [
-            TownsFundRoundFourValidationFailure(
-                sheet="Project Progress",
-                section="Projects Progress Summary",
-                column="Leading Factor of Delay",
-                message=msgs.BLANK,
-                row_indexes=invalid_indexes,
-            )
-        ]
-
-
 def validate_funding_spent(workbook: dict[str, pd.DataFrame]) -> list["TownsFundRoundFourValidationFailure"] | None:
     """Validates that total reported funding spent on the sheet is less than the amount specified as allocated in a
     separate sheet stored on S3.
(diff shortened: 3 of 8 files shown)


core/validation/specific_validations/towns_fund_round_four.py
```diff
@@ -500,6 +501,70 @@ def validate_postcodes(workbook: dict[str, pd.DataFrame]) -> list["TownsFundRoun
...
+
+    :param df: DataFrame to search for invalid rows.
+    :param column: Column name to check for null values.
+    :param na_values: Set of values considered as null.
+    :return: DataFrame containing rows with null values in the specified column.
+    """
+    invalid_rows = df[df[column].isin(na_values)]
+    if len(invalid_rows) > 0:
+        invalid_rows["null_column"] = column
+    return invalid_rows
```
Connor-OS, line 565: I would be tempted to re-factor this so to avoid altering the dataframe rows.
```
invalid_rows =[]
for row in invalid_rows:
	for column in columns_to_check:
		if row[column].isin(na_values):
			invalid_rows.append(row.name, column)

return invalid_rows
```

Connor-OS (changes requested):
Looks good but I'm not sure it always picks up all blank cells that are pd.NA

core/validation/specific_validations/towns_fund_round_four.py
Connor-OS: potentially less readably though

core/validation/specific_validations/towns_fund_round_four.py
```diff
@@ -500,6 +501,70 @@ def validate_postcodes(workbook: dict[str, pd.DataFrame]) -> list["TownsFundRoun
...
 
 
+def validate_current_project_delivery_status_and_upcoming_comms_milestones(
+    workbook: dict[str, pd.DataFrame]
+) -> list["TownsFundRoundFourValidationFailure"] | None:
+    """
+    Validates the Project Progress table in the Round 4 submission workbook.
+
+    Checks for rows with a Project Delivery Status other than "4. Completed" where both
+    the Current Project Delivery Stage and Most Important Upcoming Comms Milestone are null.
```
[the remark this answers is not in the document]
Joel-Pearce: Yes, that's probably more accurate.

core/validation/specific_validations/towns_fund_round_four.py
```diff
@@ -500,6 +501,70 @@ def validate_postcodes(workbook: dict[str, pd.DataFrame]) -> list["TownsFundRoun
...
+    :param workbook: A dictionary containing sheet names as keys and pandas DataFrames
+                     representing each sheet in the Round 4 submission.
+    :return: A list of TownsFundRoundFourValidationFailure instances if validation failures occur,
+             otherwise returns None.
+    """
+    project_progress_df = workbook["Project Progress"]
+
+    project_incomplete_rows = project_progress_df[project_progress_df["Project Delivery Status"] != "4. Completed"]
+
+    na_values = {"", "< Select >", np.nan, None}
```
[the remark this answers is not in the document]
Joel-Pearce: I think it can, I suppose the question is whether we want to have the ```na_values``` adjustable or not. For example, for dropdowns we want "< Select >" to be null, for non-dropdowns we don't both with the value. There might be a more useful use case that arises in future development, not sure.

Are there instances where np.nan will not catch a null value that is only picked up by pd.NA? In our extraction scripts we default null values to np.nan. Have you discovered null values appearing as pd.NA? If so we need to change the code for all null validation.

core/validation/specific_validations/towns_fund_round_four.py
Joel-Pearce: We need the tacked-on ```"null_column"``` for our specific use case in which there are three columns we're assessing, but one or more of them could be wrong on the same line. As each validation failure needs to return the specific column where the error has occurred, we need a way of distinguishing from which column the failure has occurred if there are errors in multiple columns in one row. This is a workaround as the alternative seemed to be three separate DataFrames over which we iterate three times.

With that in mind, I don't think the above code woks?

core/validation/specific_validations/towns_fund_round_four.py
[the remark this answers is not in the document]
cyrusdobbs: Correct me if I'm wrong but `< Select >` is removed at transformation time I think?

core/validation/specific_validations/towns_fund_round_four.py
```diff
@@ -500,6 +501,70 @@ def validate_postcodes(workbook: dict[str, pd.DataFrame]) -> list["TownsFundRoun
...
+    columns_to_check = [
+        "Current Project Delivery Stage",
+        "Most Important Upcoming Comms Milestone",
+        "Date of Most Important Upcoming Comms Milestone (e.g. Dec-22)",
+    ]
+
+    invalid_rows = pd.DataFrame()
+
+    for column in columns_to_check:
+        invalid_rows = pd.concat([invalid_rows, find_invalid_rows(project_incomplete_rows, column, na_values)], axis=0)
```
cyrusdobbs, line 534: This is quite hard to follow and is a new way of doing this check. Can it be simplified?

core/validation_schema.py
```diff
@@ -144,17 +144,16 @@
...
             "Project Adjustment Request Status",
             "Delivery (RAG)",
             "Spend (RAG)",
             "Risk (RAG)",
             "Commentary on Status and RAG Ratings",
-            "Most Important Upcoming Comms Milestone",
-            "Date of Most Important Upcoming Comms Milestone (e.g. Dec-22)",
-            # TODO: Added all fields for Round 4 (was previously only Project ID)
+            # Most Important Upcoming Comms Milestone is sometimes nullable so has its own specific validation
+            # This also applies to Date of Most Important Upcoming Comms Milestone (e.g. Dec-22)
```
cyrusdobbs, line 156: I think this is another use case for splitting dataframes up in a previous stage. We could have `ProjectProgress` and `ProjectProgressComplete` - and then we could use the schema validation as usual? Not one for now just a thought

core/validation/specific_validations/towns_fund_round_four.py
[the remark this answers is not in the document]
cyrusdobbs: I agree with Connor here - I think we want consistency. Do we create a set of NA values like this anywhere else? Can we define once and reference instead if so

cyrusdobbs (changes requested):

core/validation/messages.py
```diff
@@ -2,6 +2,7 @@
 
 # BLANK MESSAGES
 BLANK = "The cell is blank but is required."
+BLANK_IF_PROJECT_INCOMPLETE = "The cell is blank but is required for incomplete projects."
```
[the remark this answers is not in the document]
Joel-Pearce: It has come from design.

core/validation_schema.py
Joel-Pearce: I'm not sure how straightforward that would be. Splitting DataFrames makes a lot of sense when we're retaining the tables as they original appear in the spreadsheet (like footfall and non-footfall outcomes), but splitting up a 'natural' DataFrame into two artificial ones based on how our validation works might be less clear than the bespoke validation. One to look at I suppose

cyrusdobbs: Hmmm maybe you're right. We need to try to move away from this kind of specific validation though and have a repeatable, configurable way of doing it.. Moving to a more simple spreadsheet built by us will help but I think this kind of scenario could arise then as well potentially.

core/validation/specific_validations/towns_fund_round_four.py
```diff
@@ -500,6 +500,70 @@ def validate_postcodes(workbook: dict[str, pd.DataFrame]) -> list["TownsFundRoun
...
+            TownsFundRoundFourValidationFailure(
+                sheet="Project Progress",
+                section="Projects Progress Summary",
+                column=row["null_column"],
+                message=msgs.BLANK_IF_PROJECT_INCOMPLETE,
+                row_indexes=[row.name],
+            )
+            for _, row in invalid_rows.iterrows()
+        ]
+
```
cyrusdobbs, line 547: I think creating a new df and doing this concatenation is hard to follow and doesn't offer any benefit.


I propose this way instead (with `find_null_values` in utils):
```    
def validate_current_project_delivery_status_and_upcoming_comms_milestones(
    workbook: dict[str, pd.DataFrame]
) -> list["TownsFundRoundFourValidationFailure"]:
    """
    Validates the Project Progress table in the Round 4 submission workbook.

    Checks for rows with a Project Delivery Status other than "4. Completed" where both
    the Current Project Delivery Stage or Most Important Upcoming Comms Milestone are null.
    If such invalid states are found, returns a list of TownsFundRoundFourValidationFailure instances.

    :param workbook: A dictionary containing sheet names as keys and pandas DataFrames
                     representing each sheet in the Round 4 submission.
    :return: A list of TownsFundRoundFourValidationFailure instances if validation failures occur,
             otherwise returns None.
    """
    project_progress_df = workbook["Project Progress"]

    project_incomplete_rows = project_progress_df[project_progress_df["Project Delivery Status"] != "4. Completed"]

    columns_to_check = [
        "Current Project Delivery Stage",
        "Most Important Upcoming Comms Milestone",
        "Date of Most Important Upcoming Comms Milestone (e.g. Dec-22)",
    ]

    failures = []
    for column in columns_to_check:
        invalid_rows = find_null_values(project_incomplete_rows, column)
        for _, row in invalid_rows.iterrows():
            failures.append(
                    sheet="Project Progress",
                    section="Projects Progress Summary",
                    column=column,
                    message=msgs.BLANK_IF_PROJECT_INCOMPLETE,
                    row_indexes=[row.name],
                )
            )

    return failures


def find_null_values(df: pd.DataFrame, column: str) -> pd.DataFrame:
    """Helper function to find rows with null values in a specified column.

    :param df: DataFrame to search for invalid rows.
    :param column: Column name to check for null values.
    :return: DataFrame of offending rows
    """
    return df[df[column].isin(NA_VALUES)]
```

cyrusdobbs (changes requested):

core/validation/specific_validations/towns_fund_round_four.py
```diff
@@ -500,6 +500,70 @@ def validate_postcodes(workbook: dict[str, pd.DataFrame]) -> list["TownsFundRoun
...
+    Checks for rows with a Project Delivery Status other than "4. Completed" where both
+    the Current Project Delivery Stage or Most Important Upcoming Comms Milestone are null.
+    If such invalid states are found, returns a list of TownsFundRoundFourValidationFailure instances.
+
+    :param workbook: A dictionary containing sheet names as keys and pandas DataFrames
+                     representing each sheet in the Round 4 submission.
+    :return: A list of TownsFundRoundFourValidationFailure instances if validation failures occur,
+             otherwise returns None.
+    """
+    project_progress_df = workbook["Project Progress"]
```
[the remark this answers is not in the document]
Joel-Pearce: I think it can, there's probably a way to refactor ```delayed_rows``` into ```columns_to_check```, I'll look at it now

core/validation/specific_validations/towns_fund_round_four.py
Joel-Pearce: Makes sense, especially if can refactor leading_factor_of_delay into the cols to check

Joel-Pearce: The two functions require slightly different rows to check in and slightly different error messages, but can work if as below:

```
# the column to check alongside the rows it should be checked in and the failure message
    columns_to_check = [
        ("Leading Factor of Delay", project_delayed_rows, msgs.BLANK),
        ("Current Project Delivery Stage", project_incomplete_rows, msgs.BLANK_IF_PROJECT_INCOMPLETE),
        ("Most Important Upcoming Comms Milestone", project_incomplete_rows, msgs.BLANK_IF_PROJECT_INCOMPLETE),
        ("Date of Most Important Upcoming Comms Milestone (e.g. Dec-22)", project_incomplete_rows,
         msgs.BLANK_IF_PROJECT_INCOMPLETE),
    ]

    failures = []
    for column, invalid_rows, message in columns_to_check:
        invalid_rows = find_null_values(invalid_rows, column)
        for _, row in invalid_rows.iterrows():
            failures.append(
                    sheet="Project Progress",
                    section="Projects Progress Summary",
                    column=column,
                    message=message,
                    row_indexes=[row.name],
                )
            )

    return failures
    ```

File: core/validation/const.py [hunk 9]
```diff
@@ -0,0 +1,5 @@
+import numpy as np
+import pandas as pd
+
+# values we treat as null
+NA_VALUES = {"", "< Select >", np.nan, None, pd.NA}
```
cyrusdobbs, line 5: Do we actually want `< Select >` here?

core/validation/utils.py
```diff
@@ -12,3 +14,21 @@ def remove_duplicate_indexes(df: pd.DataFrame):
...
+    if this function is called for each column to be validated.
+
+    :param df: DataFrame to search for invalid rows.
+    :param column: Column name to check for null values.
+    :return: DataFrame containing rows with null values in the specified column.
+    """
+    invalid_rows = df[df[column].isin(NA_VALUES)]
+    if len(invalid_rows) > 0:
+        invalid_rows["null_column"] = column
+    return invalid_rows
```
cyrusdobbs, line 34: We don't need to set a new column. Can just be this:

```
def find_null_values(df: pd.DataFrame, column: str) -> pd.DataFrame:
    """Helper function to find rows with null values in a specified column.

    :param df: DataFrame to search for invalid rows.
    :param column: Column name to check for null values.
    :return: DataFrame of offending rows
    """
    return df[df[column].isin(NA_VALUES)]
```

cyrusdobbs (commented):
Looking good! Just 2 minor comments

core/validation/specific_validations/towns_fund_round_four.py
cyrusdobbs: Looks good

core/validation/const.py
Joel-Pearce: It was written into the tests for leading_factor_of_delay. Looking at the extraction function for project progress, we replace all instances with an empty string. I've now removed all mention of < Select > from the validation

tests/test_misc.py
```diff
@@ -0,0 +1,49 @@
...
+    invalid_rows = find_null_values(df, column)
+
+    expected_invalid_rows = pd.DataFrame(
+        index=[7],
+        data=[
+            {"not-nullable": "", "nullable": ""},
+        ],
+    )
+
+    assert_frame_equal(invalid_rows, expected_invalid_rows)
```
cyrusdobbs, line 49: These are best placed in `tests/validation_tests/test_utils` instead I think

Joel-Pearce: Makes sense, I've moved the tests and pushed

cyrusdobbs (approved):
LGTM

Connor-OS (approved):
Looks good

PR #349, opened 2023-11-02
[FMD-8] replace localstack SQLite db with PSQL

nsnape:
https://dluhcdigital.atlassian.net/browse/FMD-8

### Change description

- Pointed all tests to run only on PSQL docker container from docker repo
- Updated GH actions to run tests on Docker instance
- Changed flask run to use Docker image, no option for SQLite now.
- Deprecated some redundant CLI commands (ERD)
- Updated REAME to reflect changes
- Removed references / conditionals throughout codebase referring to SQLite.

### How to test


README.md
```diff
@@ -72,26 +72,23 @@ Then run:
...
 
-## Run locally
+## Run App or Unit Tests Locally
+Local Docker stack must be up-to-date and running with `docker compose up` in order to use the app/tests locally:
+https://github.com/communitiesuk/funding-service-design-post-award-docker-runner
+
+Make migrations to ensure Docker PostgreSQL container is up-to-date:
+
 `flask db migrate`
+
```
cyrusdobbs, line 82: after running `flask db migrate` on an fresh database container _**with**_ data-store and account-store databases created

I get the error:
```
INFO  [alembic.runtime.migration] Context impl PostgresqlImpl.
INFO  [alembic.runtime.migration] Will assume transactional DDL.
ERROR [flask_migrate] Error: Target database is not up to date.
```

and have to run `flask db upgrade` to apply all existing migrations. Should this actually say `flask db upgrade`?

cyrusdobbs: after running `flask db migrate` on an fresh database container _**without**_ data-store and account-store databases created

I get the error:
```
sqlalchemy.exc.OperationalError: (psycopg2.OperationalError) connection to server at "localhost" (::1), port 5432 failed: FATAL:  database "data_store" does not exist
```

I know how to fix this because I know that the docker-compose can run into race conditions and not create these but should we include some information for a reader that may find this confusing?

core/cli.py
```diff
@@ -56,49 +56,13 @@ def seed_test():
     def drop():
```
cyrusdobbs, line 56: Do we want a test to cover this if we want to maintain it?

config/envs/unit_test.py
```diff
@@ -7,7 +7,7 @@
 @configclass
 class UnitTestConfig(DefaultConfig):
     TESTING = True
-    SQLALCHEMY_DATABASE_URI = "sqlite:///:memory:"
+    SQLALCHEMY_DATABASE_URI = "postgresql://postgres:password@localhost:5432/data_store"
```
cyrusdobbs, line 10: At the moment tests run by fails very slowly if we can't reach this instance when running `pytest`. 

Can we have a check that fails a test instantly if this can't be reached? 

Or alternatively checks the connection first before running any of the tests (not sure if this is an existing pytest design pattern or if it's the wrong way of thinking about it)

cyrusdobbs (commented):
A long time coming - good stuff

config/envs/unit_test.py
nsnape: Good shout, I wonder if it can go in the fixture or something?

README.md
nsnape: yeah, we should. I'd kind of assumed that they would just read the README on the compose repo, and was hesitant to replicate that info here, but you're right it will probably be more confusing not to include it here

nsnape: updated the docs

README.md
```diff
@@ -72,26 +72,31 @@ Then run:
...
-`flask run`
+## Run App or Unit Tests Locally
+Local Docker stack must be up-to-date and running with `docker compose up` in order to use the app/tests locally:
+https://github.com/communitiesuk/funding-service-design-post-award-docker-runner
 
-## Updating database migrations
-Whenever you make changes to database models, please run:
-`flask db migrate`
-This will create the migration files for your changes in ./db/migrations. Please then commit and push these to github so that the migrations will be run in the pipelines to correctly upgrade the deployed db instances with your changes.
+Make migrations to ensure Docker PostgreSQL container is up-to-date:
```
cyrusdobbs, line 79: I think this should be "Apply migrations..."

core/cli.py
nsnape: I'm moving all CLI stuff to subsequent ticket, as some of the changes in there will have implications. But I may still just remove this stuff - it all depends how coupled it is to the new methods - I don't want them to be un-maintained as they are at the moment, like you say I will add tests if we keep

config/envs/unit_test.py
nsnape: Moving this to subsequent test fixtures refactor ticket just to keep things moving

README.md
nsnape: changed, thanks

core/cli.py
```diff
@@ -56,49 +56,13 @@ def seed_test():
...
-            return
 
         with current_app.app_context():
             db.session.commit()
             db.drop_all()
             db.create_all()
 
         print("Database data dropped.")
-
-    @app.cli.command("erd")
```
gidsg, line 75: Are we happy to remove this? Do we need a todo to restore?

gidsg (approved):
LGTM 👍 can we capture the follow-ups in a separate ticket?

core/cli.py
nsnape: happy to remove. You need lots of 3rd party non-python libraries for this to work and it's just not easily maintainable, and currently not possible on our locked-down machines anyway. Now we have PSQL, we can get an ERD directly from PGadmin

PR #351, opened 2023-11-06
[SMD-336]: accurate cell mappings for outcomes amount col

Joel-Pearce:
Map failures in the 'Amount' column in Outcomes back to their original cells. 

### Change description

### How to test


core/validation/validate.py
```diff
@@ -335,14 +336,23 @@ def validate_nullable(
...
-            non_nullable_constraint_failure.append(
-                vf.NonNullableConstraintFailure(
-                    sheet=sheet_name,
-                    column=column,
-                    row_indexes=failed_indexes.tolist(),
+            invalid_rows = sheet[(pd.isna(sheet[column]) | (sheet[column].astype("string") == ""))]
+
+            # we do not want to remove melted row for 'Amount' in Outcomes
+            if sheet_name != "Outcome_Data" and column != "Amount":
+                invalid_rows = remove_duplicate_indexes(invalid_rows)
```
cyrusdobbs, line 343: we don't want to be introducing spreadsheet specific logic to this module ideally...

cyrusdobbs: I'm not immediately sure how to solve this.. one option is to add another param to the schema that determines if this is done or not. 

This is another symptom of validating post transformation... I think in the future we should look to do all of these basic blank/dropdown/type checks before transformation.

cyrusdobbs: This should be made easier when we're not tied to this spreadsheet and have some control over it

cyrusdobbs: I prefer that because at least it keeps this specific logic to one place. If it starts to leak into `validation.py` then it will be hard to manage it going forwards.

Joel-Pearce: Alright, looks like we can achieve the same functionality in one line with:
```
error_messages = list(set(error_messages))
```

In testing this I noticed that we don't sort the cells, and so sometimes they can be returned as "B21, B22" and "B22, B21" so I've amended that as well.

cyrusdobbs: Ah nice - we did this before I remember

cyrusdobbs (approved):
LGTM

PR #353, opened 2023-11-06
BAU: fix null place name bug

Joel-Pearce:
Fixes bug whereby a null place name caused an error.

### DIFF
--- a/tests/validation_tests/test_pre_checks.py
+++ b/tests/validation_tests/test_pre_checks.py
@@ -256,26 +256,59 @@ def test_full_pre_transformation_validation_pipeline_failure(valid_workbook):
     ]
 
 
-def test_full_pre_transformation_validation_pipeline_failure_place_vs_fund(valid_workbook_round_four):
+@pytest.fixture
+def mocked_project_admin_sheet(valid_workbook_round_four, request):
     valid_workbook_round_four["2 - Project Admin"][4] = [
         "",
         "",
         "",
         "",
         "",
-        "Future_High_Street_Fund",
-        "Newark",
+        request.param["fund"],
+        request.param["place"],
         "",
         "",
     ]
+    return valid_workbook_round_four
 
+
+@pytest.mark.parametrize(
+    "mocked_project_admin_sheet, expected",
+    [
+        (  # Newark is only a Town_Deal and so cannot be submitted as a Future_High_Street_Fund
+            {"fund": "Future_High_Street_Fund", "place": "Newark"},
+            [
+                WrongInputFailure(
+                    value_descriptor="Place Name vs Fund Type",
+                    entered_value="HS",
+                    expected_values={"TD"},
+                )
+            ],
+        ),
+        (
+            {"fund": "Future_High_Street_Fund", "place": ""},
+            [
+                WrongInputFailure(
+                    value_descriptor="Place Name",
+                    entered_value="",
+                    expected_values=TF_PLACE_NAMES_TO_ORGANISATIONS.keys(),
+                )
+            ],
+        ),
+        (
+            {"fund": "", "place": "Newark"},
+            [
+                WrongInputFailure(
+                    value_descriptor="Fund Type",
+                    entered_value="",
+                    expected_values={"HS", "TD"},
+                )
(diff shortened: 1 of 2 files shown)


tests/validation_tests/test_pre_checks.py
```diff
@@ -279,3 +279,28 @@ def test_full_pre_transformation_validation_pipeline_failure_place_vs_fund(valid
...
+        "",
+        "",
+        "",
+        "",
+        "",
+        "Future_High_Street_Fund",
+        "",
+        "",
+        "",
+    ]
```
cyrusdobbs, line 295: Should we add a test for if both are null?

tests/validation_tests/test_pre_checks.py
```diff
@@ -279,3 +279,28 @@ def test_full_pre_transformation_validation_pipeline_failure_place_vs_fund(valid
             expected_values={"TD"},
         )
     ]
+
+
+def test_full_pre_transformation_validation_pipeline_failure_place_null(valid_workbook_round_four):
+    valid_workbook_round_four["2 - Project Admin"][4] = [
```
cyrusdobbs, line 285: These tests could also be a nice candidate for parametrisation like:

```
@pytest.fixture
def mocked_project_admin_sheet(valid_workbook_round_four, request):
    valid_workbook_round_four["2 - Project Admin"][4] = [
        "",
        "",
        "",
        "",
        "",
        request.param["fund"],
        request.param["place"],
        "",
        "",
    ]
    return valid_workbook_round_four


@pytest.mark.parametrize("mocked_project_admin_sheet, expected",
                         [({"fund": "Future_High_Street_Fund", "place": "Newark"}, [
                             WrongInputFailure(
                                 value_descriptor="Place Name vs Fund Type",
                                 entered_value="HS",
                                 expected_values={"TD"},
                             )
                         ]),
                          ({"fund": "Future_High_Street_Fund", "place": ""}, [
                              WrongInputFailure(
                                  value_descriptor="Place Name",
                                  entered_value="",
                                  expected_values=TF_PLACE_NAMES_TO_ORGANISATIONS.keys(),
                              )
                          ])], indirect=["mocked_project_admin_sheet"])
def test_full_pre_transformation_validation_pipeline_failure_place_vs_fund(mocked_project_admin_sheet, expected):
    with pytest.raises(ValidationError) as e:
        validate_before_transformation(mocked_project_admin_sheet, 4, ["Newark"])
    assert e.value.validation_failures == expected

```

cyrusdobbs (commented):
Left a couple of thoughts

tests/validation_tests/test_pre_checks.py
Joel-Pearce: I don't think it makes a difference to this specific bug as if fund_type is None the failure won't arise; however, if place_name is None (or not valid), then the look-up of PLACE_NAME_TO_FUND_TYPE will result in None, and whilst ```is None in []" won't raise an error, ```is "TD" in None``` will

However, I don't think it will hurt, and good to have the coverage whilst making this change anyway. Will add

tests/validation_tests/test_pre_checks.py
Joel-Pearce: I've slept on pytest.parameterize, used the above and added the null Fund Type case

tests/validation_tests/test_pre_checks.py
```diff
@@ -256,26 +256,59 @@ def test_full_pre_transformation_validation_pipeline_failure(valid_workbook):
...
+    [
+        (
+            {"fund": "Future_High_Street_Fund", "place": "Newark"},
+            [
+                WrongInputFailure(
+                    value_descriptor="Place Name vs Fund Type",
+                    entered_value="HS",
+                    expected_values={"TD"},
+                )
+            ],
```
cyrusdobbs, line 286: I think it's worth adding a comment to explain that Newark is only valid for TD

Joel-Pearce: Added and pushed

cyrusdobbs (approved):
LGTM 👍

PR #354, opened 2023-11-07
[SMD-391]: remove submission id autoincrement

Joel-Pearce:
Remove autoincremented generation of submission id and replace it with a randomly generated 6 character hash at the end.

### DIFF
--- a/core/controllers/ingest.py
+++ b/core/controllers/ingest.py
@@ -1,10 +1,12 @@
 """Provides a controller for spreadsheet ingestion."""
+import random
+import string
 from io import BytesIO
 
 import numpy as np
 import pandas as pd
 from flask import Response, abort, g, jsonify
-from sqlalchemy import desc, exc, func
+from sqlalchemy import exc
 from werkzeug.datastructures import FileStorage
 
 from core.const import EXCEL_MIMETYPE, EXCLUDED_TABLES_BY_ROUND, SUBMISSION_ID_FORMAT
@@ -137,23 +139,18 @@ def get_metadata(workbook: dict[str, pd.DataFrame], reporting_round: int | None)
     return metadata
 
 
-def next_submission_id(reporting_round: int) -> str:
-    """Get the next submission ID by incrementing the last in the DB.
+def new_submission_id(reporting_round: int) -> str:
+    """Generates a submission ID with the reporting_round and a randomised 6 character hash.
+
+    A submission id formatted as "S-R{0:0=2d}-{hash}" is generated in order to represent the reporting_round
+    and to ensure that submission ids concurrently generated are not the same.
 
     :return: The next submission ID.
     """
-    # Conversion to cast numpy int types from pandas data extract
-    reporting_round = int(reporting_round)
-    latest_submission = (
-        Submission.query.filter_by(reporting_round=reporting_round)
-        # substring submission number digits, cast to int and order to get the latest submission
-        .order_by(desc(func.cast(func.substr(Submission.submission_id, 7), db.Integer))).first()
-    )
-    if not latest_submission:
-        return SUBMISSION_ID_FORMAT.format(reporting_round, 1)  # the first submission
+    # Generate a randomised hash of length 6 to append to the Submission ID
+    six_character_hash = "".join(random.choices(string.ascii_uppercase, k=6))
 
-    incremented_submission_num = latest_submission.submission_number + 1
-    return SUBMISSION_ID_FORMAT.format(reporting_round, incremented_submission_num)
+    return SUBMISSION_ID_FORMAT.format(reporting_round, hash=six_character_hash)
 
 
 @transaction_retry_wrapper(max_retries=5, sleep_duration=0.6, error_type=exc.IntegrityError)
@@ -204,7 +201,7 @@ def populate_db(workbook: dict[str, pd.DataFrame], mappings: tuple[DataMapping])
             db.session.flush()
 
     else:
-        submission_id = next_submission_id(reporting_round)
+        submission_id = new_submission_id(reporting_round)
 
     for mapping in mappings:
         worksheet = workbook[mapping.worksheet_name]
--- a/tests/controller_tests/test_ingest.py
+++ b/tests/controller_tests/test_ingest.py
@@ -15,7 +15,7 @@ from core.const import EXCEL_MIMETYPE
 from core.controllers.ingest import (
     extract_data,
     get_metadata,
-    next_submission_id,
+    new_submission_id,
     remove_unreferenced_organisations,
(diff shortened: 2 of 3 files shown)


tests/controller_tests/test_ingest.py
```diff
@@ -515,12 +515,33 @@ def test_extract_data_extracts_from_multiple_sheets(example_data_model_file):
...
+        (3, "S-R03-ABCDEF"),
+        (4, "S-R04-ABCDEF"),
+    ],
+)
+def test_next_submission_id_first_submission(reporting_round, expected_submission_id, monkeypatch):
+    # Monkeypatch the random key generator to return a fixed value
+    monkeypatch.setattr("random.choices", lambda population, k: ["A", "B", "C", "D", "E", "F"])
+
+    sub_id = next_submission_id(reporting_round)
+    assert sub_id == expected_submission_id
```
nsnape, line 532: I don't think we even need to do this patching, could we just assert that the start of the string is as expected, and that the end is 6 caps letteers?

nsnape: i.e. if we parameterise the func with just the rounds?

tests/controller_tests/test_ingest.py
```diff
@@ -592,21 +596,37 @@ def test_save_submission_file(test_client):
...
-    sub = Submission(
-        submission_id="S-R01-3",
-        reporting_period_start=datetime.now(),
-        reporting_period_end=datetime.now(),
-        reporting_round=1,
-    )
-    db.session.add(sub)
-    sub_id = next_submission_id(reporting_round=np.int64(1))
-    assert sub_id == "S-R01-4"
+@pytest.mark.parametrize(
```
nsnape, line 599: parameterisation might be overkill for this one, only need to test that numpy type is handled the once?

nsnape (changes requested):

tests/controller_tests/test_ingest.py
Joel-Pearce:
```
assert sub_id.startswith(f"S-R0{1}".format(reporting_round)
assert sub_id[:-6].isupper()
assert len(sub_id) == 11
```
Something like that? To me it seems a little messier. Is there a reason we want to avoid patching random.choice?

core/controllers/ingest.py
```diff
@@ -138,22 +140,14 @@ def get_metadata(workbook: dict[str, pd.DataFrame], reporting_round: int | None)
...
-        .order_by(desc(func.cast(func.substr(Submission.submission_id, 7), db.Integer))).first()
-    )
-    if not latest_submission:
-        return SUBMISSION_ID_FORMAT.format(reporting_round, 1)  # the first submission
+    # Generate a randomised hash of length 6 to append to the Submission ID
+    six_character_hash = "".join(random.choices(string.ascii_uppercase, k=6))
 
-    incremented_submission_num = latest_submission.submission_number + 1
-    return SUBMISSION_ID_FORMAT.format(reporting_round, incremented_submission_num)
+    return SUBMISSION_ID_FORMAT.format(reporting_round, six_character_hash)
```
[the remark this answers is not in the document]
Joel-Pearce: Not necessarily, but I've not updated it anyway to make it a bit clearer

tests/controller_tests/test_ingest.py
Joel-Pearce: Yeh, you're right there. Rewritten as one test.

tests/controller_tests/test_ingest.py
nsnape: There's no reason to avoid it par se, but this kind of pattern, ie mock something, then assert it is what we just mocked it to be, doesn't really achieve much, especially WRT testing what the function actually does. The most important things to test here are:

- The bit at the start is generated as per the logic, based on the round parameter passed
- The string format looks correct
- The final chars are generated as expected ie 6 capital chars.

The last 2 are not really tested if we just tell the function what to return in a mock. And the function doesn't really produce a definitive id (as they're random) so why test that it does?

nsnape: > ```
> assert sub_id.startswith(f"S-R0{1}".format(reporting_round)
> assert sub_id[:-6].isupper()
> assert len(sub_id) == 11
> ```
> 
> Something like that? To me it seems a little messier. Is there a reason we want to avoid patching random.choice?

Yeah basically. It's not really "messier" - it's asserting the structure of the string is generated as we expect. If we just mock the expected outcome, then assert against that, if something changes in the code that should break the test, the test may still pass as it effectively ignores the pertinent part of the code that's supposed to be tested

nsnape: also, I think the code would be
```
assert sub_id.startswith(f"S-R0{reporting_round}")
assert sub_id[-6:].isupper()
assert len(sub_id) == 11
```

nsnape: But you could always split into suitable var names if you want it to be clearer what it's testing?

nsnape (changes requested):

tests/controller_tests/test_ingest.py
Joel-Pearce: Should be fine to just make those three assertions, it's a relatively straightforward method to test

tests/controller_tests/test_ingest.py
```diff
@@ -515,12 +515,32 @@ def test_extract_data_extracts_from_multiple_sheets(example_data_model_file):
     assert isinstance(list(workbook.values())[0], pd.DataFrame)
 
 
-def test_next_submission_id_first_submission(test_client):
-    sub_id = next_submission_id(reporting_round=1)
-    assert sub_id == "S-R01-1"
-
-
-def test_next_submission_id_existing_submissions(test_client):
+@pytest.mark.parametrize(
```
nsnape, line 518: This line can be tidied to be 
`@pytest.mark.parametrize("reporting_round", [1, 2, 3, 4])`

tests/controller_tests/test_ingest.py
```diff
@@ -515,12 +515,32 @@ def test_extract_data_extracts_from_multiple_sheets(example_data_model_file):
...
+    ],
+)
+def test_new_submission_id_first_submission(reporting_round):
+    sub_id = new_submission_id(reporting_round)
+    assert sub_id.startswith(f"S-R0{reporting_round}")
+    assert sub_id[-6:].isupper()
+    assert len(sub_id) == 12
+
+
+@pytest.mark.parametrize(
```
nsnape, line 534: see previous comment about readability

nsnape (changes requested):

nsnape (approved):

PR #355, opened 2023-11-07
[BAU] validation failure refactor

cyrusdobbs:

### Change description
- all `Failure`s are per cell for consistency 
- `Failure`s raised during historical ingest are separated from `failures.py` into `historical.py`
- moves common `Failure` attributes to a parent class to enforce consistency across `Failures`

Next refactor is to go further by modifying `validate.py` so separate historical validation from other validation - i.e. foreign key validation should never occur for non-historical rounds so we don't need to do that check. The plan is to make it configurable - i.e. you can configure a set of checks per dataset.

### How to test

### DIFF
--- a/core/extraction/towns_fund_round_three.py
+++ b/core/extraction/towns_fund_round_three.py
@@ -10,6 +10,7 @@ import pandas as pd
 from pandas.tseries.offsets import MonthEnd
 
 import core.validation.failures as vf
+import core.validation.messages as msgs
 from core.const import (
     OUTCOME_CATEGORIES,
     OUTPUT_CATEGORIES,
@@ -856,14 +857,19 @@ def extract_outcomes(df_input: pd.DataFrame, project_lookup: dict, programme_id:
     if invalid_projects := relevant_projects - set(project_lookup.keys()):
         raise ValidationError(
             [
-                vf.InvalidOutcomeProjectFailure(
-                    invalid_project=row["Relevant project(s)"],
+                vf.GenericFailure(
+                    sheet="Outcomes",
                     section="Outcome Indicators (excluding footfall)",
-                    row_indexes=[idx + 2],  # +2 here as caught mid-ingest before post-transformation incrementation
+                    # +2 here as caught mid-ingest before post-transformation incrementation
+                    cell_index=vf.construct_cell_index(
+                        table="Outcome_Data", column="Relevant project(s)", row_index=idx + 2
+                    ),
+                    message=msgs.DROPDOWN,
                 )
-                for idx, row in outcomes_df.loc[outcomes_df["Relevant project(s)"].isin(invalid_projects)].iterrows()
+                for idx, _ in outcomes_df.loc[outcomes_df["Relevant project(s)"].isin(invalid_projects)].iterrows()
             ]
         )
+
     # Drop rows with Section header selected as the outcome from dropdown on form - This is not a valid outcome option.
     outcomes_df.drop(outcomes_df[outcomes_df["Relevant project(s)"] == "*** ORIGINAL: ***"].index, inplace=True)
     outcomes_df.insert(0, "Project ID", outcomes_df["Relevant project(s)"].map(project_lookup))
@@ -959,9 +965,13 @@ def extract_footfall_outcomes(df_input: pd.DataFrame, project_lookup: dict, prog
         # validation error raised here as post-transformation the project lookup information is lost
         raise ValidationError(
             [
-                vf.InvalidOutcomeProjectFailure(
-                    invalid_project=row["Relevant Project(s)"], section="Footfall Indicator", row_indexes=[idx + 2 + 5]
-                )  # +2 to match original spreadsheet index, extra +5 for Outcome -> Project row
+                vf.GenericFailure(
+                    sheet="Outcomes",
+                    section="Footfall Indicator",
+                    # +2 to match original spreadsheet index, extra +5 for Outcome -> Project row
+                    cell_index=f"B{idx + 2 + 5}",
+                    message=msgs.DROPDOWN,
+                )
                 for idx, row in footfall_df.loc[footfall_df["Relevant Project(s)"].isin(invalid_projects)].iterrows()
             ]
         )
--- a/core/extraction/towns_fund_round_three.py
+++ b/core/extraction/towns_fund_round_three.py
@@ -9,7 +9,7 @@ import numpy as np
 import pandas as pd
 from pandas.tseries.offsets import MonthEnd
 
-import core.validation.failures as vf
+import core.validation.failures.user as uf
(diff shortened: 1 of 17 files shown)


Joel-Pearce (changes requested):
Looking good, definitely a move in the right direction, comments just about rebasing and how we name/distinguish things

core/validation/historical.py
Connor-OS, line 1: What would happen if we deleted all thes failures as well as the concept of HistoricalValidationFailure. All these are only referenced in tests so I'm struggling to see how any of the code in this file can be run in production. Is this just fluff?

core/validation/failures.py
```diff
@@ -19,120 +17,59 @@
...
+class GenericFailure(ValidationFailure):
+    """Class representing a sign-off failure in the Review & Sign-Off section."""
 
     sheet: str
-    column: str
+    section: str
+    cell_index: str
+    message: str
 
     def __str__(self):
```
Connor-OS, line 57: Every failure has these __str__() methods. But there functionality is eclipsed by to_message() is there anywhere on in the API where we would rather return __str__() or no?

core/extraction/towns_fund_round_three.py
```diff
@@ -856,14 +857,19 @@ def extract_outcomes(df_input: pd.DataFrame, project_lookup: dict, programme_id:
...
                     section="Outcome Indicators (excluding footfall)",
-                    row_indexes=[idx + 2],  # +2 here as caught mid-ingest before post-transformation incrementation
+                    # +2 here as caught mid-ingest before post-transformation incrementation
+                    cell_index=vf.construct_cell_index(
+                        table="Outcome_Data", column="Relevant project(s)", row_index=idx + 2
+                    ),
+                    message=msgs.DROPDOWN,
                 )
-                for idx, row in outcomes_df.loc[outcomes_df["Relevant project(s)"].isin(invalid_projects)].iterrows()
+                for idx, _ in outcomes_df.loc[outcomes_df["Relevant project(s)"].isin(invalid_projects)].iterrows()
```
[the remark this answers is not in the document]
cyrusdobbs: It was rebased shortly before your review - I imagine you didn't have the changes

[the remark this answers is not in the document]
cyrusdobbs: > Failure to select a valid project for a given outcome is a GenericFailure that relays a msgs.DROPDOWN message to the user, but we still have a class called InvalidEnumFailure that functions in a similar way and returns the same message. What qualifies something as a GenericFailure vs any of the other classes of failures?

Fair point - I'll look at this

core/validation/failures.py
```diff
@@ -19,120 +17,59 @@
...
-class NonUniqueFailure(ValidationFailure):
-    """Class representing a non-unique value failure."""
+class GenericFailure(ValidationFailure):
+    """Class representing a sign-off failure in the Review & Sign-Off section."""
 
     sheet: str
-    column: str
+    section: str
+    cell_index: str
+    message: str
```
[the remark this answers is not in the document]
cyrusdobbs: For now I've put these in the child classes that require them

core/validation/historical.py
```diff
@@ -0,0 +1,138 @@
+from abc import abstractmethod
+from dataclasses import dataclass
+from typing import Any
+
+from core.validation.exceptions import UnimplementedErrorMessageException
+from core.validation.failures import ValidationFailureMixin
+
+
+class HistoricalValidationFailure(ValidationFailureMixin):
+    """Historical Validation Failures do not support validation messages."""
```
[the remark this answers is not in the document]
cyrusdobbs: I agree - I rushed the naming at the time. I think those names make more sense 👍

core/validation/historical.py
cyrusdobbs: See slack

core/validation/failures.py
cyrusdobbs: These could probably be improved but the reason is because we may want to use the `str` magic methods when developing new pipelines

cyrusdobbs: See slack for more info

core/extraction/towns_fund_round_three.py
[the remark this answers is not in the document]
cyrusdobbs: I think the reason it makes sense to use a `GenericFailure` like this is because we know all the details of the Failure when we instantiate it because it's being called from a Towns Fund Round 3/4 specific level - just like the Round 4 specific validations. The `Failure`s produced in `validation.py` all have to go through some translation stage in order to map it back to the original spreadsheet.. I think this is the distinction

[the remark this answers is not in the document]
Joel-Pearce: Alright, so ```GenericFailure``` is effectively a failure we can't neatly slot into another category.

core/validation/validate.py
```diff
@@ -9,11 +9,14 @@
 from numpy.typing import NDArray
 
 import core.validation.failures as vf
+import core.validation.historical as hvf
```
Joel-Pearce, line 12: Do we want to the file where the non-user facing failures are to be called historical.py, and to import it as hvf?

core/validation/failures.py
```diff
@@ -19,120 +17,59 @@
...
-        raise UnimplementedErrorMessageException
+    column: str | list[str]
+    row_index: int
 
 
 @dataclass
-class NonUniqueFailure(ValidationFailure):
-    """Class representing a non-unique value failure."""
+class GenericFailure(ValidationFailure):
+    """Class representing a sign-off failure in the Review & Sign-Off section."""
```
Joel-Pearce, line 50: This also represents what was previously the OutcomeProjectFailure, should the description for this be something like, "Class representing failures which do not specifically fit into errors specific to a round or type of validation"?

core/validation/validate.py
Joel-Pearce: I suppose it's not too important as long as it's clearly mentioned what we mean by historical in the file.

cyrusdobbs: I've done some renaming / refactoring of this in my latest commit 👍

core/validation/failures.py
cyrusdobbs: corrected this

core/validation/validate.py
Joel-Pearce: Yep, much clearer now, and a good way of organising our validation/failures

core/validation/failures.py
Joel-Pearce: The docstring still indicates that GenericFailure only covers sign-off failure

Joel-Pearce: Ah no, I can see it now, it's been moved to a new file

Joel-Pearce (approved):
LGTM

PR #358, opened 2023-11-09
[SMD-211] authorise fund type and refactor

Joel-Pearce:
Refactors functionality into ```initial_check.py``` into ```pre_transformation_validate.py```.

Also enables users to be authorised on the basis of which fund type they are allowed to submit for.


core/validation/pre_transformation_validate.py
```diff
@@ -0,0 +1,148 @@
...
+    )
+    wrong_input_checks["Reporting Period"] = (sheet1.iloc[4][1], (reporting_period,))
+    wrong_input_checks["Fund Type"] = (
+        FUND_TYPE_TO_TD_OR_HS.get(sheet2.iloc[5][4], sheet2.iloc[5][4]),
+        ("TD", "HS"),
+    )
+    wrong_input_checks["Place Name"] = (
+        str(sheet2.iloc[6][4]).strip(),
+        tuple(TF_PLACE_NAMES_TO_ORGANISATIONS.keys()),
+    )
```
cyrusdobbs, line 93: This is still very specific to the TF spreadsheet.. can we pull this out as config.. something like:

```
Check = namedtuple("Check", ("sheet", "column", "row", "expected_values", "type"))

TF_ROUND_4 = {
    "checks": (
        Check(sheet="1 - Start Here", column=6, row=1, expected_values=("v4.3",)),
        Check(sheet="1 - Start Here", column=4, row=1, expected_values=("1 April 2023 to 30 September 2023",)),
        Check(sheet="2 - Project Admin", column=5, row=4, expected_values=("TD", "HS")),
        Check(
            sheet="2 - Project Admin", column=6, row=4, expected_values=tuple(TF_PLACE_NAMES_TO_ORGANISATIONS.keys())
        ),
    )
}
```

core/validation/pre_transformation_validate.py
```diff
@@ -0,0 +1,148 @@
...
+    if str(sheet.iloc[6][4]).strip() in TF_PLACE_NAMES_TO_ORGANISATIONS.keys():
+        authorisation_checks["Place Name"] = (
+            str(sheet.iloc[6][4]).strip(),
+            place_names,
+        )
+    if FUND_TYPE_TO_TD_OR_HS.get(sheet.iloc[5][4], sheet.iloc[5][4]) in {"TD", "HS"}:
+        authorisation_checks["Fund Type"] = (
+            FUND_TYPE_TO_TD_OR_HS.get(sheet.iloc[5][4], sheet.iloc[5][4]),
+            fund_types,
+        )
```
cyrusdobbs, line 59: This above may be harder to achieve here as there is a bit more logic.. there may be a way though

core/validation/pre_transformation_validate.py
```diff
@@ -0,0 +1,148 @@
...
+
+    conflicting_input_checks["Place Name vs Fund Type"] = (
+        FUND_TYPE_TO_TD_OR_HS.get(sheet.iloc[5][4]),
+        tuple(PLACE_TO_FUND_TYPE.get(str(sheet.iloc[6][4]).strip())),
+    )
+
+    failures = check_values(conflicting_input_checks, "wrong_input_failure")
+
+    if failures:
+        raise ValidationError(validation_failures=failures)
```
cyrusdobbs, line 121: This isn't very obvious what it is (or perhaps it's just Friday afternoon.. 😁)

again, it feels very specific - we need to try and refactor this so it can be used for new rounds / spreadsheets

core/validation/pre_transformation_validate.py
```diff
@@ -0,0 +1,148 @@
...
+    failure = str_repr_of_failure_to_actual_failure[failure_type]
+    failures = []
+
+    for value_descriptor, (entered_value, expected_values) in failure_dictionary.items():
+        if entered_value not in expected_values:
+            failures.append(
+                failure(value_descriptor=value_descriptor, entered_value=entered_value, expected_values=expected_values)
+            )
+
+    return failures
```
cyrusdobbs, line 148: This is a similar function to what I expect the config described above would be used in

cyrusdobbs (changes requested):
Looking like an improvement

I think we can go further and really make this reusable and extendable to additional sources (new rounds / funds etc) - these functions will only work for TF Round 4 at the moment

core/validation/pre_transformation_validate.py
Joel-Pearce: In effect, there's a hierarchy of pre-transformations failures. First, we check place_name and fund_type to see if the user is authorised to submit for both. Then, we check place_name and fund_type to see if the values entered into them are valid (i.e. they're part of the dropdown on the spreadsheet for both). Then, we check to see if place_name and fund_type are conflicting (i.e. if the place is Newark but the selected fund type is HS, but Newark is only associated with a TD, then the user could be authorised to submit for the place and the fund type, the place and the fund type could be valid, but they conflict with one another).

We don't want to raise multiple errors on the same cell (or else it might be confusing), so the three functions are checking if authorised > valid > conflicting, and stopping to raise error if any after each one.

core/validation/pre_transformation_validate.py
```diff
@@ -0,0 +1,168 @@
...
+        expected_values = tuple(map(fund_type_acronym_to_full_name.get, fund_type_of_entered_place))
+
+        conflicting_input_checks.append(
+            PreTransformationCheck(sheet=sheet, column=column, row=row, expected_values=expected_values, type=type)
+        )
+
+    failures = check_values(workbook, conflicting_input_checks, "wrong_input_failure")
+
+    if failures:
+        raise ValidationError(validation_failures=failures)
```
cyrusdobbs, line 108: I think this function still needs some more thought.

It's currently still quite unclear as to what it does. It also doesn't look very reusable because it's tightly coupled to the `PLACE_TO_FUND_TYPE` mapping.

This is possibly down to the nature of the check we're trying to do. If it's really specific like this then I'm not sure we should support it here.

core/validation/pre_transformation_validate.py
```diff
@@ -0,0 +1,168 @@
...
+    :param schema: The schema containing conflicting input checks.
+    :raises ValidationError: If conflicting input validation fails.
+    """
+    fund_type_acronym_to_full_name = {"TD": "Town_Deal", "HS": "Future_High_Street_Fund"}
+    conflicting_input_checks = list()
+    for sheet, column, row, expected_values, type in schema["conflicting_input_checks"]:
+        # TODO: if extra conflicting checks introduced, make this this logic dependent on 'type'
+        col, row = expected_values
+        fund_type_of_entered_place = PLACE_TO_FUND_TYPE.get(str(workbook[sheet].iloc[col][row]).strip())
+        expected_values = tuple(map(fund_type_acronym_to_full_name.get, fund_type_of_entered_place))
```
cyrusdobbs, line 99: I think it's generally clearer to use a comprehension instead of using map 
```
expected_values = tuple(fund_type_acronym_to_full_name.get(fund_type) for fund_type in fund_type_of_entered_place)
```
but I don't really like this because it's so TF specific

core/validation/pre_transformation_validate.py
```diff
@@ -0,0 +1,168 @@
...
+
+    for y_axis in [*town_board_chair_cells, *section_151_text_cells]:
+        if pd.isnull(sheet.iloc[y_axis, 2]):
+            failures.append(
+                uf.GenericFailure(
+                    sheet="Review & Sign-Off", section="-", cell_index="C" + str(y_axis + 2), message=msgs.BLANK
+                )
+            )
+
+    return failures
```
cyrusdobbs, line 168: Perhaps this should go somewhere else now? Probably best here `core.validation.specific_validations.towns_fund_round_four`

cyrusdobbs (changes requested):
A really good step in the right direction. 

I think we have a chance now to think past the TF context and I think this is still quite coupled to TFs specifics. We don't know exactly what the future looks like but I think it's safe to assume that we'll need to adapt for another fund now - in which case we'd have to revisit this later down the line - so lets get it right now I think

core/validation/pre_transformation_validate.py
Joel-Pearce: It's to deal with a edge case that turned up in QA prior to the launch of Submit. 

If a user enters a place_name is that one of our valid place_names from our mapping, and they select a valid fund_type, but the place_name does not match the fund_type then this will be flagged, but to get the place_name's fund type we need to do a lookup on the place's fund_type, and then make sure the fund_type is not the acronym but how it appears on the spreadsheet. Unless we envision this mapping incorporating other funds then it won't be reusable; or, at least, the function will have to come with if-statement on 'type'. 

It's actually very difficult to hit this error as the dropdown on the spreadsheet won't let a user select HS for a place only associated with Town_Deal. You have to do some very specific copy-and-pasting to get "Future_High_Street_Fund" in the fund_type box when you've selected Newark as a place. 

Other option is ```towns_fund_round_4_specific_pre_validation.py``` with this function in.

core/validation/pre_transformation_validate.py
```diff
@@ -0,0 +1,168 @@
...
+        expected_values = type_to_authorised_values[type]
+
+        authorisation_checks.append(
+            PreTransformationCheck(sheet=sheet, column=column, row=row, expected_values=expected_values, type=type)
+        )
+
+    failures = check_values(workbook, authorisation_checks, "unauthorised_submission_failure")
+
+    if failures:
+        raise ValidationError(validation_failures=failures)
```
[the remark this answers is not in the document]
Joel-Pearce: Yeh, that makes a lot of sense. In my head I was thinking that place_name and fund_type would be universal variables across funds, but that won't necessarily be the case. This change may well lead to some downtime on prod though. 

If we change the authorisation in data-store to expect a JSON object, then we also have to change Submit's payload structure, but it might be a bit of a chicken-and-the-egg problem in that a change on one repo by itself will cause the other to fall over. There's ways around this but it will require some forethought.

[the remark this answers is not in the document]
cyrusdobbs: I would argue a bit of planned downtime is ok. We can do it at a time of low traffic and the impact will be minimal to non-existent.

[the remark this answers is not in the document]
Joel-Pearce: I raise the issue because @gidsg asked me to write the changes such that they could be swapped in and out without any downtime on production.

Something like this can potentially achieve that:

```
auth = body.get("auth", {"Places Names": body.get("place_names"),
                             "Fund Types": {"Town_Deal", "Future_High_Street_Fund"}})
```

Bit hacky, but then we swap out to ```auth = body.get("auth")``` once the Submit changes are also in. There's also the problem of the private contractor mappings.

If we could get 10 minutes of downtime we can make the switch without this, but what's your opinion @gidsg?

core/validation/pre_transformation_validate.py
Joel-Pearce: I have managed to refactor the function to be fund agnostic. It now has no explicit tie to TF. Quite what we may use this function for in the future is hard to envision without knowing what ingesting another fund actually looks like, but the pre-transformation failures no longer make any explicit reference to TF-only variables.

cyrusdobbs: I think this is an improved implementation - nice one

core/validation/pre_transformation_validation_schema.py
```diff
@@ -0,0 +1,111 @@
...
+            type="Place Name",
+        ),
+    ],
+    "conflicting_input_checks": [
+        ConflictingInputCheck(
+            sheet="2 - Project Admin",
+            column_1=5,
+            row_1=4,
+            column_2=6,
+            row_2=4,
```
cyrusdobbs, line 45: Can we name these more descriptively?

core/validation/pre_transformation_validate.py
```diff
@@ -0,0 +1,131 @@
...
+    failures = []
+
+    for sheet, column, row, expected_values, type in failure_list:
+        if workbook[sheet].iloc[column][row] not in expected_values:
+            failures.append(
+                failure(
+                    value_descriptor=type,
+                    entered_value=workbook[sheet].iloc[column][row],
+                    expected_values=expected_values,
+                )
```
cyrusdobbs, line 128: Bit nit picky here but.. We currently do `workbook[sheet].iloc[column][row]` twice 

I think it also adds clarity to pull it out as a variable like 

```suggestion
    for sheet, column, row, expected_values, type in failure_list:
        entered_value = workbook[sheet].iloc[column][row]
        if  entered_value  not in expected_values:
            failures.append(
                failure(
                    value_descriptor=type,
                    entered_value=entered_value,
                    expected_values=expected_values,
                )
```

cyrusdobbs (commented):
Apologies for drawing this out a bit now - just a few small things and good to go

core/validation/pre_transformation_validate.py
```diff
@@ -0,0 +1,131 @@
...
+    :param reporting_round: The reporting round for which the validations are performed.
+    :param auth: A dictionary containing authorization information.
+    :raises ValidationError: If any of the validation functions catch a validation error.
+    """
+    if reporting_round in [1, 2]:
+        return  # do not do pre-transformation validation on historical data
+
+    round_to_schema = {3: TF_ROUND_3, 4: TF_ROUND_4}
+
+    schema = round_to_schema[reporting_round]
```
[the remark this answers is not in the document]
Joel-Pearce: Can set as a dictionary in ```validation/const.py``` (I originally brought it into the function as was causing a circular import error on ```core/const.py```.

Presumably reporting round will reset to 1 for every new fund, and so we'll have to have a two keys on the dictionary look-up in the future, but without entirely knowing what that looks like it's hard to be more specific now.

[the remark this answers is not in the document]
cyrusdobbs: Yep I agree we'll end up adding `fund` as a parameter later but if we abstract this logic out then it will mean less changes to this function when we do that.. perhapse this is fine for now though - we'll have to make a few more changes here when those changes come anyway

core/const.py
```diff
@@ -1080,3 +1078,11 @@ class ITLRegion(StrEnum):
...
     3: "O{i}",
 }
+
+# tuple used in the schema for pre-transformation checks
+PreTransformationCheck = namedtuple("Check", ("sheet", "column", "row", "expected_values", "type"))
+
+# tuple used in the schema for pre-transformation checks to check if two types of input conflict
+ConflictingInputCheck = namedtuple(
+    "Check", ("sheet", "column", "row", "column_of_value_to_be_mapped", "row_of_value_to_be_mapped", "mapping", "type")
+)
```
cyrusdobbs, line 1088: I think these are best placed in `core.validation.const`

core/validation/const.py
```diff
@@ -1,5 +1,10 @@
 import numpy as np
 import pandas as pd
 
+from core.validation.pre_transformation_validation_schema import TF_ROUND_3, TF_ROUND_4
+
 # values we treat as null
 NA_VALUES = {"", np.nan, None, pd.NA}
+
+# reporting round to schema for pre-transformation validation
+REPORTING_ROUND_TO_PRE_TRANSFORMATION_SCHEMA = {3: TF_ROUND_3, 4: TF_ROUND_4}
```
cyrusdobbs, line 10: That will cause a circular import with this though so this will need to be moved - I think `core/validation/pre_transformation_validation_schema.py` may be best?

cyrusdobbs (approved):
This is a big improvement - nice one 👍

core/validation/const.py
Joel-Pearce: OK, can move both to core/validation/pre_transformation_validation_schema.py

cyrusdobbs: Sounds good

PR #363, opened 2023-11-14
[SMD-387] during extract do lookups with stripped and lowered values

cyrusdobbs:
https://dluhcdigital.atlassian.net/jira/software/c/projects/SMD/boards/140?selectedIssue=SMD-387

### Change description
- use `strip` and `lower` when using lookups during extraction 

### How to test

### DIFF
--- a/core/extraction/towns_fund_round_three.py
+++ b/core/extraction/towns_fund_round_three.py
@@ -133,7 +133,7 @@ def extract_project_lookup(df_lookup: pd.DataFrame, df_place: pd.DataFrame) -> d
     df_lookup.columns = ["Unique Project Identifier", "Town", "Project Name"]
 
     # filter on current place / programme and convert to dict
-    df_lookup = df_lookup.loc[df_lookup["Town"] == place_name]
+    df_lookup = df_lookup.loc[df_lookup["Town"].str.lower().str.strip() == str(place_name).lower().strip()]
     project_lookup = dict(zip(df_lookup["Project Name"], df_lookup["Unique Project Identifier"]))
 
     return project_lookup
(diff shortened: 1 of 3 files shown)


Connor-OS (approved):
LGTM

core/extraction/towns_fund_round_three.py
```diff
@@ -133,7 +133,7 @@ def extract_project_lookup(df_lookup: pd.DataFrame, df_place: pd.DataFrame) -> d
     df_lookup.columns = ["Unique Project Identifier", "Town", "Project Name"]
 
     # filter on current place / programme and convert to dict
-    df_lookup = df_lookup.loc[df_lookup["Town"] == place_name]
+    df_lookup = df_lookup.loc[df_lookup["Town"].str.lower().str.strip() == str(place_name).lower().strip()]
```
Joel-Pearce, line 136: Just checking my understanding here.

If the user has a place_name of "BARNSLEY ", this lookup will work, but then it will fall over when we do the lookup to get the org name:

```return TF_PLACE_NAMES_TO_ORGANISATIONS[place_value.strip()]```

as we don't .lower() here. So what we might actually want is .capitalize().strip() to prevent there being a mismatch here (all Place Names in the look up are capitalised but now lower)?

Joel-Pearce: But it won't match the place names in the look-up in the case of Newcastle-under-lyme have the U or L interchangeably lower and uppercase

cyrusdobbs: Yep we never change the value in the "place name" cell - we just normalise it, and all lookup values, when we use either the place or project lookups. 

That mapping already covers all those edge cases so I think we're ok with that one

PR #367, opened 2023-11-16
SMD-361 return list of exception types for front end logging

bobbi-hbrown:
https://dluhcdigital.atlassian.net/jira/software/c/projects/SMD/boards/140?quickFilter=815&selectedIssue=SMD-427

Adds an extra list comprehension containing the Validation Exception types so they can be sent to the front end - they are not coupled with each exception themselves, but they can be if required.

### DIFF
--- a/core/handlers.py
+++ b/core/handlers.py
@@ -29,6 +29,7 @@ def handle_validation_error(validation_error: ValidationError) -> Exception | tu
         "title": "Bad Request",
         "pre_transformation_errors": validation_messages.get("pre_transformation_errors", []),
         "validation_errors": validation_messages.get("validation_errors", []),
+        "error_types": validation_messages.get("error_types", []),
     }, 400
 
 
--- a/core/handlers.py
+++ b/core/handlers.py
@@ -29,7 +29,6 @@ def handle_validation_error(validation_error: ValidationError) -> Exception | tu
         "title": "Bad Request",
         "pre_transformation_errors": validation_messages.get("pre_transformation_errors", []),
         "validation_errors": validation_messages.get("validation_errors", []),
-        "error_types": validation_messages.get("error_types", []),
     }, 400
 
 
--- a/core/validation/failures/user.py
+++ b/core/validation/failures/user.py
@@ -295,10 +295,11 @@ def failures_to_messages(
 ) -> dict[str, list[str]] | dict[str, list[dict[str, str | None]]]:
     """Serialises failures into messages, removing any duplicates, and groups them by tab and section.
     :param validation_failures: validation failure objects
-    :return: validation failure messages grouped by tab and section
+    :return: validation failure messages grouped by tab and section and list of error types
     """
     # filter and convert to error messages
     error_messages = [failure.to_message() for failure in validation_failures]
+    error_types = [failure.__class__.__name__ for failure in validation_failures]
 
     # one pre-transformation failure means payload is entirely pre-transformation failures
     if any(isinstance(failure, PreTransFormationFailure) for failure in validation_failures):
@@ -312,13 +313,12 @@ def failures_to_messages(
     # group cells by sheet, section and desc
     error_messages = group_validation_messages(error_messages)
     error_messages.sort()
-
     validation_errors = [
         {"sheet": sheet, "section": section, "cell_index": cell_index, "description": message}
         for sheet, section, cell_index, message in error_messages
     ]
 
-    return {"validation_errors": validation_errors}
+    return {"validation_errors": validation_errors, "error_types": error_types}
 
 
 def group_validation_messages(validation_messages: list[tuple[str, str, str, str]]) -> list[tuple[str, str, str, str]]:
--- a/core/validation/failures/user.py
+++ b/core/validation/failures/user.py
@@ -298,13 +298,12 @@ def failures_to_messages(
     :return: validation failure messages grouped by tab and section and list of error types
     """
     # filter and convert to error messages
-    error_messages = [failure.to_message() for failure in validation_failures]
-    error_types = [failure.__class__.__name__ for failure in validation_failures]
+    error_messages = [(*failure.to_message(), failure.__class__.__name__) for failure in validation_failures]
 
     # one pre-transformation failure means payload is entirely pre-transformation failures
     if any(isinstance(failure, PreTransFormationFailure) for failure in validation_failures):
         # ignore tab and section for pre-transformation failures
-        return {"pre_transformation_errors": [message for _, _, message in error_messages]}
+        return {"pre_transformation_errors": [message for _, _, message, _ in error_messages]}
 
     error_messages = remove_errors_already_caught_by_null_failure(error_messages)
     # remove duplicates resulting from melted rows where we are unable to remove duplicates at time of validation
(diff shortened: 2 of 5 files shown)


core/validation/failures/user.py
```diff
@@ -295,10 +295,11 @@ def failures_to_messages(
 ) -> dict[str, list[str]] | dict[str, list[dict[str, str | None]]]:
     """Serialises failures into messages, removing any duplicates, and groups them by tab and section.
     :param validation_failures: validation failure objects
-    :return: validation failure messages grouped by tab and section
+    :return: validation failure messages grouped by tab and section and list of error types
     """
     # filter and convert to error messages
     error_messages = [failure.to_message() for failure in validation_failures]
+    error_types = [failure.__class__.__name__ for failure in validation_failures]
```
Joel-Pearce, line 302: I think some test coverage on this would be good - even if it's only a relatively small piece of code

core/validation/failures/user.py
```diff
@@ -312,13 +313,12 @@ def failures_to_messages(
     # group cells by sheet, section and desc
     error_messages = group_validation_messages(error_messages)
     error_messages.sort()
-
     validation_errors = [
```
Connor-OS, line 316: Could you put the error_type as an attribute of the validation error. This groups them and should remove the need for your submit PR

Connor-OS: I can see two ways to do this:
- Add error type as a return value from ever validation error and change all tests to reflect this - Lots of work
- Add error types into the validation errors here - Quick, but a bit messy
```
for validation_error, error_type in zip(validation_errors,error_types):
    validation_error["type"] = error_type
```

Connor-OS (changes requested):
I think the error type should be an attribute of the validation error. But I can see that retro fitting that could be annoying and time consuming

core/validation/failures/user.py
```diff
@@ -295,10 +295,11 @@ def failures_to_messages(
 ) -> dict[str, list[str]] | dict[str, list[dict[str, str | None]]]:
     """Serialises failures into messages, removing any duplicates, and groups them by tab and section.
     :param validation_failures: validation failure objects
-    :return: validation failure messages grouped by tab and section
+    :return: validation failure messages grouped by tab and section and list of error types
     """
     # filter and convert to error messages
     error_messages = [failure.to_message() for failure in validation_failures]
```
cyrusdobbs, line 301: I think a an improvement could be to add the `class` to the message tuple instead, like

```
error_messages = [(*failure.to_message(), failure.__class__.__name__) for failure in validation_failures]
```

this way we keep all the information for a single failure together and minimise the changes to the return structure of `/ingest`

core/validation/failures/user.py
cyrusdobbs: The difficulty with this is that some errors are combined in a previous step so they won't always line up to be zipped like this

core/validation/failures/user.py
cyrusdobbs: I didn't see @Connor-OS's comment before I added this. We're alluding to the same point

cyrusdobbs:
Just a reminder that the OpenAPI documentation should also be updated in line with this change

https://github.com/communitiesuk/funding-service-design-post-award-data-store/blob/96ce32bc6bb2e5c8d45fc1506eadffa059752350/openapi/components.yml#L165-L182

core/validation/failures/user.py
```diff
@@ -376,14 +378,14 @@ def normalise_risk_section_name(section):
     null_messages = [msgs.BLANK, msgs.BLANK_ZERO, msgs.BLANK_PSI, msgs.BLANK_UNIT_OF_MEASUREMENT]
 
     null_failures = [
-        (sheet, section, cell_index.strip(), message)
-        for sheet, section, cell_index, message in errors
+        (sheet, section, cell_index.strip(), message, _)
+        for sheet, section, cell_index, message, _ in errors
```
cyrusdobbs, line 382: We should use a variable name here (and throughout this function). Good practice to use `_` if you don't use the variable that's unpacked

tests/validation_tests/test_failures.py
```diff
@@ -435,28 +435,40 @@ def test_failures_to_messages():
...
         # C - do not combine these due to different sections
-        ("Risk Register", "Project Risks - Project 1", "G24", "Select from the dropdown."),
-        ("Risk Register", "Project Risks - Project 2", "G43", "Select from the dropdown."),
+        ("Risk Register", "Project Risks - Project 1", "G24", "WrongInputFailure", "Select from the dropdown."),
+        ("Risk Register", "Project Risks - Project 2", "G43", "WrongInputFailure", "Select from the dropdown."),
         # D - do not combine these due to different descriptions
-        ("Outcomes", "Programme-level Outcomes", "E5", "You left cells blank."),
-        ("Outcomes", "Programme-level Outcomes", "E7", "Select from the dropdown."),
+        ("Outcomes", "Programme-level Outcomes", "E5", "WrongInputFailure", "You left cells blank."),
+        ("Outcomes", "Programme-level Outcomes", "E7", "WrongInputFailure", "Select from the dropdown."),
```
cyrusdobbs, line 454: Looks like your tests are failing because the `error_type` needs to be the last value in the tuple, with `message` as the penultimate one

core/handlers.py
```diff
@@ -29,6 +29,7 @@ def handle_validation_error(validation_error: ValidationError) -> Exception | tu
         "title": "Bad Request",
         "pre_transformation_errors": validation_messages.get("pre_transformation_errors", []),
         "validation_errors": validation_messages.get("validation_errors", []),
+        "error_types": validation_messages.get("error_types", []),
```
cyrusdobbs, line 32: This can be removed

core/validation/failures/user.py
```diff
@@ -295,15 +295,15 @@ def failures_to_messages(
 ) -> dict[str, list[str]] | dict[str, list[dict[str, str | None]]]:
     """Serialises failures into messages, removing any duplicates, and groups them by tab and section.
     :param validation_failures: validation failure objects
-    :return: validation failure messages grouped by tab and section
+    :return: validation failure messages grouped by tab and section and list of error types
```
cyrusdobbs, line 298: Can revert this

cyrusdobbs (commented):
Looking better now 👍

core/validation/failures/user.py
```diff
@@ -392,11 +394,10 @@ def normalise_risk_section_name(section):
         for error in errors
         if not any(
             ((sheet, normalise_risk_section_name(section), cell_index.strip()) in unique_identifiers_from_null_failures)
-            for sheet, section, cell_index, _ in [error]
+            for sheet, section, cell_index, _, error_type in [error]
```
cyrusdobbs, line 397: this can be `_` as we don't use `error_type`

core/validation/failures/user.py
```diff
@@ -376,14 +378,14 @@ def normalise_risk_section_name(section):
...
-        for sheet, section, cell_index, message in errors
+        (sheet, section, cell_index.strip(), message, error_type)
+        for sheet, section, cell_index, message, error_type in errors
         if message in null_messages
     ]
 
     unique_identifiers_from_null_failures = {
         (sheet, normalise_risk_section_name(section), cell_index.strip())
-        for sheet, section, cell_index, _ in null_failures
+        for sheet, section, cell_index, _, error_type in null_failures
```
cyrusdobbs, line 388: > this can be `_` as we don't use `error_type`

same here

cyrusdobbs (approved):
LGTM

PR #369, opened 2023-11-16
Feature/fmd 91 f s ability to re ingest original files from db

Connor-OS:
Adds an endpoint to retrieve submission spreadsheets from the DB. This is the first step to a re-ingestion process that will join file retreval to the ingest endpoint to reingest files in the DB

Only tested with round 3 and 4 data so far. May be more complexity with historical data

### Change description

### How to test


core/controllers/retrieve.py
```diff
@@ -0,0 +1,42 @@
...
+
+    response = make_response(file_bytes)
+    response.headers.set("Content-Type", EXCEL_MIMETYPE)
+    response.headers.set(
+        "Content-Disposition",
+        "attachment",
+        filename=f"{'_'.join(programme_name.split())}_round_{reporting_round}_submission.xlsx",
+    )
+
+    return response
```
cyrusdobbs, line 42: [flask.send_file](https://tedboy.github.io/flask/generated/flask.send_file.html) may be a cleaner way to do this

core/controllers/retrieve.py
```diff
@@ -0,0 +1,42 @@
...
+        .join(ProgrammeProgress, Submission.id == ProgrammeProgress.submission_id)
+        .join(Programme, ProgrammeProgress.programme_id == Programme.id)
+        .where(Programme.programme_name == programme_name)
+        .where(Submission.reporting_round == reporting_round)
+    )
+
+    file_bytes = db.session.scalars(query).first()
+
+    if not file_bytes:
+        return abort(400, "No file was found that matches these criteria")
```
cyrusdobbs, line 32: I think `404` is a more relevant HTTP error code

core/controllers/retrieve.py
```diff
@@ -0,0 +1,42 @@
...
+    reporting_round = request.args.get("reporting_round")
+    programme_name = request.args.get("programme_name")
+
+    query = (
+        select(Submission.submission_file)
+        .join(ProgrammeProgress, Submission.id == ProgrammeProgress.submission_id)
+        .join(Programme, ProgrammeProgress.programme_id == Programme.id)
+        .where(Programme.programme_name == programme_name)
+        .where(Submission.reporting_round == reporting_round)
+    )
```
cyrusdobbs, line 27: I suppose we have to join to a programme-level table with our current data model... 

We could leave a comment to point this out as it's arbitrary which one we join to and is not clear on the face of it?

cyrusdobbs (commented):
Looks good - just a few minor comments

One other thought was that we could make this fund agnostic now so we don't have to come back later..

I think we could include a param for `fund_type_id` to achieve this? We'd also want to remove the `enum` validation for `reporting_round`

tests/controller_tests/test_retrieve.py
```diff
@@ -0,0 +1,17 @@
...
+
+def test_invalid_reporting_round(seeded_test_client):
+    response = seeded_test_client.get("/retrieve?reporting_round=1&programme_name=Leaky Cauldron regeneration")
+    assert response.status_code == 400
+
+
+def test_retrieval(seeded_test_client):
+    response = seeded_test_client.get("/retrieve?reporting_round=3&programme_name=Leaky Cauldron regeneration")
+    assert response.status_code == 200
+    assert response.content_type == EXCEL_MIMETYPE
```
Joel-Pearce, line 17: Is it possible to do some more granular assertions than just the status_code and content_type?

core/controllers/retrieve.py
```diff
@@ -0,0 +1,42 @@
...
+    :return: Flask response object containing the requested spreadsheet.
+    """
+    reporting_round = request.args.get("reporting_round")
+    programme_name = request.args.get("programme_name")
+
+    query = (
+        select(Submission.submission_file)
+        .join(ProgrammeProgress, Submission.id == ProgrammeProgress.submission_id)
+        .join(Programme, ProgrammeProgress.programme_id == Programme.id)
+        .where(Programme.programme_name == programme_name)
```
Joel-Pearce, line 25: I might be wrong here as it has been a while since I worked on the db, but programme_name isn't unique, and as it's just the name of the place, isn't it possible to have two programmes with the same name in the same round (e.g. "Rotherham" for both TD and HS)? The unique key in Programme is programme_id and not name

Joel-Pearce (changes requested):

core/controllers/retrieve.py
Joel-Pearce: But even if that's not the case I'm not sure "reporting_round" and "programme_name" necessarily form a composite key

Connor-OS: Yes, I agree. I'm not fully happy with this way of selecting files, and if we can have duplicate keys I should change it. I was trying to think of how we would like to retrieve these files and how to automate retrieving whole rounds of returns at once. Originally I wanted it to be human readable so I went with place_name and reporting round. Since this isn't going to be user facing prehaps I can just go for programme or submission ID.... Thinking out loud here

core/controllers/retrieve.py
Connor-OS: I spent some time trying to think this one through. I think the answer might be to select files by submission_id instead, this might simplify the process.

Connor-OS: it does seem like we could have a programme_id field in the submission table to allow this kind of look up though

core/controllers/retrieve.py
Joel-Pearce: I think any select statement that expects a single value using a where clause has to use it on a unique key (or something assume that the first value being returned will always be the correct one). 

Any serious db debugging requires us to read the UUIDs anyway, there's not much getting around it.

Joel-Pearce:
Could we call ```remove_unreferenced_organisations()``` at the end of re-ingest script?

core/controllers/retrieve.py
```diff
@@ -0,0 +1,33 @@
...
+    """Handle the download request and return the originally submitted spreadsheet.
+
+    Select file by:
+    - UUID
+
+    :return: Flask response object containing the requested spreadsheet.
+    """
+    submission_UUID = request.args.get("submission_UUID")
+
+    query = select(Submission.submission_file, Submission.reporting_round).where(Submission.id == submission_UUID)
```
cyrusdobbs, line 22: I thought we decided to go with Submission ID rather than UUID?

core/controllers/retrieve.py
```diff
@@ -0,0 +1,33 @@
...
+
+    try:
+        file_bytes, reporting_round = db.session.execute(query).one()
+    except Exception:
+        return abort(404, "Could not find a file that matches this UUID")
+
+    return flask.send_file(
+        BytesIO(file_bytes), mimetype=EXCEL_MIMETYPE, as_attachment=True, download_name=f"{submission_UUID}.xlsx"
+    ), {
+        "reporting_round": reporting_round
```
cyrusdobbs, line 32: Does this work? You can only send one request - this looks like two?

scripts/re-ingest.py
```diff
@@ -0,0 +1,118 @@
...
+
+    :param file_path: Path to the file to be posted.
+    :param endpoint_url: URL of the endpoint to post the file to.
+    :return: Response from the server.
+    """
+    response = requests.post(
+        endpoint_url,
+        files={
+            "excel_file": (
+                f"reingest{datetime.now().strftime('%Y%m%dT%H%M%S%z')}",
```
cyrusdobbs, line 56: Are we happy to overwrite existing file names with this? It will update the `Submission` table to this so we will lose the original file name. Maybe this is fine? Maybe it also makes saving the filename redundant..

tests/controller_tests/test_retrieve.py
```diff
@@ -0,0 +1,19 @@
...
+    response = seeded_test_client.get(f"/retrieve?submission_UUID={invalid_UUID}&attachment=true")
+    assert response.status_code == 404
+
+
+def test_retrieve(seeded_test_client):
+    submission_UUID = db.session.scalar(select(Submission.id))
+    response = seeded_test_client.get(f"/retrieve?submission_UUID={submission_UUID}&attachment=false")
+    assert response.status_code == 200
+    assert response.headers["content-type"] == "application/vnd.openxmlformats-officedocument.spreadsheetml.sheet"
+    assert response.content_type == EXCEL_MIMETYPE
```
cyrusdobbs, line 19: Looks like these should be expanded to assert the rest of the response that is in the schema

cyrusdobbs (commented):
Looking good

Connor-OS:
> Could we call `remove_unreferenced_organisations()` at the end of re-ingest script?

The point of the script is that it runs without the application context so doesn't have a direct connection to the database. It really just links the two endpoints /retrieve and /ingest together. Therfore I don't think calling `remove_unreferenced_organisations()` at that stage will work

core/controllers/retrieve.py
Connor-OS: Yes, I'm swapping over to use submission_id instead

core/controllers/retrieve.py
Connor-OS: This works, the reporting round ends in the response headers, we will need to also return fund_type, because that will be expected by the ingest function as well

cyrusdobbs: Nice 👍

scripts/re-ingest.py
Connor-OS: I've changed this to pull through the file_name stored in the DB. This is hard to test though because in the mock example the file name seems to be: `'C:\\Users\\OShaughnessyC\\Documents\\python\\funding-service-design-post-award-data-store\\tests\\resources\\Post_transform_EXAMPLE_data.xlsx'` but obviously different on every machine

cyrusdobbs: oh that's not intended. 

We should be just using the file name rather than the full address. 

You could probably include a small fix in this PR to use `os.path.basename` when we save it to the db

Connor-OS: I would do this now but it actually seems like Noels lates PR is going to change this example data mock in quite a large way. I'll probably wait and see how this works once the test refactors are fully merged

openapi/api.yml
```diff
@@ -83,6 +83,36 @@ paths:
...
 
+  /retrieve:
+    get:
+      operationId: core.controllers.retrieve.retrieve
+      summary: retrieve specific originally uploaded files from the data-store
+      description: takes an ID and reporting round to retrieve a file
+      parameters:
+        - in: query
+          name: submission_id
+          example: "S-R04-01"
```
cyrusdobbs, line 94: nitpick but the first submission would be in the format `S-R04-1`

openapi/api.yml
```diff
@@ -83,6 +83,36 @@ paths:
...
+                type: string
+                format: binary
+                description: The file content in a binary format.
+            application/json:
+              schema:
+                type: object
+                properties:
+                  # 'file' will be the field name in this multipart request
+                  file_name:
+                    type: string
```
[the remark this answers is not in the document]
cyrusdobbs: We can better describe the headers here though so that the docs truly reflect what `flask.send_file` sets the headers as

cyrusdobbs (changes requested):
Looking good - just a few minor things to correct

openapi/api.yml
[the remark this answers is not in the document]
Connor-OS: you are right about it already being in the headers, but I got into a weird situation where it seems to be very difficult to get it out of the headers and I ended up using a line like:

```response.headers.get("Content-Disposition").split("filename=")[1]```

There must be a cleaner way

File: core/controllers/retrieve.py [hunk 12]
```diff
@@ -0,0 +1,34 @@
...
+
+def retrieve():
+    """Handle the download request and return the originally submitted spreadsheet.
+
+    Select file by:
+    - submission_id
+
+    :return: Flask response object containing the requested spreadsheet.
+    """
+    submission_id = request.args.get("submission_id")
```
cyrusdobbs, line 21: Using Connexion, we can put `submission_id` as a function param and it will grab it from the request rather than having to fetch it from the `Flask` request object.

See https://connexion.readthedocs.io/en/stable/request.html

openapi/api.yml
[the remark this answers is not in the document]
cyrusdobbs: Looks like this is unchanged?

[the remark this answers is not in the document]
cyrusdobbs: Sorry I've only just seen your comment. I see what you mean... In that case I don't mind us also sending the file name in header as it's own KV. Either way, this OpenAPI doc should be made accurate.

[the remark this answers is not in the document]
Connor-OS: I think I actually prefer just sending the file. It is neater for the endpoint and the file name retreval is only for the script so I don't mind using the method above

core/controllers/retrieve.py
Connor-OS: Thanks for this comment it was very informative in deciding how to build this endpoint

Joel-Pearce (approved):
LGTM

PR #373, opened 2023-11-21
[SMD-404]: ingest pipeline tests

Joel-Pearce:
Write tests for the /ingest endpoint with files corresponding to the actual TF R4 template filled with dummy data. Hits every function that raises a ```ValidationFailure```.


cyrusdobbs (commented):
Nice work, this will be handy

I think we plan to move all integration tests (that call endpoints or commit to the DB) to a new subfolder under tests - we could put these there ready so it's one less thing to move in that PR

-->tests 
-----> integration_tests        <----- these go here
-----> unit_tests

PR #375, opened 2023-11-23
[SMD-403] handle internal validation errors

cyrusdobbs:
https://dluhcdigital.atlassian.net/browse/SMD-403

### Change description
- renames DevFailures -> InternalFailures
- logs internal Failures as an error and saves the Submission to S3 as with uncaught exceptions

### How to test


core/handlers.py
```diff
@@ -32,27 +45,30 @@ def handle_validation_error(validation_error: ValidationError) -> Exception | tu
...
+    """Saves the file that fails ingest to S3 and returns a generic uncaught exception 500 error.
+
+    :return: A 500 error response
+    """
+    failure_uuid = uuid.uuid4()
+    s3_object_name = FAILED_FILE_S3_NAME_FORMAT.format(failure_uuid, datetime.now().strftime(DATETIME_ISO_8601))
+    upload_file(file=g.excel_file, bucket=Config.AWS_S3_BUCKET_FAILED_FILES, object_name=s3_object_name)
+    current_app.logger.error(f"Uncaught ingest exception - failure_id={str(failure_uuid)}", exc_info=True)
+    return {
+        "detail": "Uncaught ingest exception.",
```
Connor-OS, line 70: Can we return any more detail about the error in this 500 message? Can we pass {internal_failures} from the calling function and return it here. Might help when building a pipeline to see them here

Connor-OS (commented):
Looks good, only one minor comment. Seems like a more sensible way to handle these errors

core/handlers.py
cyrusdobbs: Good shout - I've added this so it returns as part of the response

tests/integration_tests/test_ingest_component.py
```diff
@@ -90,6 +100,54 @@ def test_ingest_with_r4_file_success(test_client_function, towns_fund_round_4_fi
...
+        "loaded": True,
+        "metadata": {
+            "FundType_ID": "HS",
+            "Organisation": "Worcester City Council",
+            "Programme ID": "HS-WRC",
+            "Programme Name": "Blackfriars - Northern City Centre",
+        },
+        "status": 200,
+        "title": "success",
+    }
```
Joel-Pearce, line 128: I'm not sure there's much value in having a function for both ```do_load=True``` and ```do_load=False``` on a successful file. The only reason I turned off the load parameter for that test was to save time.  ```do_load=True``` is probably a more illustrative use case so we can remove the previous ```test_ingest_with_r4_file_success``` function I think.

Connor-OS (approved):
LGTM 👍

Joel-Pearce (changes requested):

PR #379, opened 2023-11-24
[FMD-139] refactor DB test fixture scope

nsnape:
https://dluhcdigital.atlassian.net/browse/FMD-139

### Change description

- Updated "seed" method to use csv files loaded directly into the DB (rather than using the app ingest endpoints via app context)
- Set app context / DB pre-population fixtures to module scope to prevent tear-down after every test
- Added rollback fixtures to revert any non-committed DB changes in tests
- Refactored CLI seed methods to use new DB population methods.

### How to test


core/util.py
```diff
@@ -89,3 +95,37 @@ def get_project_number_by_id(project_id: str, active_project_ids: list[str]) ->
...
+        "project_progress",
+        "funding",
+        "funding_comment",
+        "private_investment",
+        "output_data",
+        "outcome_data",
+        "risk_register",
+    ]:
+        table_df = pd.read_csv(resources / f"{table}.csv")
+        table_df.to_sql(table, con=db.session.connection(), index=False, index_label="id", if_exists="append")
```
Joel-Pearce, line 130: Just clarifying my understanding. With ```if_exists="append"``` we expect for the purposes of this function that there may already be data in the db, and that whatever data is currently in the db we don't want to remove or replace. Would  ```if_exists="replace"``` not be better if we want to ensure a clean dataset for the purposes of testing?

tests/conftest.py
```diff
@@ -318,6 +313,11 @@ def additional_test_data():
 
 @pytest.fixture(scope="function")
 def example_data_model_file() -> BinaryIO:
-    """An example spreadsheet in towns-fund schema format."""
+    """
+    An example spreadsheet in towns-fund schema format.
+
+    This format of Excel file is not really used for anything other than testing - this fixture should not be used
+    to replicate real form ingest behaviour
+    """
```
Joel-Pearce, line 321: We now have permission to store the actual R3 templates in the repo with dummy data in. Currently we have a bunch of R4 templates + dummy data, and I'm intending to put some dummy data in a legitimate R3 template and write a test with it. At that point, I suspect we can remove this entirely as it will be redundant.

Joel-Pearce (changes requested):
Just a thought about the csvs - now we're able to hold legitimate templates in the repo, we can just use one of the existing methods in integration_tests/test_ingest_component.py to populate the db (if you rebase this branch, one has been added with loading behaviour). We can also create our own legitimate R3 template w/ dummy data file. 

It's probably better to just load the data in the DB with the csvs though as if anything goes awry with our ingestion pipeline then we'd be prevented from seeding the DB. Also the csvs should be easier to read.

core/util.py
nsnape: I think I'd prefer to keep it as is, as the whole point of this refactor is there should be either a clean database or some pre-existing populated data (depending on the fixture used). If somehow test data "leaks" through from another test via incorrect implementation of fixture scope, using "append" rather than "replace" is more likely to cause this to error, at least in my mind (it falling over is preferable to just hiding the problem, so at least we spot it happening). This hopefully would help avoid the kind of situation we had before where tests were accidentally sharing DB changes from each other?

Joel-Pearce: "help avoid the kind of situation we had before where tests were accidentally sharing DB changes from each other"

Wouldn't using replace be more likely to ensure this rather than append, as append will retain db changes from one test to another?

Joel-Pearce: Oh, sorry, just re-read. We want the error the occur rather than masking it. Makes sense.

Joel-Pearce (approved):
LGTM

core/util.py
nsnape: Also, replace drops the entire table. Which means that our module scope seed fixture will have effectively been wiped during a single test and the other tests that need that seed data will not be able to see it. It also means we then recreate the table each test, which is effectively what the refactor was replacing (with only creating the table at module setup)

PR #382, opened 2023-11-28
add auth default to api

Joel-Pearce:
Adds a default value in the ingest API for auth.


Connor-OS (approved):
LGTM, thanks for adding

cyrusdobbs (commented):
I've noticed that we also need a check to only run auth if we pass the values

Here seems sensible:
https://github.com/communitiesuk/funding-service-design-post-award-data-store/blob/e99f774923a6a24e4cb8e69f736dae9df800254e/core/validation/pre_transformation_validate.py#L35
to
```
if reporting_round == 4 and auth:
```

cyrusdobbs (approved):
nice one adding this

PR #384, opened 2023-11-29
[BAU] add robustness to dropping of null and merged columns in fundin…

cyrusdobbs:
…g questions extraction

Background:

The "Yes" dropdown value spread across the 8 columns in the picture below (some of which are merged) only appears in the pandas DataFrame in the first column of every merged set of cells in all of our test data and dummy data.

However, when a particular production submission is ingested, it puts the "Yes" value across ALL merged cells. This then changed the output of a subsequent dropna .

We've made the code robust to this in this one case now but we couldn't decipher what causes this as we can't reproduce it in a new spreadsheet (even with the same data from prod etc)..

image

### Change description
- makes the extraction code for funding question robust against the issue above by dropping Excel cell-merged columns based on the columns lack of header value rather than requiring the whole column to be null to be dropped

### How to test
Difficult to test as we can't reproduce it outside of some production data that caused this in the first instance. That submission now returns a valid response when passed to ingest.

### DIFF
--- a/core/extraction/towns_fund_round_three.py
+++ b/core/extraction/towns_fund_round_three.py
@@ -390,7 +390,9 @@ def extract_funding_questions(df_input: pd.DataFrame, programme_id: str) -> pd.D
         # return empty dataframe if fund_type is Future High Street Fund
         return pd.DataFrame(columns=["Question", "Guidance Notes", "Indicator", "Response", "Programme ID"])
 
-    df_input = df_input.iloc[12:19, 2:13].dropna(axis=1, how="all")
+    df_input = df_input.iloc[12:19, 2:13]
+    first_row = df_input.iloc[0]
+    df_input = df_input.dropna(axis=1, how="all", subset=[first_row.name])
 
     # Use the first row as the column headers
     fund_questions_df = df_input.rename(columns=df_input.iloc[0]).iloc[1:]


Joel-Pearce (changes requested):
To test this we either need:

1) an Excel file that recreates the error this fixes
2) to write a test specifically for ```extract_funding_questions``` in which we mock-up ```df_input```
3) put the code change into a method called something like, ```drop_cols_where_col_value_null```

The first two are a bit too onerous for a small code change, and given we can't even recreate the error perhaps the 1st isn't possible. However, maybe we can put the following lines of code in a function:

```
first_row = df_input.iloc[0]
df_input = df_input.dropna(axis=1, how="all", subset=[first_row.name])
```

There's a possibility we might need to use this functionality again.

cyrusdobbs:
> To test this we either need:
> 
> 1. an Excel file that recreates the error this fixes
> 2. to write a test specifically for `extract_funding_questions` in which we mock-up `df_input`
> 3. put the code change into a method called something like, `drop_cols_where_col_value_null`
> 
> The first two are a bit too onerous for a small code change, and given we can't even recreate the error perhaps the 1st isn't possible. However, maybe we can put the following lines of code in a function:
> 
> ```
> first_row = df_input.iloc[0]
> df_input = df_input.dropna(axis=1, how="all", subset=[first_row.name])
> ```
> 
> There's a possibility we might need to use this functionality again.

I'd argue that we've shifted our position and that we don't plan to reuse code between pipelines. I also think testing two lines of pandas code is a bit redundant.

I'm happy to push this through as it just fixes an edge case and our tests cover that it doesn't break the main cases.

Joel-Pearce:
> > To test this we either need:
> > 
> > 1. an Excel file that recreates the error this fixes
> > 2. to write a test specifically for `extract_funding_questions` in which we mock-up `df_input`
> > 3. put the code change into a method called something like, `drop_cols_where_col_value_null`
> > 
> > The first two are a bit too onerous for a small code change, and given we can't even recreate the error perhaps the 1st isn't possible. However, maybe we can put the following lines of code in a function:
> > ```
> > first_row = df_input.iloc[0]
> > df_input = df_input.dropna(axis=1, how="all", subset=[first_row.name])
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > There's a possibility we might need to use this functionality again.
> 
> I'd argue that we've shifted our position and that we don't plan to reuse code between pipelines. I also think testing two lines of pandas code is a bit redundant.
> 
> I'm happy to push this through as it just fixes an edge case and our tests cover that it doesn't break the main cases.

Fair enough if we're not planning on reusing any of this code.

PR #385, opened 2023-11-30
[FMD-151]: add r3 file for tests, remove redundant example data file

Joel-Pearce:
- Adds an R3 file with dummy data to be used by the integration tests
- Removes the example data model file and replaces it with csvs that can be directly loaded into populate_db()
- Removes redundant tests already captured by integration tests
- Moves tests that hit the /ingest endpoint into integration tests folder
- Removes functionality in which the /ingest endpoint does not validate post-transformed files


tests/controller_tests/test_ingest.py
```diff
@@ -29,71 +27,57 @@
...
-        "detail": "Spreadsheet successfully validated and ingested",
-        "metadata": {},
-        "status": 200,
-        "title": "success",
-        "loaded": True,
-    }
-
+def example_table_data_to_be_loaded_via_populate_db():
+    """
+    Helper function that ingests example data via populate_db() in order to be loaded into the database.
```
cyrusdobbs, line 36: Can we make this a `fixture` with module level scope and then we only need to do it once. The tests can then assert on the db contents?

Although I wonder if the fact that `populate_db` commits should be a consideration

tests/controller_tests/test_ingest.py
```diff
@@ -29,71 +27,57 @@
...
-    assert response.json == {
-        "detail": "Spreadsheet successfully validated but NOT ingested",
-        "metadata": {},
-        "status": 200,
-        "title": "success",
-        "loaded": False,
-    }
+    We cannot use the csvs in the resources folder for test/ because they are in a state in which they can
+    be loaded into the database without invocation of populate_db() and will cause the function to fall over
+    if called as a parameter in it.
```
cyrusdobbs, line 44: We can't use the other CSVs because they represent an untransformed set of data right? But we need transformed data, in the structure of the data model in order to load it straight in via populate.

Not sure if this current explanation can be clearer as I didn't really understand it

cyrusdobbs (commented):
I think we can also move the tests in `tests/controller_tests/test_ingest.py` into integration tests because they rely on having a DB connection

tests/controller_tests/test_ingest.py
cyrusdobbs: Ah yeah I see. Yep a fixture that returns the `data_dict` etc would be ideal. Then we can call `populate_db` in the tests directly, just passing the `mock_r3_data_dict` etc from the fixture

tests/integration_tests/test_ingest_component.py
```diff
@@ -594,3 +588,33 @@ def test_ingest_with_r4_round_agnostic_failures(test_client, towns_fund_round_4_
...
+
+def test_ingest_endpoint_missing_file(test_client):
+    """Tests that, given a sheet name but no file, the endpoint returns a 400 error."""
+    endpoint = "/ingest"
+    response = test_client.post(
+        endpoint,
+        data={},  # empty body
+    )
+
+    decoded_response = json.loads(response.data.decode())
```
cyrusdobbs, line 601: I think we can use `response.json()` instead of the `json` library

tests/integration_tests/test_ingest_component.py
```diff
@@ -594,3 +588,33 @@ def test_ingest_with_r4_round_agnostic_failures(test_client, towns_fund_round_4_
             },
         ],
     }
+
+
+def test_ingest_endpoint_missing_file(test_client):
+    """Tests that, given a sheet name but no file, the endpoint returns a 400 error."""
```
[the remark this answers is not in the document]
Joel-Pearce: Other than asserting on response.json() I'm not really sure what you mean by updating?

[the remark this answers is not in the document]
cyrusdobbs: `given a sheet name but no file`

core/controllers/ingest.py
```diff
@@ -44,12 +44,8 @@ def ingest(body: dict, excel_file: FileStorage) -> Response:
...
     original_workbook = extract_data(excel_file=excel_file)
 
-    if reporting_round:
-        pre_transformation_validations(original_workbook, reporting_round, auth)
-        workbook = transform_data(original_workbook, reporting_round)
-    else:
-        # when no reporting round, source workbook is already in a transformed state
-        workbook = original_workbook  # TODO: Deprecate ability to ingest this workbook with no round.
+    pre_transformation_validations(original_workbook, reporting_round, auth)
+    workbook = transform_data(original_workbook, reporting_round)
```
cyrusdobbs, line 48: 😍

tests/controller_tests/test_ingest.py
cyrusdobbs, line 1: Sounds like the plan is to move all tests that have external dependencies to the separate `integration_tests` folder. This was we can more easily run the fast unit tests in isolation from the integration tests with bigger overheads e.g. DB set up / tear down. Many of these look like they use the db so we could move them as part of this PR?

Otherwise this is really nice. Very glad to see the back of being able to pass no `reporting_round` to ingest - nice work 👍

cyrusdobbs: There are some unit tests nestled in that test module - i.e. not all of them use the DB. Yeah I agree. I think the csvs can go in into their own folders that are named more clearly than just being in various `resources` folders

Joel-Pearce: Separated out the unit tests and have created two folders in ```integration_tests``` to distinguish between the transformed r3 mock data and the mock tf returns.

cyrusdobbs (approved):
Good stuff LGTM

PR #386, opened 2023-11-30
[BAU] strip values before being checked during pre transformation val…

cyrusdobbs:
…idation

### Change description
- strip values from the spreadsheet during pre-validation

### How to test

### DIFF
--- a/core/validation/pre_transformation_validate.py
+++ b/core/validation/pre_transformation_validate.py
@@ -110,7 +110,7 @@ def conflicting_input_validation(workbook: dict[str, pd.DataFrame], schema: dict
 def check_values(
     workbook: dict,
     failure_list: [PreTransformationCheck],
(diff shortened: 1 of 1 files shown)


core/validation/pre_transformation_validate.py
```diff
@@ -123,7 +123,7 @@ def check_values(
     failures = []
 
     for sheet, column, row, expected_values, type in failure_list:
-        expected_value = workbook[sheet].iloc[column][row]
+        expected_value = str(workbook[sheet].iloc[column][row]).strip()
```
Joel-Pearce, line 126: I think this should actually be ```entered_value``` as we use the column and row to get the actual value. Otherwise a bit confusing as we're supposed to do entered vs expected.

Joel-Pearce (changes requested):

PR #387, opened 2023-12-01
[BAU] improve exception handling for when extracting data from an Exc…

cyrusdobbs:
…el file

### Change description
- logs more informative error messages when Exceptions occur on data extract

### How to test

### DIFF
--- a/core/controllers/ingest.py
+++ b/core/controllers/ingest.py
@@ -2,10 +2,11 @@
 import json
 from io import BytesIO
 from json import JSONDecodeError
+from zipfile import BadZipFile
 
 import numpy as np
 import pandas as pd
-from flask import Response, abort, g, jsonify
+from flask import Response, abort, current_app, g, jsonify
 from sqlalchemy import desc, exc, func
 from werkzeug.datastructures import FileStorage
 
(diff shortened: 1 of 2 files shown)


core/controllers/ingest.py
```diff
@@ -104,8 +105,12 @@ def extract_data(excel_file: FileStorage) -> dict[str, pd.DataFrame]:
             sheet_name=None,  # extract from all sheets
             engine="openpyxl",
         )
-    except ValueError:
+    except ValueError as cannot_read:
+        current_app.logger.error(f"Cannot read the excel file. {cannot_read}")
         return abort(500, "Internal Ingestion Error")
+    except BadZipFile as corrupt_error:
+        current_app.logger.error(f"Tried to read excel but the file was corrupted. {corrupt_error}")
+        return abort(400, "excel_file is corrupted")
```
Joel-Pearce, line 113: Do we need a test for this? There's currently one for a .txt file, nothing for something that specifically raises ```BadZipFile```.

Joel-Pearce (changes requested):

core/controllers/ingest.py
cyrusdobbs: We can add a test but all it will do is mock out the response from pd.read_excel and check that the abort and log occurs..  I suppose it's worth it anyway to ensure that `except` block isn't changed / removed

cyrusdobbs: Added tests for both exceptions. Combined them to return the same response and log the same message format as they both mean the file is bad

Joel-Pearce (approved):
LGTM

PR #392, opened 2023-12-04
Feature/smd 245/refactor schema types

bobbi-hbrown:
https://dluhcdigital.atlassian.net/jira/software/c/projects/SMD/boards/140?quickFilter=815&selectedIssue=SMD-245

Documentation as part of the readme will need to be updated with these changes, however I thought it best to leave that in its own documentation ticket.


core/validation/casting.py
```diff
@@ -18,24 +20,23 @@ def cast_to_schema(data: dict[str, pd.DataFrame], schema: dict) -> None:
...
+
+
+_NUMPY_TO_PY_TYPES = {
+    "object": "object",
+    "int64": int,
+    "bool": bool,
+    "float64": float,
+    "datetime64[ns]": datetime,
+    "<M8[ns]": datetime,
+}
```
Joel-Pearce, line 42: What are we now using this for?

bobbi-hbrown: @cyrusdobbs I think it's because we were still having issues with multiple values being cast as objects, and a mixture of datetime/timestamp types when we just removed the mapping completely - does that sound accurate to you?

cyrusdobbs: Looking at the code this isn't actually used anywhere now.. so we can delete it

core/validation/failures/user.py
```diff
@@ -75,7 +76,7 @@ def __str__(self):
 class WrongTypeFailure(SchemaUserValidationFailure):
     """Class representing a wrong type failure that is raised when data is of an incorrect type."""
 
-    expected_type: str
+    expected_type: str | float | int | datetime | object | bool | list
```
Joel-Pearce, line 79: For now this doesn't include list. Only ```Postcodes``` is a list, and it's a column populated by our own extraction function - and converted into a list from a series of extracted strings of postcodes - so we don't want to validate that the type is correct and return a message to the user about this.

core/validation/casting.py
```diff
@@ -18,24 +20,15 @@ def cast_to_schema(data: dict[str, pd.DataFrame], schema: dict) -> None:
...
-                    except ValueError:
-                        continue  # this will be caught during validation
-
-        if table_retyped:
-            data[table] = table_data
+
+        for pos, (index, row) in enumerate(table_data.iterrows()):
+            for column, value in row.items():
+                # Postcodes is now a list
+                if pd.isna(value) or isinstance(value, (datetime, pd.Timestamp, list)):
```
Joel-Pearce, line 27: Do we not want to validate that a cell is a datetime, or am I missing something?

Joel-Pearce (changes requested):
Left a couple of small comments, let me know if you have any questions.

Joel-Pearce:
I just ran these changes with the R1 file and got:

```
 if pd.isna(value) or isinstance(value, (datetime, pd.Timestamp, list)):
ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```

The problem is here:

```if pd.isna(value) or isinstance(value, (datetime, pd.Timestamp, list)):```

We can't call ```pd.isna()``` on a list with multiple elements. The easiest solution is basically what we have now - to switch the two conditions in the if statement around. That will mean if the value is a list it will never hit the 2nd condition.

I wrote a test to account for this scenario, so if you rebase it should help catch this.

**Comment by bobbi-hbrown** [comment]
pre-commit.ci autofix

File: core/validation/validate.py [hunk 4]
```diff
@@ -152,8 +152,12 @@ def validate_types(data_dict: dict[str, pd.DataFrame], table: str, column_to_typ
...
 
             # do not raise an exception for pandas Timestamp, datetime or number values
-            if isinstance(got_value, numbers.Number) and exp_type in [int, float] or \
-               isinstance(got_value, (datetime, pd.Timestamp)) and exp_type == datetime:
+            if (
+                isinstance(got_value, numbers.Number)
+                and exp_type in [int, float]
+                or isinstance(got_value, (datetime, pd.Timestamp))
+                and exp_type == datetime
+            ):
```
Joel-Pearce, line 160: I think this is clearer as:

```
if (
    (isinstance(got_value, numbers.Number) and exp_type in [int, float])
    or (isinstance(got_value, (datetime, pd.Timestamp)) and exp_type == datetime)
):
```

Joel-Pearce (changes requested):
One comment left, then happy to approve

bobbi-hbrown:
pre-commit.ci autofix

Joel-Pearce (approved):
LGTM

PR #393, opened 2023-12-04
[SMD 439] Split out validation message logic from failure classes 

Connor-OS:
Removes all the towns fund specific validation message logic from the failure classes and associates it all with a TFMessaging class that can handle the construction of TF specific messages from failure objects.

Currently messages are generated in four ways and handled slightly differently:

- Pre_transformation_validation -> process_validation_failures, Messaging object
- mid_transformatio_validation -> raise validation_error, validation_error_handler
- Schema_validation -> process_validation_failures, Messaging object
- round_specific_validation -> bespoke module, process_validation_failures, Messaging object

I'm not entirely happy with the way the validation_error_handler() is implemented as I feel it duplicates a lot of code from process_validation_failures()

### Change description

### How to test


core/messaging/TF_message_bank.py
cyrusdobbs, line 1: Some of these are very generic and I'd expect them to be shared between different templates/funds

core/messaging/message.py
```diff
@@ -0,0 +1,356 @@
...
+        for error in errors
+        if not any(
+            ((sheet, normalise_risk_section_name(section), cell_index.strip()) in unique_identifiers_from_null_failures)
+            for sheet, section, cell_index, _, _ in [error]
+            for cell_index in error[2].split(",")
+        )
+    ]
+
+    filtered_errors.extend(null_failures)
+    return filtered_errors
```
cyrusdobbs, line 61: Not a comment for this PR necessarily but this function is a monster.. we should look to improve the way we do this

core/messaging/message.py
```diff
@@ -0,0 +1,356 @@
...
+            case "InvalidEnumValueFailure":
+                return self.invalid_enum_value_failure_message(validation_failure)
+            case "NonNullableConstraintFailure":
+                return self.non_nullable_constraint_failure_message(validation_failure)
+            case "TownsFundRoundFourValidationFailure":
+                return self.towns_fund_round_four_validation_failure(validation_failure)
+            case "WrongInputFailure":
+                return self.wrong_input_failure_message(validation_failure)
+            case "UnauthorisedSubmissionFailure":
+                return self.unauthorised_submission_failure(validation_failure)
```
cyrusdobbs, line 153: This looks good. Can we compare classes rather than stings here?

core/controllers/ingest.py
```diff
@@ -51,7 +56,8 @@ def ingest(body: dict, excel_file: FileStorage) -> Response:
         # when no reporting round, source workbook is already in a transformed state
         workbook = original_workbook  # TODO: Deprecate ability to ingest this workbook with no round.
 
-    validate(workbook, original_workbook, reporting_round)
+    if validation_failures := validate(workbook, original_workbook, reporting_round):
+        return process_validation_failures(validation_failures)
```
cyrusdobbs, line 60: I prefer this to raising an Exception - at least in this instance. But validation failures can be instantiated during PTV and transformation so we need to think about how we handle that as well

core/controllers/ingest.py
```diff
@@ -69,6 +75,40 @@ def ingest(body: dict, excel_file: FileStorage) -> Response:
...
+        return {
+            "detail": "Internal ingest exception.",
+            "id": failure_uuid,
+            "status": 500,
+            "title": "Internal Server Error",
+            "internal_errors": internal_failures,
+        }, 500
+
+    # create messaging object here and handle messages
+    messager = _messaging_class_factory("Towns Fund")  # TODO: Replace with fund type from ingest params
```
cyrusdobbs, line 96: I like the use of the factory pattern here. 

Although, this shouldn't be private if we're importing it from the `message` module

core/validation/failures/user.py
cyrusdobbs, line 1: Nice 👍

cyrusdobbs (commented):
Good stuff - going in the right direction here

core/messaging/TF_message_bank.py
Connor-OS: I'm trying to think best how to store the messages, to allow them to be shared easily by all messaging objects, but also allow them to be updated and added to with fund specific messaging.

currently we have this module message_bank. I'm now leaning towards making shared messages attributes of the MessagingBase class so they are inherited, then each messaging class can add to/overload these, Do you think this is a sensible/scalable approach?

core/controllers/ingest.py
Connor-OS: I'm finding it very complex to unite these three paths into seperate sources of validation failures that can all benifit from a Messeger class.
- schema validation errors, and pre-transformation errors are fine
- TF round 4 and errors raised during transformation are much more complicated because they are constructed so differently

core/messaging/TFMessaging.py
```diff
@@ -0,0 +1,234 @@
...
+            case WrongInputFailure():
+                return self.wrong_input_failure_message(validation_failure)
+            case UnauthorisedSubmissionFailure():
+                return self.unauthorised_submission_failure(validation_failure)
+            case GenericFailure():
+                return self.generic_failure(validation_failure)
+            case _:
+                raise UnimplementedErrorMessageException
+
+    def non_unique_composite_key_failure_message(self, validation_failure) -> tuple[str, str, str, str]:
```
cyrusdobbs, line 57: We spoke about making these private. Would be good to also update the type hints.

core/messaging/TFMessaging.py
```diff
@@ -0,0 +1,234 @@
...
+
+    def wrong_input_failure_message(self, validation_failure) -> tuple[str | None, str | None, str]:
+        return None, None, self.tf_msgs.PRE_TRANSFORMATION_MESSAGES[validation_failure.value_descriptor]
+
+    def unauthorised_submission_failure(self, validation_failure) -> tuple[str | None, str | None, str]:
+        places_or_funds = join_as_string(validation_failure.expected_values)
+        message = self.msgs.UNAUTHORISED.format(
+            wrong_place_or_fund_type=validation_failure.entered_value, allowed_places_or_fund_types=places_or_funds
+        )
+        return None, None, message
```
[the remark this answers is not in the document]
Connor-OS: No, because these are pre-transformation messages they have a different function signature. This migth lead to confusion as there is no clear way to tell what is a pre and post validation message at this level

core/messaging/message.py
Connor-OS: This is the final outstanding issue with this PR. I'd like not to have to refactor this twice so I need to do it right.

One option, to remove this function entirely is: we do not pick up composite key errors if the key contains a null with a line like:
`duplicated_rows = duplicated_rows.dropna()`

This mostly does the trick, however we run into the issue that Risks section composite key contains at least one null, proggramme/project ID. meaning we won't pick up any composite key errors for Risks

core/const.py
cyrusdobbs, line 1: It won't let me comment on the line but this is also unused, so lets delete: `FUND_TYPE_TO_TD_OR_HS`

Constants `FINANCIAL_YEAR_TO_ORIGINAL_COLUMN_LETTER_FOR_NON_FOOTFALL_OUTCOMES` and `MONTH_TO_ORIGINAL_COLUMN_LETTER_FOR_FOOTFALL_OUTCOMES` are used by `get_cell_indexes_for_outcomes` which current sits in `validation/utils` but wants to be moved to the new TFMessaging class as well I think

core/validation/failures/user.py
```diff
@@ -266,140 +114,16 @@ class UnauthorisedSubmissionFailure(PreTransFormationFailure):
...
-    cell index, and keeps only those present in null_failures. Additionally, includes all null_failures and errors
-    not present in null_failures or any part of the cell index is not already captured by a null failure.
-
-    :param errors: List of error tuples (sheet, section, cell_index, message, error_type).
-    :return: Filtered list of errors, including all null_failures and errors not present in null_failures or any part
-    of the cell index is not already captured by a null failure.
-    """
+@dataclass
+class TownsFundRoundFourValidationFailure(UserValidationFailure):
+    """Generic Towns Fund Round 4 Validation Failure."""
```
cyrusdobbs, line 120: I don't think this should be TF specific - I expect new funds will also want to be able to instantiate a generic failure with the same signature - so lets just make this `GenericUserFailure` or similar I think?

scripts/validation-spike.py
```diff
@@ -447,7 +447,7 @@ def get_error_type(error):
             ),
             "Before 2020/21": (pa.Column(float), "F"),
             "Financial Year 2020/21 (£s), H1 (Apr-Sep), Actual": (
-                pa.Column(float, pa.Check.greater_than(0)),
```
cyrusdobbs, line 450: Do you want to commit this?

tests/validation_tests/test_failures.py
cyrusdobbs, line 1: All tests that import from `tf_messaging` should be moved to a new module `test_tf_messaging` under a new `messager_tests` test package.

We also need to move all tests that import from `messaging`

`failures_to_messages`
 `group_validation_messages`
 `remove_errors_already_caught_by_null_failure`

to a generic `test_messaging` test class. 

In order to test `failures_to_messages` it would be best to create a generic `TestMessager` class that implements `to_messages` in a simple way. Feel free to reach out and we can pair on it if you're not sure about this

core/controllers/ingest.py
```diff
@@ -78,6 +86,36 @@ def ingest(body: dict, excel_file: FileStorage) -> Response:
     return jsonify(success_payload)
 
 
+def process_validation_failures(validation_failures):
+    internal_failures = [failure for failure in validation_failures if isinstance(failure, InternalValidationFailure)]
```
cyrusdobbs, line 90: docs and type hints for this please

also, we'll need some unit tests for this function as well I think. We may be able to use some of the existing ones from when this was used as a handler function?

Best to just mock out the return value of `faulures_to_messages` etc

core/controllers/ingest.py
```diff
@@ -57,10 +62,13 @@ def ingest(body: dict, excel_file: FileStorage) -> Response:
...
-    pre_transformation_validations(original_workbook, reporting_round, auth)
-    workbook = transform_data(original_workbook, reporting_round)
+    try:
+        pre_transformation_validations(original_workbook, reporting_round, auth)
+        workbook = transform_data(original_workbook, reporting_round)
+        validate(workbook, original_workbook, reporting_round)
 
-    validate(workbook, original_workbook, reporting_round)
+    except ValidationError as validation_error:
+        return process_validation_failures(validation_error.validation_failures)
```
cyrusdobbs, line 71: This looks good

core/messaging/messaging.py
```diff
@@ -0,0 +1,121 @@
+from core.messaging import Message
+from core.messaging.tf_messaging import SharedMessages as msgs
+from core.messaging.tf_messaging import TFMessaging
+from core.validation.failures.user import PreTransFormationFailure
+
+
+def group_validation_messages(validation_messages) -> list[Message]:
```
cyrusdobbs, line 7: typing hint on val messages

core/messaging/messaging.py
```diff
@@ -0,0 +1,121 @@
...
+            for cell_index in error[2].split(",")
+        )
+    ]
+
+    filtered_errors.extend(null_failures)
+    return filtered_errors
+
+
+def failures_to_messages(
+    validation_failures, messager
```
cyrusdobbs, line 82: type hints

core/messaging/messaging.py
```diff
@@ -0,0 +1,121 @@
...
+
+    error_messages.sort()
+
+    # group cells by sheet, section and desc
+    error_messages = group_validation_messages(error_messages)
+
+    return {"validation_errors": [vars(message) for message in error_messages]}
+
+
+def messaging_class_factory(fund: str):
```
cyrusdobbs, line 111: return type hint

cyrusdobbs (commented):
This is a big one.. Some really important and effective changes here though I think!

minor comments, some regarding unit testing

core/validation/failures/user.py
Connor-OS: Agreed, we have two types of "Generic" failures with different signatures one with all the message return values instantiated and this new one with column and row_index. I dont' think there is some neet way to overload or combine them though so I think both implementations must stay and we use different names for each

cyrusdobbs (approved):
Nice work on this - LGTM

PR #394, opened 2023-12-05
FMD-9 Store postcodes as an array (not as CSV)

mattpep:
### Change description

### How to test
NB Note that this PR contains schema changes (in the form of migrations) so would require extra care and testing in those areas.


core/validation/casting.py
```diff
@@ -25,7 +25,7 @@ def cast_to_schema(data: dict[str, pd.DataFrame], schema: dict) -> None:
             if column in table_types:
                 original_type = table_types[column]
 
-                if original_type != target_type:
+                if (original_type != target_type) and (column != "Postcodes"):
```
Joel-Pearce, line 28: Ideally we don't want ```casting.py``` to be specifying actual column names.

Joel-Pearce (changes requested):
We want to remove any specific mention of "Postcodes" from code not in ```core/extraction/towns_fund_round_three.py```. Ideally, we want to validate the "Postcodes" column as a list/array, and junk all mention of ```object```. This is difficult as-is because I don't think ```astype()``` accepts list as a parameter. We could probably throw a conditional around that line of code such that lists/arrays are treated differently. 

Bobbi is working on a branch to refactor out all mention of numpy types, and that might simplify this. May be worth catching up with her, or rebasing onto her branch depending on what state it's in. 

Additionally, we want to validate "Postcodes" not only as a column containing arrays, but as a column containing arrays of strings. That's where things get a bit more difficult. Happy to catch up with you on this if any of it is unclear.

core/validation/validate.py
```diff
@@ -141,6 +141,10 @@ def validate_types(data_dict: dict[str, pd.DataFrame], table: str, column_to_typ
     for index, row in data_dict[table].iterrows():
         for column, exp_type in column_to_type.items():
             got_value = row.get(column)
+
+            if column == "Postcodes":
+                continue
+
```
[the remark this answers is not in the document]
Joel-Pearce: We also very much want to validate that this column does have the correct data types. Turning off validation to see if everything was working as intended as a test is fine, but we need to turn the validation for the column back on. Happy to have a mull over how to do this with you.

core/util.py
```diff
@@ -127,6 +125,10 @@ def load_example_data():
         "risk_register",
     ]:
         table_df = pd.read_csv(resources / f"{table}.csv")
+        if table == "project_dim":
```
nsnape, line 128: can this be done much more efficiently with Pandas vectorised .str.split method? ie something like

`table_df["postcodes"] = table_df["postcodes"].str.split(",")`

core/validation/casting.py
Joel-Pearce:
```
if target_type == "list":
       # do bespoke convert values in col to list surrounded by try/catch
```
is what we want

db/migrations/versions/009_postcode_as_array.py
```diff
@@ -0,0 +1,35 @@
...
+revision = "009_postcode_as_array"
+down_revision = "008_add_new_round_4_cols"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    with op.batch_alter_table("project_dim", schema=None) as batch_op:
+        batch_op.add_column(sa.Column("postcode_list", sa.ARRAY(sa.VARCHAR()), nullable=True))
+    op.execute("UPDATE project_dim SET postcode_list = STRING_TO_ARRAY(postcodes, ',')")
```
nsnape, line 21: I'm wondering, and this might be crazy, but could we simplify the whole thing by keeping ingest / validation as it is, and just executing STRING_TO_ARRAY on the value at ingest time for this particular column?

core/db/entities.py
```diff
@@ -242,7 +242,7 @@ class Project(BaseModel):
     primary_intervention_theme = sqla.Column(sqla.String(), nullable=False)
     location_multiplicity = sqla.Column(sqla.String, nullable=True)
     locations = sqla.Column(sqla.String, nullable=False)
-    postcodes = sqla.Column(sqla.String, nullable=True)
+    postcodes = sqla.Column(sqla.ARRAY(sqla.String), nullable=True)
```
nsnape, line 245: so I pulled the branch and checked the extract behaviour in tests, and our potcode data now in quare brackets (python list) in serialisation tests. like so: 

`['BS3 1AB']`

I'm wondering if:
1) the fact this didn't cause any test failures means we probably don't have specific coverage of this particular field being serialised
2) How does Excel output handle this?

nsnape: Yeah I've tested locally and can confirm that now "postcodes" column in Excel file looks like:

`['SW1A 2AA', 'BT1 1AA']`

When really it should look like it did before, ie just comma separated list, this is from a previous Excel file:

`SW1A 2AA,BT1 1AA`

nsnape: We should probably:
1) Try and fix the extract / serialisation to match the old format on extract (I presume some of Hrishi's E2E tests will flag this failure)
2) add a test assertion on the format of this particular field, probably in serialisation_tests

nsnape (changes requested):

core/util.py
```diff
@@ -30,9 +30,9 @@ def postcode_to_itl1(postcode: str) -> str | None:
     return POSTCODE_AREA_TO_ITL1.get(postcode_area.upper())
 
 
-def get_itl_regions_from_postcodes(postcodes: str) -> set[str]:
+def get_itl_regions_from_postcodes(postcodes: list[str]) -> set[str]:
     """
-    Transform a comma separated string of postcodes into set of corresponding ITL regions.
+    Transform an array of postcodes into set of corresponding ITL regions.
 
     :param postcodes: A string representing a comma separated sequence of postcodes.
```
nsnape, line 37: this docstring line for postcodes param needs updating

core/validation/validate.py
```diff
@@ -141,7 +142,10 @@ def validate_types(data_dict: dict[str, pd.DataFrame], table: str, column_to_typ
     for index, row in data_dict[table].iterrows():
         for column, exp_type in column_to_type.items():
             got_value = row.get(column)
-            if got_value is None or pd.isna(got_value):
+
+            if (isinstance(got_value, list) and got_value == []) or (
+                not isinstance(got_value, list) and got_value in NA_VALUES
+            ):
```
Joel-Pearce, line 148: The Postcodes column could potentially be something in ```NA_VALUES``` - if it's empty it won't necessarily be instantiated as an empty list, it could be treated as a null value. Better to do simply do ```if got_value in NA_VALUES or not got_value```.

core/validation/casting.py
```diff
@@ -26,6 +26,11 @@ def cast_to_schema(data: dict[str, pd.DataFrame], schema: dict) -> None:
                 original_type = table_types[column]
 
                 if original_type != target_type:
+                    if target_type == "list":
+                        # TODO: Ideally we'd like to turn this into a real list, but that type is not supported
+                        # in pandas. We'll remove this when we refactor out the numpy types.
```
Joel-Pearce, line 31: Better change the comment to something like: ```Forcible conversion of values to target_type == "list" occurs in the relevant extraction function, and .astype does not support lists.```

Joel-Pearce (changes requested):

core/extraction/towns_fund_round_three.py
```diff
@@ -288,9 +288,7 @@ def extract_project(df_project: pd.DataFrame, project_lookup: dict, programme_id
         lambda row: row[single_postcode] if row["Single or Multiple Locations"] == "Single" else row[multiple_postcode],
         axis=1,
     )
-    df_project["Postcodes"] = df_project.apply(
-        lambda row: ",".join(extract_postcodes(row["Locations"])), axis=1
-    )  # TODO: keep as a list and store as array of strings in db (not supported by SQLite)
+    df_project["Postcodes"] = df_project.apply(lambda row: ",".join(extract_postcodes(row["Locations"])), axis=1)
```
nsnape, line 291: I think this line, and the corresponding sub-function (split_string_by_delimiter) can all be replaced with something much more simple like 

`df_project["Postcodes"] = [extract_postcodes(x) for x in df_project["Locations"]]`

And then update extract_postcodes function to return None (or np.nan) type if no postcodes found. This then removes this double-handling of carting array to a string, then invoking another function per-line to iteratively convert back to array. It just does it in one go, without using apply twice.

Obvs, this will require some changes to everything else that extract_potcodes uses (I think some older extraction rounds also use it), but I think they are sensible, as otherwise we will have different extraction behaviour b/w rounds.

core/serialisation/data_serialiser.py
```diff
@@ -44,6 +44,24 @@
 )
 
 
+class PostcodeList(fields.Field):
```
nsnape, line 47: I like this solution, custom field class. Good to know we can do this

core/serialisation/data_serialiser.py
```diff
@@ -44,6 +44,24 @@
...
+    """Used to convert between a postcode array (as stored in the DB) and a CSV string (as appears in json/excel)"""
+
+    def _serialize(self, value, attr, obj, **kwargs):
+        if value is None:
+            return ""
+        return ", ".join(value)
+
+    def _deserialize(self, value, attr, data, **kwargs):
```
nsnape, line 55: is _deserialize necessary? I can't see that it's actually called anywhere when I run debugger on test_suite? Correct me if I'm wrong but this converts JSON to sqlalchemy, which we don't do?

db/migrations/versions/009_postcode_as_array.py
nsnape: yeah, not needed now you've extended the serialiser. I'll close this

core/serialisation/data_serialiser.py
```diff
@@ -44,6 +44,24 @@
 )
 
 
+class PostcodeList(fields.Field):
+    """Used to convert between a postcode array (as stored in the DB) and a CSV string (as appears in json/excel)"""
+
+    def _serialize(self, value, attr, obj, **kwargs):
```
nsnape, line 50: I think I mentioned this before, but it may have got lost in the comments or something:

Please could you add a test to test_serialisation, that specifically looks for a multi-postcode row in ProjectDetails, and asserts it's format. 
running debugger on `tests.serialisation_tests.test_serialisation.test_serialise_download_data_no_filters` should give you a clue how to implement (row 05 of that dict has 2 postcodes)

nsnape (changes requested):

Joel-Pearce:
I'd forgot entirely about historical data, but ultimately the historical data needs to be standardised alongside R3 & R4 (i.e. we can't have R1 and R2 continue to treat Postcodes as a string field if it's now an array field). 

The best solution IMO is to pull out the logic in ```towns_fund_round_three.py``` that converts the Postcodes column into arrays, and put it in a method in utils called ```convert_col_values_to_list()```. This can then be called either in the appropriate place of every extraction method, or, perhaps better, can be called in ```casting.py``` if ```target_type == "list"```.

core/validation/casting.py
```diff
@@ -26,6 +26,11 @@ def cast_to_schema(data: dict[str, pd.DataFrame], schema: dict) -> None:
                 original_type = table_types[column]
 
                 if original_type != target_type:
+                    if target_type == "list":
+                        # Forcible conversion of values to target_type == # "list" occurs in the relevant
+                        # extraction function, and .astype does not support lists.
+                        continue
```
Joel-Pearce, line 32: This can be now updated to:

```# Forcible conversion of values to target_type == "list" occurs in extract_postcodes() and .astype does not support lists```

**Review by Joel-Pearce — CHANGES_REQUESTED** [review]
Two remaining things:

- make sure we've tested behaviour where postcode field is empty. If there's an issue arising from ```clean_data``` we will have to amend it accordingly. 
- change the schema type of the Postcodes column in the R1 & R2 schemas

Then happy to approve :+1:

**Comment by mattpep** [comment]
> * make sure we've tested behaviour where postcode field is empty. If there's an issue arising from `clean_data` we will have to amend it accordingly.
The test data I'm using for manual imports contains several postcode-less records which are both ingested and downloaded correctly. 

> * change the schema type of the Postcodes column in the R1 & R2 schemas
I've added these changes too.

**Review by Joel-Pearce — APPROVED** [review]
LGTM. The arduous journey from string to array is almost complete!

File: core/extraction/utils.py [hunk 14]
```diff
@@ -89,16 +89,19 @@ def datetime_excel_to_pandas(excel_dates_column: pd.Series) -> pd.Series:
     return python_dates_column
 
 
-def extract_postcodes(s: str | float) -> list[str]:
+def extract_postcodes(s: str | float) -> list[str] | None:
```
nsnape, line 92: sorry missed this one before, does this function actually take a float? That seems like a weird thing to have in there, maybe it's a hang-over from some old implementation? Could you check, and if this is nto the case, remove the union with float?

If it is the case then maybe we should re-think the function design, but I'm pretty certain it's a mistake

core/util.py
nsnape: This not resolved, please switch to above format. It should work, I previously tried it locally and all tests still passed

nsnape: so replace 
```
for x in range(len(table_df.postcodes.values)):
                table_df.postcodes.values[x] = table_df.postcodes.values[x].split(",")
```
with
`table_df["postcodes"] = table_df["postcodes"].str.split(",")`

nsnape: this is much faster as it uses the dataframe.str vector method which applies string transformations as a vector to all rows, rather than the much less efficient iteration though all DF rows

core/validation/schema.py
```diff
@@ -75,6 +75,7 @@ class SchemaError(ValueError):
 
 _PY_TO_NUMPY_TYPES = {
     "str": "string",
+    "list": "list",  # This whole dict will be removed when we refactor out the numpy types.
```
nsnape, line 78: is there a ticket for this? Also, can the comment be appended with "TODO:" and moved up to the actual dict definition line so it's harder to miss / ignore

tests/conftest.py
```diff
@@ -271,7 +271,7 @@ def additional_test_data() -> dict[str, Any]:
         project_name="TEST-PROJECT-NAME4",
         primary_intervention_theme="TEST-PIT4",
         locations="TEST-LOCATIONS4",
-        postcodes="",  # no postcode == no region
+        postcodes=[],  # no postcode == no region
```
nsnape, line 274: interestingly, if you change this to None, the data comes out the same from the serialisation (postcodes field for this row has "" value), but the records come out in a different order (so the test fails). I'm not suggesting we change anything, it's just weird, and makes me uneasy about asserting things in tests based on record order, when in reality that record order is arbitrary

nsnape (changes requested):

core/validation_schema.py
```diff
@@ -415,7 +415,7 @@
             "Primary Intervention Theme": "str",
             "Single or Multiple Locations": "str",
             "Locations": "str",
-            "Postcodes": "str",
+            "Postcodes": "list",
```
cyrusdobbs, line 418: Conscious that there are no changes to the validation unit tests... do we want to add some unit tests to cover the new scenarios and code paths introduced with this new type? 

I would usually say that this should be hard requirement for this PR, but given that we may be moving on from this validation code, perhaps it's not as critical

Joel-Pearce: I've just realised that validating the type for Postcodes doesn't actually make any sense. If we find that the type is not a list, and that instead it's a string, are we going to return a message to the user saying, "You have entered text instead of a list?". It doesn't make any sense as the application logic is what converts the inputted postcodes from the spreadsheet into a list.

Joel-Pearce (changes requested):
I've had an eleventh hour realisation that we cannot currently validate the Postcode column's type as a list because of the following:

> I've just realised that validating the type for Postcodes doesn't actually make any sense. If we find that the type is not a list, and that instead it's a string, are we going to return a message to the user saying, "You have entered text instead of a list?". It doesn't make any sense as the application logic is what converts the inputted postcodes from the spreadsheet into a list.

Let me have a think and come back to this at SU tomorrow.

core/validation_schema.py
Joel-Pearce: I don't think it's even possible for the user to enter a 'wrong type' into the Postcode field as it's generated from the ```extract_postcodes``` method, but if for whatever reason our application side logic fails then the user will be greeted with a rather confusing message.

Joel-Pearce: I think we should keep it as a string for validation purposes and convert to an array afterwards. Ultimately, the user is putting 'text' into the field from which the value is generated. It also shares the same cell location as 'Locations', which would make it doubly confusing if there's two types expected in one cell. What if they're both somehow wrong? The user has to put 'text' and a 'list' into the same cell, but only one?

core/util.py
nsnape: Hmm, just reading more about this, I think in this case iteration might be faster. The general consensus seems to be that whilst vectorised methods should usually be much faster, the string ones are not necessarily: https://stackoverflow.com/questions/54028199/are-for-loops-in-pandas-really-bad-when-should-i-care

nsnape (approved):

Joel-Pearce (approved):
On further reflection I'm approving this PR. 

1) We can only trigger a type error if the application side logic fails. This logic is confined to ```extract_postcodes```, which is a small, robust, and well-tested function.
2) It's not possible to trigger a type error on the ```Postcodes``` column in an integration test without mocking side-effects, which isn't useful.
3) We're changing the way in which validation works going forward and any refactor of how the ```Postcodes``` column is converted and validated will soon be obsolescent.

cyrusdobbs:
Do any existing integration tests contain postcodes at least @Joel-Pearce? Conscious there is not any coverage of the "list" logic throughout validation otherwise

Joel-Pearce:
> Do any existing integration tests contain postcodes at least @Joel-Pearce? Conscious there is not any coverage of the "list" logic throughout validation otherwise

There's no inputted value that will cause a ```WrongTypeFailure``` without mocking side-effects. However, we have a test called ```test_ingest_with_r4_file_project_admin_project_progress_failure```, which tests that ```validate_postcodes``` returns an appropriate error, and that test returns the description: "You entered an invalid postcode. Enter a full UK postcode, for example SW1A 2AA."

cyrusdobbs:
> > Do any existing integration tests contain postcodes at least @Joel-Pearce? Conscious there is not any coverage of the "list" logic throughout validation otherwise
> 
> There's no inputted value that will cause a `WrongTypeFailure` without mocking side-effects. However, we have a test called `test_ingest_with_r4_file_project_admin_project_progress_failure`, which tests that `validate_postcodes` returns an appropriate error, and that test returns the description: "You entered an invalid postcode. Enter a full UK postcode, for example SW1A 2AA."

I don't necessarily mean to assert on an error - but just that a template contains some postcodes so that the new code is run. That way we catch if the happy path ever starts raising uncaught exceptions

Joel-Pearce:
> > > Do any existing integration tests contain postcodes at least @Joel-Pearce? Conscious there is not any coverage of the "list" logic throughout validation otherwise
> > 
> > 
> > There's no inputted value that will cause a `WrongTypeFailure` without mocking side-effects. However, we have a test called `test_ingest_with_r4_file_project_admin_project_progress_failure`, which tests that `validate_postcodes` returns an appropriate error, and that test returns the description: "You entered an invalid postcode. Enter a full UK postcode, for example SW1A 2AA."
> 
> I don't necessarily mean to assert on an error - but just that a template contains some postcodes so that the new code is run. That way we catch if the happy path ever starts raising uncaught exceptions

We have postcodes in both the R3 & R4 success files.

PR #401, opened 2023-12-14
[FMD-113]: refactor ingest logic

Joel-Pearce:
PR for refactoring ```populate_db()``` such that each table is loaded by a bespoke function rather than as part of a larger, more general function that distinguishes what table is being loaded with if statements.

Methods have also been pulled out of ```ingest.py``` and into ```load_functions.py``` where deemed appropriate. 

### DIFF
--- a/core/controllers/ingest.py
+++ b/core/controllers/ingest.py
@@ -7,13 +7,23 @@ from zipfile import BadZipFile
 import numpy as np
 import pandas as pd
 from flask import Response, abort, current_app, g, jsonify
-from sqlalchemy import desc, exc, func
+from sqlalchemy import exc
 from werkzeug.datastructures import FileStorage
 
-from core.const import EXCEL_MIMETYPE, EXCLUDED_TABLES_BY_ROUND, SUBMISSION_ID_FORMAT
+from core.const import EXCEL_MIMETYPE, EXCLUDED_TABLES_BY_ROUND
+from core.controllers.load_functions import (
+    delete_existing_submission,
+    get_or_generate_submission_id,
+    get_outcomes_outputs_to_insert,
+    get_table_to_load_function_mapping,
+)
+from core.controllers.load_functions_historical import (
+    get_programmes_newer_round,
+    get_programmes_same_round_or_older,
+)
 from core.controllers.mappings import INGEST_MAPPINGS, DataMapping
 from core.db import db
-from core.db.entities import Organisation, Programme, Project, Submission
+from core.db.entities import Organisation, Submission
 from core.db.utils import transaction_retry_wrapper
 from core.extraction import transform_data
 from core.validation import validate
@@ -78,10 +88,6 @@ def load_data(workbook: dict[str, pd.DataFrame], excel_file: FileStorage, report
         populate_db_historical_data(workbook, INGEST_MAPPINGS)
     else:
         populate_db(workbook=workbook, mappings=INGEST_MAPPINGS)
-    # provisionally removing unreferenced entities caused by updates to ingest process
-    # TODO: DELETE THIS WHEN R1 AND R3 RE-INGESTED, OR NO MORE DUPLICATE ORGS
-    if reporting_round in [1, 3]:
-        remove_unreferenced_organisations()
     submission_id = workbook["Submission_Ref"]["Submission ID"].iloc[0]
     save_submission_file(excel_file, submission_id)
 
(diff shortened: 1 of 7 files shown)


core/controllers/ingest.py
```diff
@@ -162,72 +146,14 @@ def populate_db(workbook: dict[str, pd.DataFrame], mappings: tuple[DataMapping])
...
-                db.session.merge(org_to_merge)  # There can only be 1 org per ingest.
-                continue
-
-        # for outcome and output dim (ref) data, if record already exists, do nothing.
-        if mapping.table in ["Outputs_Ref", "Outcome_Ref"]:
-            models = get_outcomes_outputs_to_insert(mapping, models)
-
-        db.session.add_all(models)
+        load_function = TABLE_TO_LOAD_FUNCTION[mapping.table]
+        if load_function.__name__ == "generic_load":
```
Connor-OS, line 153: why do you need to use.__name__? can't you just compare functions
```if load_function == generic_load:```

Comment (Connor-OS) [reply core/controllers/ingest.py:153]: since generic_load is basicly the default, you could consider only storing the entries for the other cases:
```
TABLE_TO_LOAD_FUNCTION = {   
    "Organisation_Ref": load_organisation_ref,
    "Programme_Ref": load_programme_ref,
    "Programme Progress": generic_load,
    "Funding": generic_load,
    "Outputs_Ref": load_outputs_outcomes_ref,
    "Outcome_Ref": load_outputs_outcomes_ref,
 }
 ```

 then do something like:
 
 ```
 if load_function := TABLE_TO_LOAD_FUNCTION.get(mapping.table):
     load_function() 
 else:
     generic_load()
```    
 But if you don't think this is an improvement then no worries

**Review by Connor-OS — COMMENTED** [review]
much clearer to read, only minor stylistic comments

File: core/controllers/ingest.py [hunk 2]
```diff
@@ -162,72 +146,14 @@ def populate_db(workbook: dict[str, pd.DataFrame], mappings: tuple[DataMapping])
...
-                org_to_merge.id = organisation_exists.id
-                db.session.merge(org_to_merge)  # There can only be 1 org per ingest.
-                continue
-
-        # for outcome and output dim (ref) data, if record already exists, do nothing.
-        if mapping.table in ["Outputs_Ref", "Outcome_Ref"]:
-            models = get_outcomes_outputs_to_insert(mapping, models)
-
-        db.session.add_all(models)
+        load_function = TABLE_TO_LOAD_FUNCTION[mapping.table]
```
nsnape, line 152: how would you envisage this being used for, say, a different fund with a different set of tables to load? Would you have to build a new mapping for each and a new TABLE_TO_LOAD_FUNCTION? Or just one of them?


For instance, a submission with no projects, and only half of the "event" data tables?

Ideally it'd be whatever option makes it the most easily configurable without the need for loads of new code definition, so like a subset of TABLE_TO_LOAD_FUNCTION for each fund type. But then does this seem backwards, as it wwill stil literate through mappings, many of which won't be applicable?

core/controllers/load_functions.py
```diff
@@ -0,0 +1,257 @@
...
+        .first()
+    )
+
+    if programme_exists_same_round:
+        matching_project = None
+        for project in programme_exists_same_round.projects:
+            if project.submission.reporting_round == reporting_round:
+                matching_project = project
+                break
+        if matching_project:
```
nsnape, line 133: this logic definitely needs splitting out further. Deleting a submission shouldn't come under the scope of "get_submission_id"

nsnape: So somewhere like, in higher level function (populate_db) check if submission id exists in a function and return that id or None if not.
Then, also from high level function, if exists already (id not None) call another function to do the delete
Then, call next_submission_id if None
Then continue with insert as normal.

Does that make sense? HAving the workflow slpit up more allows all the components to be more easily unit tested / mocked, and also makes it clearer what the scope of each bit does based on the function names etc

core/controllers/load_functions.py
```diff
@@ -0,0 +1,257 @@
...
+    If the programme exists in a previous round, the incoming programme's information supersedes the existing programme.
+    Data is added to the session via this method, but not committed.
+
+    :param workbook: a dictionary of DataFrames of table data to be inserted into the db.
+    :param mapping: the mapping of the relevant DataFrame to its attributes as they appear in the db.
+    """
+    reporting_round = int(workbook["Submission_Ref"]["Reporting Round"].iloc[0])
+    programme_id = workbook["Programme_Ref"]["Programme ID"].iloc[0]
+    # if added before or during this round, get the programme entity to merge (update)
+    programme_exists_previous_round = (
```
nsnape, line 22: maybe move this db query stuff into a separate function, then it's easier to test/mock, and scope definition / function roles are clearer?

nsnape: It could even go in core/db/queries, just as a simple function that only has the query in it. That way, all DB query code is in the same place, and is also easier to re-use etc.

core/controllers/load_functions.py
```diff
@@ -0,0 +1,257 @@
...
+    The submission ID is then retrieved, and the existing submission is removed.
+    If there is no matching programme for the current round, a new submission ID is generated.
+
+    :param workbook: a dictionary of DataFrames containing table data to determine the submission ID.
+    :return: a string representing the submission ID.
+    """
+    reporting_round = int(workbook["Submission_Ref"]["Reporting Round"].iloc[0])
+    programme_id = workbook["Programme_Ref"]["Programme ID"].iloc[0]
+    # if already added this round, this entity used to drop existing round data
+    programme_exists_same_round = (
```
nsnape, line 119: as per other comment, could consider splitting DB queries into separate functions that live in queries module?

core/controllers/load_functions.py
```diff
@@ -0,0 +1,257 @@
...
+
+    organisations_without_child = Organisation.query.filter(~Organisation.programmes.any()).all()
+    if organisations_without_child:
+        for organisation in organisations_without_child:
+            db.session.delete(organisation)
+
+    db.session.commit()
+
+
+def get_programmes_same_round_or_older(reporting_round: int, programme_ids: list[str]) -> dict[str, str]:
```
nsnape, line 198: I wonder, should we split these historical ingest methods into a different module, something like ingest_historical or load_functions_historical? Just to keep them separate from the core ingest code? 

Might make it easier for people to mentally de-couple the different ingest methods going forwards

nsnape (changes requested):
In general this is looking loads better than before, should be much clearer to follow. I've added a few layout / org comments but functionally it looks like a great start to making this more reusable.

core/controllers/ingest.py
Joel-Pearce: Well, I imagine the difference between funds will be that some will not load any data into certain tables. We have the same situation with historical data. Our solution to that has been to iterate through the mapping but to 'jump over' explicitly excluded tables. If we move away from the mapping we may be able to configure this elsewhere.

Joel-Pearce: Otherwise something like, ``` if load_function := TABLE_TO_LOAD_FUNCTION.get(mapping.table)```. That way if there is no such table for a specific fund it will be None and the line can be skipped.

core/controllers/load_functions.py
Joel-Pearce: Yep, makes sense. Can split that out.

core/controllers/load_functions.py
Joel-Pearce: Can do

core/controllers/load_functions.py
Joel-Pearce: Yes, probably a good idea

core/controllers/ingest.py
Joel-Pearce: ingest.py has no concept of ```generic_load``` so it would need to be imported. This is probably a quicker and less complicated look-up

Connor-OS: Ah yeah, those functions are stored in a separate module. This way is probably best then

core/controllers/load_functions.py
Joel-Pearce: I've abstracted out the deletion based on submission to a separate function. However, the code used to assess whether to retrieve a submission id if it exists is the same as the code used to determine if we should delete all of a submission's children. I've now got a new 'master' function - upsert_or_get_submission_id() - which calls the deletion function if it needs to upsert. This way we can test the functionality separately without repeating code. 

The other possibility is repeating the code that does the checks if the submission id already exists for a given round.

nsnape (changes requested):
Looking great now, added a few comments, mostly just general house-keeping ones really. Ta

core/controllers/load_functions.py
```diff
@@ -0,0 +1,219 @@
...
+    worksheet = workbook[mapping.table]
+    models = mapping.map_data_to_models(worksheet)
+
+    if programme_exists_previous_round:
+        # Set incoming model pk to match existing DB row pk (this record will then be updated).
+        programme_to_merge = models[0]
+        programme_to_merge.id = programme_exists_previous_round.id
+        db.session.merge(programme_to_merge)
+    else:
+        db.session.add_all(models)
```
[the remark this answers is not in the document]
Joel-Pearce: This results in ```sqlalchemy.orm.exc.UnmappedInstanceError: Class 'builtins.list' is not mapped```

core/controllers/load_functions.py
```diff
@@ -0,0 +1,219 @@
...
+    organisations_without_child = Organisation.query.filter(~Organisation.programmes.any()).all()
+    if organisations_without_child:
+        for organisation in organisations_without_child:
+            db.session.delete(organisation)
+
+    db.session.commit()
+
+
+# a mapping of the table name to the function that handles the specific table's loading behaviour
+TABLE_TO_LOAD_FUNCTION = {
```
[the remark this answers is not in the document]
Joel-Pearce: Until we have onboarded other funds it's difficult to envision the specifics of how we're going to adjust the parameters necessary to change variables and methods tailored to each fund. I have created a function with comments that currently just returns this mapping and makes clear it's a tf mapping.

core/controllers/load_functions.py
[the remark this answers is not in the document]
Joel-Pearce: I suppose we'd have to do ```db.session.add(models[0])```? Not sure if that's any clearer or not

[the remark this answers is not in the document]
nsnape: I think unpack the Programme instance directly after grabbing the attribute from mapping. Since it also gets unpacked with `programme_to_merge = models[0]` inside the if block. So just move that line outside the block and rename it to something more generic? ie:

```
    models = mapping.map_data_to_models(worksheet)
    programme = models[0]

    if programme_exists_previous_round:
        programme.id = programme_exists_previous_round.id
        db.session.merge(programme)
    else:
        db.session.add(programme )
```

core/controllers/load_functions.py
```diff
@@ -0,0 +1,233 @@
...
+    a new submission ID is generated.
+
+    :param workbook: a dictionary of DataFrames containing table data to determine the submission ID.
+    :return: a string representing the submission ID, and the Submission to delete
+    """
+    reporting_round = int(workbook["Submission_Ref"]["Reporting Round"].iloc[0])
+    programme_id = workbook["Programme_Ref"]["Programme ID"].iloc[0]
+    programme_exists_same_round = get_programme_by_id_and_round(programme_id, reporting_round)
+
+    submission_to_del = None
```
nsnape, line 137: can we move this to inside the else block, otherwise it gets re-assigned in the if block.

nsnape (changes requested):

nsnape (approved):

PR #403, opened 2023-12-20
Feature/fmd 186 open api housekeeping

cyrusdobbs:
https://dluhcdigital.atlassian.net/browse/FMD-186

### Change description
- General OpenAPI house keeping (e.g. improving consistency, updating out of date values, removing depreciated code)
- Removes the definition of "schemas" unless required or used in multiple locations
- Combines all OpenAPI documentation into a single file for easier reading
- Modifies the `download` controller to use [Connexion's request handling features](https://connexion.readthedocs.io/en/latest/request.html)

### How to test


core/controllers/download.py
```diff
@@ -8,53 +8,46 @@
...
+def download(
+    file_format: str,
+    funds: list[str] = None,
+    organisations: list[str] = None,
+    regions: list[str] = None,
+    rp_start: str = None,
+    rp_end: str = None,
+    outcome_categories: list[str] = None,
+) -> Response:
+    """Query the database with the provided parameters and serialised the resulting data in the specified format.
```
Joel-Pearce, line 27: Should be "and serialise"?

openapi/api.yml
```diff
@@ -1,36 +1,38 @@
...
+  version: 1.0.0
 
 paths:
   /download:
     get:
       operationId: core.controllers.download.download
-      summary: Returns some data in the requested format.
-      description: Takes a file format and returns some data in that format.
+      summary: Get monitoring and evaluation data from the data-store.
+      description: Query the database with the provided parameters and serialised the resulting data in the specified format.
```
Joel-Pearce, line 12: And here

Joel-Pearce (changes requested):

core/controllers/download.py
```diff
@@ -118,3 +111,17 @@ def sort_output_dataframes(df: pd.DataFrame, sheet: str) -> pd.DataFrame:
     df.reset_index(drop=True, inplace=True)
 
     return df
+
+
+def date_to_string(obj) -> str:
+    """Converts a datetime object suitable for JSON serialisation.
+
+    :param obj: A datetime object (or date) that want to serialise into a string.
```
Joel-Pearce, line 119: "A datetime object (or date) to be serialised into a string."?

Joel-Pearce (changes requested):

core/controllers/download.py
nsnape: I have a few gripes with this function in general (the author refused to address them on PR). But basically:

- The function docstring has specific use case in it, but really, this function has a more general purpose as a black-box for casting datetime objects to string. There could be a case for changing the docstring to be more generic, as it could feasibly be re-sued outside the current use case of extending the built-in JSON serialiser.
- The function itself feels more like a general util than anything download specific, maybe move it to utils module?

nsnape (changes requested):

core/controllers/download.py
nsnape: Also, that param isn't specifically typed to only take date or datetime object, I wonder if we should either update the type hint, or at least make the docstring param clear that it takes any object, and that it will attempt to cast to a str rep of date, otherwise raise

nsnape: probably the latter

cyrusdobbs: I didn't really read this function, just moved it as I thought the module made more sense with the main `download` function at the top.

Agree with you though - I'll make some changes in line with your points

Joel-Pearce (approved):
LGTM

nsnape (approved):
LGTM

PR #404, opened 2023-12-20
BAU: tidy pre-transformation docstrings

Joel-Pearce:
Add additional context to and tidy docstrings pertaining to pre-transformation validation.


core/validation/pre_transformation_validate.py
```diff
@@ -1,3 +1,10 @@
+"""Module for methods relating to pre-transformation validation.
+
+Pre-transformation validation occurs before any stages in the ETL pipeline.
+It validates user input based on the data as it appears in a dictionary of
+DataFrames read directly from the submission file.
+"""
```
cyrusdobbs, line 6: May be able to get across a bit clearer that it's primary purpose is to do some high level "submission-level" checks before extraction/transformation and validation of the data itself

core/validation/pre_transformation_validate.py
```diff
@@ -18,33 +25,44 @@ def pre_transformation_validations(
...
 
 
 def authorisation_validation(workbook: dict[str, pd.DataFrame], auth: dict, schema: dict):
-    """Performs authorization validation based on the provided schema.
+    """Performs authorisation validation based on the provided schema.
+
+    An 'auth' dictionary is sent to the /ingest API when the user attempts to submit.
+    This dictionary contains variables against which the submission form is checked
+    to ensure that the user is able to submit the selected information based on data
+    collected from the frontend of the application.
```
cyrusdobbs, line 59: "is able to submit the selected information based on their user level permissions, which is handled by the frontend" may be more clear?

core/validation/pre_transformation_validate.py
```diff
@@ -1,3 +1,10 @@
+"""Module for methods relating to pre-transformation validation.
```
cyrusdobbs, line 1: nitpicky 😃  but its a module of functions rather than methods

core/validation/pre_transformation_validate.py
```diff
@@ -18,33 +25,44 @@ def pre_transformation_validations(
...
+        return
 
     schema = REPORTING_ROUND_TO_PRE_TRANSFORMATION_SCHEMA[reporting_round]
 
-    wrong_input_validation(workbook, schema)  # first, check if valid
+    wrong_input_validation(workbook, schema)
     conflicting_input_validation(workbook, schema)
 
     if reporting_round == 4 and auth:
         authorisation_validation(workbook, auth, schema)
```
cyrusdobbs, line 50: Just thinking, instead of passing reporting round etc. We could refactor this so the schema determines the checks. i.e. if the schema has wrong input checks then you do that, but if it didn't include any authorisation then don't execute that. 

Means this is even more schema driven and would help with extendibility to get rid of the concept of reporting rounds in the code itself

I realise this is a bit out of scope but may be worth doing for clarity.. We can then document about the difference in validation functionality where the schema is defined, rather than here - which should really be round/fund context free

cyrusdobbs (commented):
good stuff 👍

core/validation/pre_transformation_validate.py
Joel-Pearce: We can do this with:

```
    if schema.get("authorisation_checks"):
        authorisation_validation(workbook, auth, schema)
```

If the schema specifies the functions that are to be checked then we will have to use a string representation of each function in the schema (to avoid a circular import), and map these back to a set of functions via a dictionary somewhere else in the code. 

The above code snippet is a bit easier and means we escape reliance on a reporting round to determine if ```authorisation_validation``` is called. If there ends up being a ballooning of pre-transformation validation functions then this can be revisited, but for now I think it's good enough?

cyrusdobbs: We still rely on `reporting_round` above but we'll refactor that out later. This change looks fine for now. 

Although you've removed the check on whether `auth` is truthy, which has broken ingest. 

We should still check `auth` exists inline with the old code. I'd recommend adding that back and ideally adding a test to cover this scenario as it should have made a test fail.

Joel-Pearce: Nice spot - I overlooked the removal/importance on the check on whether ```auth``` is truthy. Added an extra test case in to account for that.

cyrusdobbs: nice 👍

cyrusdobbs (approved):
LGTM

PR #414, opened 2024-01-10
SMD-495 clarify flask/docker instructions on the readme

bobbi-hbrown:
https://dluhcdigital.atlassian.net/browse/SMD-495

Part of our process to audit and update all of the setup instructions on each README, to ensure correct setup for new starters.


README.md
```diff
@@ -136,7 +139,7 @@ To run the app alongside other related microservices see https://github.com/comm
 docker build -t communitiesuk/funding-service-design-post-award-data-store .
 docker run -p 8080:8080 communitiesuk/funding-service-design-post-award-data-store
 ```
-App should be available at `http://localhost:8080`
+App should be available at `http://localhost:4001`
```
cyrusdobbs, line 142: Looks like this was correct as is

If it's built and run as a standalone contain then it uses the port defined in the `Dockerfile` here
https://github.com/communitiesuk/funding-service-design-post-award-data-store/blob/f87dc108aa3ca9c6bdf51c0890bd5230994b1f2d/Dockerfile#L11

as opposed to the port defined in the docker-runner repo `docker-compose.yml` which is `4001`

README.md
```diff
@@ -101,7 +104,7 @@ so that the migrations will be run in the pipelines to correctly upgrade the dep
 
 ## CLI Commands
 
-These commands can be run from the command-line from a terminal using the same python environment as a running flask application.
+If running the data-store app locally using Flask (i.e not through the Docker runner), these commands can be run from the command-line from a terminal using the same python environment as the running flask application.
```
cyrusdobbs, line 107: I think this could currently be a bit misleading, as you can still run these commands if it's running in Docker - you'd just have to use the containers console

cyrusdobbs (changes requested):

cyrusdobbs (approved):
LGTM

PR #417, opened 2024-01-12
FMD-185 Instructions for extending serialisation methods

bobbi-hbrown:
https://dluhcdigital.atlassian.net/jira/software/c/projects/FMD/boards/139/backlog?selectedIssue=FMD-185

How to extend serialisation - I moved information from the docstring to the top of the file, so as not the overwhelm the docstring with information that actually applied to the whole module. Any feedback on this placement is appreciated


core/extraction/towns_fund_round_two.py
```diff
@@ -1370,7 +1371,7 @@ def convert_date(date_str: str) -> tuple[datetime, datetime]:
     return start_date, end_date
 
 
-def get_actual_forecast(row: pd.Series) -> str:
+def get_actual_forecast(row: pd.Series) -> str | float:
```
nsnape, line 1374: is it possible to be a bit more explicit here and make that return type `str | np.nan`? I think int might be abit misleading

core/serialisation/data_serialiser.py
```diff
@@ -43,6 +43,19 @@
     risk_register_query,
 )
 
+# This project uses the Marshmallow library to define the serialization schema. Additional classes can be defined inside
```
nsnape, line 46: Please can you move this into the module level docstring? (that's basically the multi-line string at the top of the file)

nsnape (changes requested):

core/extraction/towns_fund_round_two.py
bobbi-hbrown: I'm getting a type error when I change it to np.nan explicitly, as one is a type and the other isn't. I can substitute type(np.nan) instead, though not sure if that's standard practice

nsnape: ok my bad, I looked this up and float is correct for np.nan (albeit a little misleading and vague)

core/serialisation/data_serialiser.py
```diff
@@ -1,4 +1,20 @@
...
+
+Each class defines the fields it serialises by directly importing attributes from DB model definitions in
+core/db/entities.py via the auto_field method. Defining a model in the class Meta sets the default model for any
+serialization class, but ones from other models (looked up via joins etc) can be explicitly defined as a KWARG to
+auto_field when defining the attribute.
+
+Each class should correspond to a table definition in core/db/entities.py, ensuring that the attribute name matches
+the DB field name as defined in entities. The data_key argument is the corresponding field name that will be output
+into the data extract, allowing for easy customization. The order of attributes defined in each class is the order
+they will appear in any download, providing a configurable way to control the output order.
```
nsnape, line 15: Probs just change this last line slightly as "configurable" kind of implies it's config driven (which this isn't). Probs something more like "providing an easy way to customise the output order"

nsnape (changes requested):
one tiny change

nsnape (approved):
Approved. Please squash the commits on merge. Ta

PR #407, opened 2024-01-03
BAU: fix fixture score for AWS integration tests

Connor-OS:
Move out the fixture to create and delete the unit test buckets for to conftest file so that they can be shared by all integration tests

### How to test
_If manual testing is needed, give suggested testing steps_

PR #420, opened 2024-01-15
[BAU] fix scoping of aws test fixtures

cyrusdobbs:
### Change description
Current scoping of an autouse module-level aws fixture in `integration_tests/contest` meant that it was being used in every test in `integration_tests`. This meant if localstack was not reachable then every test would fail. 

Moved all AWS fixtures to `test_aws.py` as they aren't used by any other test module and so don't need to be shared via a conftest.

### How to test


Connor-OS (changes requested):
THis appears to be the opposite of my PR here: https://github.com/communitiesuk/funding-service-design-post-award-data-store/pull/407/files

I initally moved it out to conftest as that test bucket is required in some of the integration tests so needs to be spun up for those as well. I belive this will sometimes break those tests locally

cyrusdobbs:
> THis appears to be the opposite of my PR here: https://github.com/communitiesuk/funding-service-design-post-award-data-store/pull/407/files
> 
> I initally moved it out to conftest as that test bucket is required in some of the integration tests so needs to be spun up for those as well. I belive this will sometimes break those tests locally.
> 
> specifically I belive it was: `test_ingest_with_r4_corrupt_submission`

The test you mention doesn't currently require that fixture and passes with this change. This is because an error is just logged if the S3 bucket can't be reached but it doesn't exit the code.

I suppose we should either:
1. undo this change (but turn `autouse` off) and actually test the AWS integration here by making some assertion that the S3 bucket is used as part of this integration test
2. leave the AWS integration untested here, leaving it to be tested separately as is

The second option seems ok IMO, especially as the endpoint can function without S3 anyway (it just logs to Sentry if it can't access it)

Connor-OS:
> > THis appears to be the opposite of my PR here: https://github.com/communitiesuk/funding-service-design-post-award-data-store/pull/407/files
> > I initally moved it out to conftest as that test bucket is required in some of the integration tests so needs to be spun up for those as well. I belive this will sometimes break those tests locally.
> > specifically I belive it was: `test_ingest_with_r4_corrupt_submission`
> 
> The test you mention doesn't currently require that fixture and passes with this change. This is because an error is just logged if the S3 bucket can't be reached but it doesn't exit the code.
> 
> I suppose we should either:
> 
>     1. undo this change (but turn `autouse` off) and actually test the AWS integration here by making some assertion that the S3 bucket is used as part of this integration test
> 
>     2. leave the AWS integration untested here, leaving it to be tested separately as is
> 
> 
> The second option seems ok IMO, especially as the endpoint can function without S3 anyway (it just logs to Sentry if it can't access it)

Also happy with the second option, if you can edit the integration tests to always pass locally.

We had some confusing behaviour last time. 
- All tests passed on the pipeline
- On Joels machine they every time
- On Noel's, Bobbi's and my machine the integration tests requiring S3 passed most of the time but sometimes failed, looking a the localstack image I could see it was returning an error "no such bucket" every time the test was run

Connor-OS (approved):
Approving, because this works on the pipeline, and added unit test to cover the NoSuchBucket error. We will see if anyone else has any trouble running these tests locally. Could do with another reviewer

PR #424, opened 2024-01-22
FMD-156 Store ingest files in S3

Connor-OS:
Basic functionality to save file to S3 on successful ingest.

Currently both functionalities are maintained so files are first saved to the db, and then uploaded to S3 this is just the first step towards the transition to storing in S3.

We have chosen to save files using there submission ID as the file name, this will mean that when the same submission is reuploaded it will overwrite the exisitng file in S3. If this system is used with a version controled S3 bucket with the retention period set to an acceptable threshold, we can balance maintaining old file versions for restoring to and reducing the required storage.

### Change description

### How to test


tests/controller_tests/test_retrieve.py
cyrusdobbs, line 1: I think if the unit tests here use localstack then we should move this module to `integration_tests`. We can also then pull the `create_bucket` and `delete_bucket` out into a `conftest.py` rather than importing them like normal functions.

Also wondering if we could create a generic `setup_teardown_buckets` fixture that takes a set of bucket names within the `conftest` that looks like:

```
@pytest.fixture
def setup_teardown_buckets(buckets: list[str]):
    for bucket in buckets:
        create_bucket(bucket)
    yield
    for bucket in buckets:
        delete_bucket(bucket)
```

and then within `test_retrieve` and `test_aws` we can just call with auto-use like (in the case of `test_retrieve`)

```
@pytest.fixture(autouse=True, scope="module")
def test_buckets(setup_teardown_buckets):
    """Sets up and tears down buckets used by this module.
    On set up:
    - creates data-store-successful-files-unit-tests
    On tear down, deletes all objects stored in the buckets and then the buckets themselves.
    """
    yield setup_teardown_buckets(buckets=[Config.AWS_S3_BUCKET_SUCCESSFUL_FILES])
```

tests/controller_tests/test_retrieve.py
```diff
@@ -1,4 +1,35 @@
...
+
+
+@pytest.fixture()
+def uploaded_mock_file():
+    """Uploads a mock generic file and deletes it on tear down."""
+    fake_file = io.BytesIO(b"0x01010101")
+    fake_filename = "S-R03-1"
+    _S3_CLIENT.upload_fileobj(fake_file, Config.AWS_S3_BUCKET_SUCCESSFUL_FILES, fake_filename)
+    yield
+    _S3_CLIENT.delete_object(Bucket=Config.AWS_S3_BUCKET_SUCCESSFUL_FILES, Key=fake_filename)
```
cyrusdobbs, line 32: We have a similar fixture in `test_aws.py` - we should be able to pull out most of the logic from both and reuse it
https://github.com/communitiesuk/funding-service-design-post-award-data-store/blob/197952210b27ddfc87ea5343409d5f30fdb4098f/tests/integration_tests/test_aws.py#L58-L65

core/controllers/ingest.py
```diff
@@ -338,6 +340,15 @@ def save_submission_file(excel_file, submission_id):
     db.session.commit()
 
 
+def save_submission_file_S3(excel_file, submission_id):
```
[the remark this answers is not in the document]
nsnape: could we have type hints too?

core/controllers/retrieve_db.py
```diff
@@ -0,0 +1,30 @@
...
+    Select file by:
+    - submission_id
+
+    :return: Flask response object containing the requested spreadsheet.
+    """
+    query = select(Submission.submission_file, Submission.submission_filename, Submission.reporting_round).where(
+        Submission.submission_id == submission_id
+    )
+
+    try:
```
nsnape, line 25: rather than a try except, you could try using the one_or_none alchemy method

core/controllers/retrieve_db.py
```diff
@@ -0,0 +1,30 @@
+from io import BytesIO
```
nsnape, line 1: could we rename the module, retreive_db is a bity vague and misleading

nsnape (changes requested):

core/controllers/retrieve_db.py
```diff
@@ -0,0 +1,30 @@
...
+from flask import abort
+from sqlalchemy import select
+from sqlalchemy.exc import NoResultFound
+
+from core.const import EXCEL_MIMETYPE
+from core.db import db
+from core.db.entities import Submission
+
+
+def retrieve_db(submission_id):
```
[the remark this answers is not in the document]
Connor-OS: The reason for the poor naming is becasue these functions are intended to be removed once we remove all the files from the db and delete this column

core/handlers.py
```diff
@@ -2,34 +2,11 @@
...
-            "id": failure_uuid,
-            "status": 500,
-            "title": "Internal Server Error",
-            "internal_errors": None,
-        }, 500
-
-    return uncaught_exception
-
-
 def save_failed_submission(file: BinaryIO):
```
cyrusdobbs, line 10: We could move this to `ingest` now that the handlers have been deleted

copilot/data-store/addons/data-store-successful-files.yml
```diff
@@ -0,0 +1,83 @@
...
+          - Sid: S3ObjectActions
+            Effect: Allow
+            Action:
+              - s3:GetObject
+              - s3:PutObject
+            Resource: !Sub ${ datastoresuccessfulfilesBucket.Arn}/*
+          - Sid: S3ListAction
+            Effect: Allow
+            Action: s3:ListBucket
+            Resource: !Sub ${ datastoresuccessfulfilesBucket.Arn}
```
gidsg, line 72: Not sure that the service needs to list files, since it can should know the key?
So can we remove this permission?

core/aws.py
```diff
@@ -27,38 +27,34 @@
...
     :param object_name: S3 object name
     :return: retrieved file as a BytesIO
     """
-    file = BytesIO()
     try:
-        _S3_CLIENT.download_fileobj(bucket, object_name, file)
+        response = _S3_CLIENT.get_object(Bucket=bucket, Key=object_name)
     except (ClientError, EndpointConnectionError) as bucket_error:
         current_app.logger.error(bucket_error)
         return None
```
gidsg, line 55: Would it be better to re-throw the exception after logging. Is it ambiguous to return `None`?

core/controllers/ingest.py
```diff
@@ -83,6 +85,18 @@ def ingest(body: dict, excel_file: FileStorage) -> tuple[dict, int]:
...
+        current_app.logger.error(
+            f"Uncaught ingest exception: {type(e).__name__}: {str(e)}, - failure_id={str(failure_uuid)}", exc_info=True
+        )
+        return {
+            "detail": f"Uncaught ingest exception: {type(e).__name__}: {str(e)}",
+            "id": failure_uuid,
+            "status": 500,
+            "title": "Internal Server Error",
+            "internal_errors": [],
+        }, 500
```
cyrusdobbs, line 99: Just a thought - this could be encapsulated in a function like `process_uncaught_exception` to keep this top level function easier to skim over

cyrusdobbs: also it's best practice to use a more descriptive variable name than `e`

MarcUsher: Makes a lot of sense, thanks Cyrus! Just making these changes and will push an update shortly.

nsnape (changes requested):

nsnape (approved):

core/aws.py
```diff
@@ -27,38 +25,29 @@
...
+    :param metadata: optional dictionary containing metadata for upload
     :return: True if successful else False
     """
     file.seek(0)
-    try:
-        _S3_CLIENT.upload_fileobj(file, bucket, object_name)
-    except (ClientError, EndpointConnectionError) as bucket_error:
-        current_app.logger.error(bucket_error)
-        return False
+    _S3_CLIENT.upload_fileobj(file, bucket, object_name, ExtraArgs={"Metadata": metadata if metadata else {}})
```
gidsg, line 38: Can we set the content type on the uploaded file also?

core/controllers/retrieve_submission_file_db.py
```diff
@@ -9,10 +9,15 @@
...
 from core.db.entities import Submission
 
+# TODO: [FMD-227] Remove submission files from db
 
-def retrieve(submission_id):
+
+def retrieve_submission_file_db(submission_id):
     """Handle the download request and return the originally submitted spreadsheet.
 
+    This end point is deprecated, as files are intended to be saved in S3 from now on. The functionality to etrieve from
```
gidsg, line 18: fix typo:
```suggestion
    This end point is deprecated, as files are intended to be saved in S3 from now on. The functionality to retrieve from
```

gidsg, line 18: Maybe prefix this comment with `TODO:`?

gidsg (approved):

PR #426, opened 2024-01-23
[SMD-500] implement the new validation

cyrusdobbs:
### Change description

- creates `tables` as a root level package outside of the main application code that leverages [`pandera`](https://pandera.readthedocs.io/en/stable/dataframe_schemas.html) for `pd.DataFrame` validation
- `tables` implements a class `TableSchema` used to define a table's ID tag, columns, error messages and extraction parameters
- `TableSchema` implements two public methods - `extract` and `validate` 
- `extract` can be passed a "worksheet" in the form of a `pd.DataFrame` and any tables marked with the relevant ID tag will be extracted and returned
- `validate` can be passed extracted tables, which are then validated against using the types and `pa.Check`s defined in the `TableSchema.columns`
- `TableSchema.messages` will need to be passed in order for `validate` to return messages depending on the type of error raised, these messages can be mapped to either an error type or an error type and a column. This allows for messages to be defined at either table scope or column scope for each type of error.

### How to test
- `scripts/new-validation.py` has been added with an example use case that can be modified and experimented with for testing etc


cyrusdobbs:
Outstanding considerations:
- behaviour of the `TableSchema` may be unstable if attributes are modified post-init. Thinking that it's either best to make sure this is robust, with the use of properties and setters or to just make the object instances "immutable"
- `pa.validate` carries out validation and type "coercion", but if there are any errors then it doesn't return the coerced table. This means that downstream processes may have to use a mix of correctly types tables and some that are all `str`/`pd.object`.. this doesn't immediately seem to be an issue because our current transformation layer returned all columns as strings but this may mean that further downstream validation checks need to consider that columns may not be typed accurately (e.g. ints are still strings) OR we need to ensure that all tables are coerced that are returned from `TableSchema.validate` even if there are validation errors

scripts/new-validation.py
```diff
@@ -0,0 +1,73 @@
...
+            "Likelihood score": pa.Column(str, pa.Check.isin(scores)),
+            "Impact score": pa.Column(str, pa.Check.isin(scores)),
+            "Total risk score": pa.Column(int),  # Comment this out and this column won't be extracted and validated
+            "Mitigations": pa.Column(str),
+        },
+        row_idxs_to_drop=[0],  # Drop the first row with help text
+        drop_empty_rows=True,
+        # num_dropped_columns=1,  # Uncomment this if you remove "Total risk score" from the columns
+        messages={
+            "isin": "You’ve entered your own content, instead of selecting from the dropdown list provided. "
```
Joel-Pearce, line 39: Not sure if ```isin``` is the most intuitive key. Something like ```enum_failure``` (which I think this always relates to?) might be better for readability.

scripts/new-validation.py
```diff
@@ -0,0 +1,73 @@
...
+    validated_tables = []
+    all_errors = []
+    for table in tables:
+        validated_table, errors = risk_schema.validate(table)
+        if validated_table is not None:
+            validated_tables.append(validated_table)
+        if errors:
+            all_errors.append(errors)
+
+    print(all_errors)
```
Joel-Pearce, line 69: Presumably this print statement was put in for debugging purposes, but is it still something we want in the code inside of e.g logging?

tables/schema.py
```diff
@@ -0,0 +1,281 @@
...
+
+            # "dtype" check is series-wise and so will not produce element-wise messages. The solution to
+            # this is to use "coerce_dtype" errors instead as these are raised per element.
+            if error_type == "dtype":
+                continue
+
+            if error_type == "column_in_schema":
+                # "column_in_schema" means that the table didn't fit the schema - dev error rather than user
+                # TODO: should we use a different exception?
+                raise TableExtractError("Validated table contains columns from outside of the schema")
```
Joel-Pearce, line 146: ```TableExtractError``` sounds correct for this use case. Is there a way of including these extraneous column names in the message or error?

tables/schema.py
```diff
@@ -0,0 +1,281 @@
...
+            column_letter = table.header_to_letter[failure.column]
+
+            # first try error type + column, then error_type
+            description = self.messages.get((error_type, failure.column)) or self.messages.get(error_type)
+            if not description:
+                raise ValueError(f"No message configured for column {failure.column} with error type {error_type}")
+
+            error_messages.append(
+                ErrorMessage(
+                    self.worksheet_name, self.section, f"{column_letter}{failure.index + 1}", description, error_type
```
Joel-Pearce, line 157: Why ```failure.index + 1```? Perhaps a comment to explain this as glancing at the code I don't see the reason.

tables/schema.py
```diff
@@ -0,0 +1,281 @@
+from collections import defaultdict
```
nsnape, line 1: What is a table schema? How is it used etc? Could this go in a module docstring here?

nsnape (changes requested):

tables/tests/resources/test_worksheet.csv
```diff
@@ -0,0 +1,103 @@
...
+,,,,,,,,,,,,,,,,,,,,,,,,,,,,
+,TESTID12S,,,,,,,,,,,,,,,,,,,,,,,,,,,
+,Column1,,,,,,,,,,,,,,,,,,,,,,,,,,,
+,A,,,,,,,,,,,,,,,,,,,,,,,,,,,
+,,,,,,,,,,,,,,,,,,,,,,,,,,,,
+,,,,,,,,,,,,,,,,,,,,,,,,,,,,
+,TESTID13S,,,,,,,,,,,,,,,,,,,,,,,,,,,
+,Column1,,,,,,,,,,,,,,,,,,,,,,,,,,,
+,A,,,,,,,,,,,,,,,,,,,,,,,,,,,
+,,TESTID13E,,,,,,,,,,,,,,,,,,,,,,,,,,
```
Joel-Pearce, line 103: Apologies if I've missed something similar, but a good test case might be:
```
col1,col2,col3
useful_data,useful_data,useful_data
guidance notes, guidance notes, guidance notes
useful_data,useful_data,useful_data
```
This is a pattern we might encounter?

Joel-Pearce: Are we expecting all tables to be vertical like a DataFrame (with columns names on top of all data) from now on? Is there not a circumstance where the columns will appear horizontally in the original sheet, and if so how do we account for this?

Joel-Pearce (changes requested):
Really like the standardisation of extraction & transformation. I've only looked at the code once, and will probably need to go through it with an interactive debugger before being able to get a proper understanding. I would also recommend somebody working on Submit reviewing.

scripts/new-validation.py
cyrusdobbs: This is just a standalone script so I don't think it matters

tables/schema.py
cyrusdobbs: yeah good shout

tables/schema.py
cyrusdobbs: This is because Excel files are 1-indexed - will add a comment 👍

tables/tests/resources/test_worksheet.csv
cyrusdobbs: To your first comment that is definitely something we may encounter but it's actually already tested here (specifically line 249). Agree it isn't that clear from the test that it can be used in the "guidance notes" context that we see often in the actual spreadsheets as the parameter is more general than that

https://github.com/communitiesuk/funding-service-design-post-award-data-store/blob/dd080490bc4f471db8e00c195adbc1b67863aa72/tables/tests/test_schema_extract.py#L242-L264


And to your second one, yeah I've been thinking about that - we'll definitely come across those style of tables. I'm planning to implement like a `transposed` flag or something, that will just flip the extracted table during processing and make the first column the header row. I think it should be ok. It will need to be implemented for Pathfinders, but I may not do it as part of this PR.

tables/schema.py
cyrusdobbs: Added some guidance to the module docstring with an example

tables/schema.py
```diff
@@ -0,0 +1,319 @@
...
+    - dropping "non-data" rows specified in the schema (e.g. rows containing descriptions / help text)
+    - dropping empty rows
+
+Extracted tables can then be passed to TableSchema.validate(), which leverages the OS library Pandera to validate the
+data against any datatypes and pa.Checks defined in the schema.
+
+Any validation errors raised by pandera are then converted to ErrorMessage objects that refer back to the source Excel
+file and should contain guidance on how to correct any errors. TableSchema.messages will need to be set in order for
+validate to return messages. These messages can be mapped to either an error type or an error type and a column. This
+llows for messages to be defined at either table scope or column scope for each type of error.
```
nsnape, line 25: typo

tables/utils.py
```diff
@@ -0,0 +1,30 @@
+import pandas as pd
+
+
+def column_index_to_excel_letters(col_idx: int) -> str:
+    """Converts an integer column index (0-indexed) to Excel column letters.
```
nsnape, line 5: this may seem a bit picky, but if passed a negative integer as col_idx (say last position in Python style indexing) it will return "", but should it just error? I.e. only accept positive integers?

nsnape: in a similar vein, the max number of columns in Excel in 16,384, so could stipulate the input index must be between 0 and 16,384 otherwise ValueError?

nsnape (changes requested):

tables/utils.py
cyrusdobbs: good shout - have added some input validation this function now

Joel-Pearce (approved):
LGTM

nsnape (approved):

PR #428, opened 2024-01-31
SMD-561 - Refactor initial validation

wjrm500:
### Ticket
[SMD-561](https://dluhcdigital.atlassian.net/browse/SMD-561)

### Suggested work
Short-circuiting initial validation particularly by avoiding ValidationFailureBase and MessengerBase.

### Implementation summary
The `Check` and `ConflictingCheck` classes used in `core/validation/initial_validation/schemas.py` have been changed from simple namedtuples to proper classes that live in their own module, the new `core/validation/checks.py`. The idea is to encapsulate checking logic within the checks themselves. Previously the checking logic sat inside `core/validation/initial_validation/validate.py`, with one function per check type (e.g., `‎conflicting_input_validation`) for calculating expected values and calling a common function `check_values` that checked that the actual value was one of the expected values. Now that the logic sits inside the checks themselves, the schema from `schemas.py` is simply a list of checks, and the `initial_validate` function can simply iterate through the checks and call the `Check.run` method, passing in the workbook.

The `Check` class no longer includes a property `type` and the `validate.py` module no longer creates instances of `ValidationFailureBase`, that were previously decoded by `MessengerBase` into instances of `Message` in a process that was so convoluted that it gave rise to this very ticket. Instead, each check comes with its own error message, with support for parameterisation, and any error messages generated during check running are assigned to the `error_messages` property (a simple list of strings) of a new exception class `InitialValidationError`, which is caught in `core/controllers/ingest.py` and handled by a new method `process_initial_validation_errors`. This method simply returns the `error_messages` as part of the JSON body of a 400 response. The new exception class and response handling method were created to avoid interference with the existing logic used for further schema validation. The idea will be that when the new schema validation changes have been merged, we can look at unifying more of the code here.

### Further thoughts / future work
Initially we thought about modifying the structure of the JSON response from the `/ingest` endpoint to unify the errors under a single property name `errors` (see differences in current responses returned from `process_internal_failures` and `process_user_failures`), but this meant Jinja changes in the `submit` repo so we decided against.

Also considered was further simplification to `initial_validate`, such that instead of batching up checks from the schema into `basic`, `conflicting` and `authorisation`, we would simply iterate over the checks and run each one in turn, collecting all error messages together. The reason we didn't do this in the end is because we wanted to maintain the current behaviour, whereby conflicting checks are only run if basic checks pass, and authorisation checks are only run if conflicting checks pass, which prevents situations where a missing place name for example gives three error messages, of which only the first one is really useful:

```
[
    "Cell E8 in the "project admin" must contain a place name from the dropdown list provided. Do not enter your own content.",
    "We do not recognise the combination of fund type and place name in cells E7 and E8 in "project admin". Check the data is correct.",
    "You're not authorised to submit for None. You can only submit for Newark.",
]
```

A solution to allow simple iteration over checks (no batching based on instance type, or based on keys in a dictionary version of the schema, as is currently done) without excessive error messaging was suggested, but dismissed in the end in favour of reducing complexity. This solution was all about facilitating the creation of dependencies between instances of `Check`, and would have looked something like this in `schemas.py`:

```
...
place_name_exists = Check(
    sheet="2 - Project Admin",
    column=6,
    row=4,
    expected_values=tuple(TF_PLACE_NAMES_TO_ORGANISATIONS.keys()),
    error_message="Cell E8 in the "project admin" must contain a place name from the dropdown list provided. Do not enter your own content.",
)
place_name_matches_fund = MappedCheck(
    sheet="2 - Project Admin",
    column=5,
    row=4,
    dynamic_params={
        "mapped_column": 6,
        "mapped_row": 4,
        "mapping": PLACE_TO_FUND_TYPE,
    },
    error_message_with_placeholders="We do not recognise the combination of fund type and place name in cells E7 and E8 in "project admin". Check the data is correct.",
    dependencies = [place_name_exists],
)
...
TF_ROUND_4_INIT_VAL_SCHEMA = [
    ...,
   place_name_exists,
   place_name_matches_fund,
   ...
]
```

With `Check.run` modified to something like this:

```
def run(self, workbook: dict[str, pd.DataFrame]) -> bool:
        for dependency in self.dependencies:
            if not dependency.passed:
                return True
        self.actual_value = str(workbook[self.sheet].iloc[self.column][self.row]).strip()
        self.passed = self.actual_value in self.expected_values
        return self.passed
```

Possibly my least favourite thing about the new logic is the way the `auth` dict is conditionally set on `AuthorisationCheck` instances to facilitate its access during the calculation of expected values. In the end we considered the slight hackiness of this logic a price worth paying for encapsulating checking logic within the checks, although we're open to suggestions on how we can better engineer this.

[SMD-561]: https://dluhcdigital.atlassian.net/browse/SMD-561?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


core/validation/checks.py
cyrusdobbs, line 1: This implementation stores the configuration of the `Check`s as in the previous implementation but I think we also now store the state of the current/last execution at run time. I think this means that the schema will mutate during runtime which could lead to an unclear schema state and possibly strange behaviour between ingests.. 

for example, the `error_message` attribute is overridden with the formatted version which would then cause erroneous results for subsequent uses of the `Check` (I haven't tested this but it looks as if this would happen?)

I suppose we should either: 
- instantiate a new list of `Check`s for each call to ingest because at the moment they're stored as constants
- keep the state of the current execution of a `Check` outside of the `Check` objects (i.e. `actual_value`, the formatted `error_message` and the `expected_values` for dynamic checks)

core/validation/checks.py
```diff
@@ -0,0 +1,131 @@
...
+    calculating the expected values on a cell, and substituting placeholders in error messages with the calculated
+    parameters.
+
+    Attributes:
+        _dynamic_params (dict): Dictionary containing values used to calculate the expected values dynamically.
+        _error_message_with_placeholders (str): The error message template before substitution.
+
+    """
+
+    _dynamic_params: dict  # Values used to calculate the expected values
```
cyrusdobbs, line 58: If we're creating concrete child classes for both `Conflicting` and `Authorisation` checks then we could make their `dynamic_param` dependencies instance variables as well (`mapping`, `mapped_column`, `auth_type` etc)

Not sure how we benefit from this being a dynamic dictionary

core/validation/checks.py
```diff
@@ -0,0 +1,131 @@
...
+    _error_message_with_placeholders: str  # The error message before substitution
+
+    def __init__(self, sheet: str, column: int, row: int, dynamic_params: dict, error_message_with_placeholders: str):
+        super().__init__(sheet, column, row, None, None)
+        self._dynamic_params = dynamic_params
+        self._error_message_with_placeholders = error_message_with_placeholders
+
+    @abstractmethod
+    def calculate_expected_values(self) -> tuple:
+        pass
```
cyrusdobbs, line 68: One alternative could be for this `ABC` to just contain this `abstractmethod` and then the children override `run` and init any additional instance variables they need. `AuthorisationCheck` could then implement `substitute_error_message` without sharing it with `ConflictingCheck` and we wouldn't need `_dynamic_params` either

core/validation/initial_validation/validate.py
```diff
@@ -1,163 +1,62 @@
...
-    :param schema: The schema containing wrong input checks.
-    :raises ValidationError: If wrong input validation fails.
+def initial_validate(workbook: dict[str, pd.DataFrame], schema: list[Check], auth: dict | None):
     """
-    wrong_input_checks = schema["wrong_input_checks"]
+    Executes initial checks based on the provided schema.
 
-    failures = check_values(workbook, wrong_input_checks, WrongInputFailure)
+    The submission form structures of TF Round 1 & Round 2 do not permit initial validation as they have already been
+    transformed.
```
cyrusdobbs, line 27: I think we can probably omit this context here. It could be included as a comment in the TF Round 1 and 2 `IngestDependencies` instantiation instead

tests/validation_tests/initial_validation_tests/test_pre_transformation_validate.py
cyrusdobbs, line 1: Ideally we should test this independently of any real application as it's supposed to be dynamic and generic. I think we'd want to rely on component tests to check that the system works in the context of a particular round e.g. TF Round 3

In other words, this test module shouldn't use the TF Round 3/4 schemas, we could create one for testing instead. Otherwise eventually when we discontinue TF round 3/4 then we'll have to modify this.

We could split this out into another ticket - although I think there's always a risk it's forgotten and not prioritised if we do that. Thoughts?

core/exceptions.py
```diff
@@ -3,3 +3,10 @@ class ValidationError(RuntimeError):
 
     def __init__(self, validation_failures):
         self.validation_failures = validation_failures
+
+
+class InitialValidationError(ValidationError):
```
cyrusdobbs, line 8: Do we want this to be a subclass of `ValidationError`? It would lead to odd behaviour if the two `except` statements were swapped in `ingest`, for example

cyrusdobbs (changes requested):
I really like the OOP approach to this and think the changes make a lot of sense - nice work. I like how we now bypass `process_validation_failures` which will allow us to more easily look to depreciate that flow.

My main comment is about state of the `Checks` between ingest - happy to chat about this further or hear any alternative suggestions!

Otherwise there are a few more minor comments on some of the Check classes and then also the way we current unit test these changes.

core/validation/checks.py
```diff
@@ -11,121 +11,137 @@ class Check(ABC):
...
+    def __init__(self, sheet: str, column: int, row: int, expected_values: tuple, error_message: str, auth_type: str):
+        super().__init__(sheet, column, row, expected_values, error_message)
+        self.auth_type = auth_type
 
-    def calculate_expected_values(self, workbook: dict[str, pd.DataFrame]) -> tuple:
-        auth_type = self._dynamic_params["auth_type"]
-        expected_values = self.auth[auth_type] if self.auth else ()
-        self.expected_values = expected_values
+    def set_auth_dict(self, auth_dict: dict | None):
+        self.auth_dict = auth_dict
```
cyrusdobbs, line 129: This would also mutate the Check between ingests. We could pass `auth` to `AuthorisationCheck.run` I suppose but it's unfortunate that it would give it a different signature to the other classes..

core/validation/checks.py
```diff
@@ -11,121 +11,137 @@ class Check(ABC):
...
-        if self.actual_value not in self.expected_values:
-            return False
-        return True
+    def get_actual_value(self, workbook: dict[str, pd.DataFrame]) -> str:
+        return str(workbook[self.sheet].iloc[self.column][self.row]).strip()
+
+    def run(self, workbook: dict[str, pd.DataFrame]) -> tuple[bool, str]:
+        if self.get_actual_value(workbook) not in self.expected_values:
+            return False, self.error_message
+        return True, ""
```
cyrusdobbs, line 37: We could pull `run` and `expected_values` out to `BasicCheck`? We could add `run` as an abstract method to this base class to enforce it's implementation in all sub classes.. Just a thought not a hard suggestion

core/validation/checks.py
wjrm500: Will discuss this with Bobbi tomorrow morning and try to find an elegant solution. Will also discuss the comment relating to tests above

wjrm500:
Hey @cyrusdobbs, the commit made yesterday I believe addressed most of your comments, including:

- Preventing mutation of Check instances between `/ingest` requests by not setting modified versions of `actual_value`, `expected_values` or `error_message` on those instances
- Removal of `_dynamic_params` property from `DynamicCheck` in favour of specific properties `auth_type`, `mapping`, `mapped_column` and `mapped_row` on `AuthorisationCheck` and `ConflictingCheck`
- `DynamicCheck` becoming a simple interface containing only the abstract method `get_expected_values`, with implementation of `run` delegated to `AuthorisationCheck` and `ConflictingCheck` as they have their own logic (e.g., `AuthorisationCheck` requires error message substitution to retain existing error message, while `ConflictingCheck` does not)
- Extend `InitialValidationError` from `RuntimeError` not `ValidationError` (agreed this was nonsensical)

As you suggested, there were two options for preventing mutability - create new `Check` instances on a per-request basis, or don't mutate `Check` state between requests. In the end I chose the second option, because the first option would require the schemas in `core/validation/initial_validation/schemas.py` to be lists of dictionaries instead of lists of `Check` instances, with each dictionary having a new `check_type` property with a reference to the instantiatable `Check` class, and I wasn't a fan of the idea of having these complex dictionaries whose properties needed to be mapped directly into the specific constructor of the corresponding `Check` class at runtime.

The comments you made subsequent to this commit highlighted remaining issues:
- `auth_dict` setting meant `AuthorisationCheck` instances would still be mutated
- Best to pull out `run` implementation into `BasicCheck` with abstract method in base `Check`

Myself and @bobbi-hbrown have addressed those issues this morning:
- `auth_dict` is no longer set on the `AuthorisationCheck` instance for checking in the `run` method - instead, we pass `auth` as an argument to `AuthorisationCheck.run`, and then pass that through to `AuthorisationCheck.get_expected_values`. To keep the interface of `get_expected_values` and `run` the same across classes, we use `**kwargs`.

core/validation/checks.py
```diff
@@ -0,0 +1,143 @@
...
+    elsewhere in the source code to map the values from one cell to a list of expected values for the other cell.
+
+    Attributes:
+        Inherits attributes from the DynamicCheck class.
+        mapping (dict): A dictionary mapping the values from the mapped column and row to the expected values.
+        mapped_column (int): The column index of the cell to map from.
+        mapped_row (int): The row index of the cell to map from.
+    """
+
+    mapping: dict
```
cyrusdobbs, line 73: the return type and implementation `get_expected_values` suggests that we know the type of this will be `dict[str, tuple]`

core/validation/checks.py
```diff
@@ -0,0 +1,143 @@
...
+        mapped_row: int,
+    ):
+        super().__init__(sheet, column, row, expected_values, error_message)
+        self.mapping = mapping
+        self.mapped_column = mapped_column
+        self.mapped_row = mapped_row
+
+    def get_expected_values(self, workbook: dict[str, pd.DataFrame], **kwargs) -> tuple:
+        value_to_map = str(workbook[self.sheet].iloc[self.mapped_column][self.mapped_row]).strip()
+        return self.mapping.get(value_to_map, [])
```
cyrusdobbs, line 95: return type is `tuple` but this can return a `list`.. should this return `self.mapping.get(value_to_map, tuple())`?

core/validation/checks.py
cyrusdobbs, line 1: As these `Check`s are specific to `initial-validation` we could keep this module in that sub-package

core/exceptions.py
```diff
@@ -3,3 +3,10 @@ class ValidationError(RuntimeError):
 
     def __init__(self, validation_failures):
         self.validation_failures = validation_failures
+
+
+class InitialValidationError(RuntimeError):
+    error_messages: list[str]
+
+    def __init__(self, error_messages: str):
```
cyrusdobbs, line 11: typing should be `list[str]` here

cyrusdobbs (commented):
Great work on this @wjrm500 @bobbi-hbrown. Only nitpicks/small comments left from me then will be good to approve

Best to include an update to the `IngestDependencies.initial_validation_schema` typing here as well https://github.com/communitiesuk/funding-service-design-post-award-data-store/blob/e993c74f6d2ea0d081767e2e00378426a8d9b1eb/core/controllers/ingest_dependencies.py#L47

wjrm500:
Hey @cyrusdobbs thanks for that, I've resolved all of your most recent comments. Re type hinting `get_expected_values` and `ConflictingCheck.mapping`, I've done this with `Iterable`, `set` and `list`, as per the existing code. Basically...

The `PLACE_TO_FUND_TYPE` contains sets (so `ConflictingCheck.mapping` should be `dict[str, set]` and `ConflictingCheck.get_expected_values` should return a `set`), while `auth` is of type `dict[str, list]`, so `AuthorisationCheck.get_expected_values should return a `list`. Then I used `Iterable` to type hint the return type of the `get_expected_values` in the parent class `DynamicCheck` as this covers both `set` and `list`.

cyrusdobbs (approved):
LGTM 👍

cyrusdobbs:
Please merge with squash

PR #435, opened 2024-02-07
SMD-555 - Funding Profiles B2 table extract transform validate

wjrm500:
*WIP*

### Ticket
[SMD-555](https://dluhcdigital.atlassian.net/browse/SMD-555)

### Change description
To follow...


core/extraction/towns_fund_round_four.py
```diff
@@ -92,3 +97,32 @@ def extract_programme_progress(df_data: pd.DataFrame, programme_id: str) -> pd.D
     df_data = r3.extract_programme_progress(df_data, programme_id)
     df_data = df_data.drop(df_data.iloc[5].name)  # Question 6 isn't required for Round 4
     return df_data
+
+
+def extract_programme_management(df_data: pd.DataFrame, programme_id: str) -> pd.DataFrame:
+    # if programme_id.split("-")[0] != "TD":  # Return an empty DataFrame if the programme ID is not Town Deal
+    #     return pd.DataFrame(columns=[
+    #         "Programme ID", "Payment Type", "Reporting Period", "Spend for Reporting Period"
+    #     ])
```
cyrusdobbs, line 106: Looks like this should be uncommented I think? Would be good to also have a test that checks we return an empty df with the correct headers if the fund type is not TD

core/extraction/towns_fund_round_four.py
```diff
@@ -48,6 +50,9 @@ def ingest_round_four_data_towns_fund(df_ingest: dict[str, pd.DataFrame]) -> dic
     towns_fund_extracted["Project Progress"] = r3.extract_project_progress(
         df_ingest["3 - Programme Progress"], project_lookup, round_four=True
     )
+    towns_fund_extracted["Programme Management"] = extract_programme_management(
+        df_ingest["4a - Funding Profiles"], programme_id
+    )
```
cyrusdobbs, line 55: Following our current pattern this should be defined as a round 3 extraction function to be reused here in round 4, as we want to extract this for both R3 and R4.

cyrusdobbs (changes requested):
Looking good, just need to make sure it's executed for both TF R3 and R4 👍 

I think it could be a good idea to deploy this to dev then and run a round 3 and 4 re-ingest to check that it doesn't break re-ingest for any of the existing data

wjrm500:
Thanks @cyrusdobbs - the process now happens inside the R3 pipeline as well, and there's a new test for an empty DataFrame given non-TD input. Currently the two new unit tests only sit inside `tests/extraction_tests/test_round_four_extraction.py` - wasn't sure whether to duplicate them to `test_round_three_extraction.py` or not, happy to do whatever is conventional.

tests/extraction_tests/test_round_four_extraction.py
```diff
@@ -85,6 +85,31 @@ def test_extract_programme_progress(mock_progress_sheet, mock_programme_lookup):
...
+    extracted_programme_management = tf.extract_programme_management(mock_funding_sheet, programme_lookup)
+    expected_programme_management = pd.read_csv(
+        resources_assertions / "programme_management_non_td_expected.csv", index_col=0
+    )
+    expected_programme_management["Spend for Reporting Period"] = expected_programme_management[
+        "Spend for Reporting Period"
+    ].astype(object)
+    assert_frame_equal(extracted_programme_management, expected_programme_management)
+
+
```
cyrusdobbs, line 112: Following our current convention, these tests and the assertion data should be associated with round 3 (i.e. `test_round_three_extraction.py` and the expected data to be in `resources/assertion_data/round_three`)

cyrusdobbs: but don't need to be duplicated across both

cyrusdobbs (approved):
LGTM - before you merge this I think it would be good to deploy it to test and re-ingest round 3 files as a test

PR #439, opened 2024-02-09
BAU - Remove ambiguity around use of the term "workbook" for transformed data

wjrm500:
### Change description
In current code, we use the term "workbook" and "worksheet" post-transformation (transformation from the raw Excel workbook data into the data in the format of our data model, ready to be ingested into the database) to refer to the dictionary of transformed dataframes and the individual transformed dataframes respectively, which is misleading as "workbook" and "worksheet" are tied to Excel and Excel is not relevant post-transformation. This PR aims to resolve that.

### DIFF
--- a/core/controllers/ingest.py
+++ b/core/controllers/ingest.py
@@ -8,7 +8,6 @@ from json import JSONDecodeError
 from typing import BinaryIO
 from zipfile import BadZipFile
 
-import numpy as np
 import pandas as pd
 from flask import abort, current_app, g
 from sqlalchemy import exc
@@ -78,14 +77,14 @@ def ingest(body: dict, excel_file: FileStorage) -> tuple[dict, int]:
 
     ingest_dependencies: IngestDependencies = ingest_dependencies_factory(fund, reporting_round)
 
-    original_workbook = extract_data(excel_file=excel_file)
+    sheet_data = extract_data(excel_file)
     try:
         if iv_schema := ingest_dependencies.initial_validation_schema:
-            initial_validate(original_workbook, iv_schema, auth)
-        data_dict = ingest_dependencies.transform_data(original_workbook)
+            initial_validate(sheet_data, iv_schema, auth)
+        transformed_data = ingest_dependencies.transform_data(sheet_data)
         validate(
-            data_dict,
-            original_workbook,
+            transformed_data,
+            sheet_data,
             ingest_dependencies.validation_schema,
             ingest_dependencies.fund_specific_validation,
         )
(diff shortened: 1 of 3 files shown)


core/controllers/ingest.py
```diff
@@ -245,47 +244,46 @@ def extract_data(excel_file: FileStorage) -> dict[str, pd.DataFrame]:
...
     :param data: A dictionary where the keys are table names and the values are pd.DataFrames
     :return: None
     """
-    for table in data.values():
-        table.fillna("", inplace=True)  # broad replace of np.NA with empty string
-        table.replace({np.nan: None}, inplace=True)  # replaces np.NAT with None
+    for table in transformed_data.values():
+        table.fillna("", inplace=True)
+        # fillna with empty string does not work for datetime columns, so we need to replace NaT with None
+        table.replace({pd.NaT: None}, inplace=True)
```
Joel-Pearce, line 257: Can we add a unit test for ```clean_data``` with an example of pd.NaT in? Can we also make sure this test has an example of np.nan in to make sure that functionality still works (if such a test doesn't already exist)?

core/controllers/ingest.py
```diff
@@ -78,14 +77,14 @@ def ingest(body: dict, excel_file: FileStorage) -> tuple[dict, int]:
 
     ingest_dependencies: IngestDependencies = ingest_dependencies_factory(fund, reporting_round)
 
-    original_workbook = extract_data(excel_file=excel_file)
+    sheet_data = extract_data(excel_file)
```
Joel-Pearce, line 80: This might make more sense as ```pre_transformed_data```, or ```raw_data```. I feel like ```sheet_data``` is less clear than ```original_workbook```. Maybe ```untransformed_sheet_data``` if we want to keep the reference to sheets.

Joel-Pearce (changes requested):

core/controllers/ingest.py
wjrm500: Have changed to `workbook_data` as feel this best represents the object

Joel-Pearce (approved):
I'm going to approve the PR as everything looks good but I have a couple of nitpicks I'll write here:

1) I think we can dispense with boilerplate comments like ```# Call the function under test``` if it's obvious what is happening.

2) It's probably best to be consistent with test assertions. For the ```np.nan``` values you're asserting whether ```.isna().any()``` over the entire Series, but for ```np.NaT``` values you're retrieving a specific index and checking whether it's ```None```. It can be annoying to write assertions on Series and DataFrames because they're not truthy but you can use [assert_series_equals](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.testing.assert_series_equal.html) and [assert_dataframe_equals](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.testing.assert_frame_equal.html) to check whether the Series or DataFrame in its entirety is what you expect after a function call.

wjrm500:
Thanks @Joel-Pearce, have removed the comments and switched to using `assert_frame_equal`. Will merge now

PR #443, opened 2024-02-13
SMD-560 pf initial validation schema and tests

bobbi-hbrown:
https://dluhcdigital.atlassian.net/jira/software/c/projects/SMD/boards/140?quickFilter=815&selectedIssue=SMD-560

I removed the PF mock spreadsheets and validation tests (these are on a previous branch), and replaced them with integration tests. With regards to reporting round, the current spreadsheet is treated as round 1, though this may need to be adjusted to account for historical data. If a user selects a higher round (2, 3 and 4 are available in the UI) this will cause an unhandled exception - so we should discuss whether to keep this behavior or add specific error handling.

An if statement checks during ingest if the fund is Pathfinders, and if so does not carry out full validation, as this is yet to be implemented.

Manual testing requires an updated template spreadsheet with an added Metadata tab containing reporting round, version and fund name.


core/validation/initial_validation/schemas.py
```diff
@@ -105,3 +105,33 @@
         '"Future_High_Street_Fund".',
     ),
 ]
+
+PF_INITIAL_VAL_SCHEMA = [
```
cyrusdobbs, line 109: bit of a nitpick but would be good to maintain the same naming convention across these constants

core/validation/initial_validation/schemas.py
```diff
@@ -105,3 +105,33 @@
...
+    ),
+    AuthorisationCheck(
+        sheet="Metadata",
+        column=0,
+        row=2,
+        expected_values=(),
+        error_message="You’re not authorised to submit for {wrong_place_or_fund_type}. You can only submit for "
+        "{allowed_places_or_fund_types}.",
+        auth_type="Fund Types",
+    ),
```
cyrusdobbs, line 136: I don't think an `AuthorisationCheck` on fund type is relevant for PF because unlike the TF templates where the same template can either be FHSF or TD, PF is just PF. This means we can just use a `BasicCheck` to check this field instead as the expected value does not vary.

core/validation/initial_validation/schemas.py
```diff
@@ -105,3 +105,33 @@
...
+
+PF_INITIAL_VAL_SCHEMA = [
+    BasicCheck(
+        sheet="Metadata",
+        column=0,
+        row=0,
+        expected_values=("1",),
+        error_message="The expected reporting period is Q3 Oct - Dec 23/24",
+    ),
+    BasicCheck(sheet="Metadata", column=0, row=1, expected_values=("4",), error_message="The expected value is V 4.0"),
```
cyrusdobbs, line 117: I think it makes more sense to begin the versioning at 1

core/validation/initial_validation/schemas.py
```diff
@@ -105,3 +105,33 @@
...
+    # should this be renamed to organisation name to reflect the spreadsheet?
+    AuthorisationCheck(
+        sheet="Admin",
+        column=11,
+        row=1,
+        expected_values=(),
+        error_message="You’re not authorised to submit for {wrong_place_or_fund_type}. You can only "
+        "submit for {allowed_places_or_fund_types}.",
```
cyrusdobbs, line 125: I think we could rename the variable `allowed_places_or_fund_types`/`wrong_place_or_fund_type` to be more generic like `expected`/`actual`

core/validation/initial_validation/schemas.py
```diff
@@ -105,3 +105,33 @@
...
+        auth_type="Place Names",
```
cyrusdobbs, line 126: Agree with your comment above that `Place Names` isn't relevant. I think we should look to make this more generic as well as both `Place Name` and `Organisation` are both fund specific terms. I think `Programme` would be best, as it's internal facing.

tests/integration_tests/test_ingest_component.py
```diff
@@ -99,6 +99,145 @@ def wrong_format_test_file() -> BinaryIO:
...
 
 
+@pytest.fixture(scope="function")
+def pathfinders_round_1_file_success() -> BinaryIO:
+    """An example spreadsheet for reporting round 4 of Towns Fund that should ingest without validation errors."""
+    with open(Path(__file__).parent / "mock_pf_returns" / "PF_Round_1_Success.xlsx", "rb") as file:
+        yield file
+
+
+@pytest.fixture(scope="function")
```
cyrusdobbs, line 109: I think we can make both of these fixtures `module` scope, as we'll only be reading the return value anyway

tests/integration_tests/test_ingest_component.py
cyrusdobbs, line 1: These tests look good. Only thought is that it would be more organised to pull PF integration tests out into a separate test module otherwise this may become a bit unwieldy as we add more tests I think

core/controllers/ingest.py
```diff
@@ -81,6 +81,10 @@ def ingest(body: dict, excel_file: FileStorage) -> tuple[dict, int]:
     try:
         if iv_schema := ingest_dependencies.initial_validation_schema:
             initial_validate(workbook_data, iv_schema, auth)
+
+        if fund == "Pathfinders":
+            return {"detail": "PF initial validation success"}, 200
```
cyrusdobbs, line 86: I think a comment explaining that this is a temporary stopgap until more of the Pathfinders pipeline is developed could be useful

cyrusdobbs (changes requested):
Looks good - mostly just slight tweaks to the initial validation schema 👍

tests/integration_tests/test_ingest_component.py
bobbi-hbrown: I noticed I'm getting intermittent 'read of closed file' errors during testing with these example spreadsheets when scop is set to moduke, and it varies between each test run whether it succeeds or not. It seems to pass consistently when set to function scope - I'm not sure why this is happening though

core/extraction/towns_fund_round_one.py
cyrusdobbs, line 1: Can we revert these? Not sure why there's changes in these files - I guess because of a find all and replace?

core/validation/initial_validation/schemas.py
```diff
@@ -63,18 +63,18 @@
...
         sheet="2 - Project Admin",
         column=6,
         row=4,
         expected_values=(),
-        error_message="You’re not authorised to submit for {wrong_place_or_fund_type}. You can only "
-        "submit for {allowed_places_or_fund_types}.",
-        auth_type="Place Names",
+        error_message="You’re not authorised to submit for {entered_value}. You can only "
+        "submit for {allowed_values}.",
+        auth_type="Programme",
```
cyrusdobbs, line 77: We need to keep the key for TFs programe auth as `Place Names` because otherwise integration with the front-end will break atm. I think this should be changed in a later PR, after a change to the front-end to allow for this to not be a breaking change.

tests/integration_tests/test_pathfinders_ingest_validation.py
cyrusdobbs, line 1: I think file naming should be consistent across both of these files. I also think as this is testing the endpoint as a whole we should keep the term validation out of the naming. We could call this `test_ingest_component_pathfinders` and rename `test_ingest_component` to `test_ingest_component_towns_fund`?

openapi/api.yml
```diff
@@ -121,7 +121,7 @@ paths:
                   description: JSON string; restrict ingest to submissions based on some fields in the data file. If
                     this is omitted then authorisation checks will not be run.
                     e.g. Towns Fund checks against Place Name and Fund Type.
-                  default: "{\"Place Names\": [\"Blackfriars - Northern City Centre\"], \"Fund Types\": [\"Town_Deal\", \"Future_High_Street_Fund\"]}"
+                  default: "{\"Programme\": [\"Blackfriars - Northern City Centre\"], \"Fund Types\": [\"Town_Deal\", \"Future_High_Street_Fund\"]}"
```
cyrusdobbs, line 124: It might be misleading if we change this before modifying TF's auth keys - probs best to revert these for now as well

cyrusdobbs (changes requested):
Main thing is that we need to keep the key for TFs programme auth as `Place Names` for now because otherwise integration with the front-end will break. It's built in a way that it can be dynamic per fund, so PF can be `Programmes` and TF can use `Place Name` for now. Although we'll probably want to align these going forwards, but in a way that won't break integration

core/validation/initial_validation/schemas.py
```diff
@@ -105,3 +105,31 @@
         '"Future_High_Street_Fund".',
     ),
 ]
+# TODO replace the placeholder error messages with those from design
+PF_ROUND_1_INIT_VAL_SCHEMA = [
+    BasicCheck(
+        sheet="Metadata",
+        column=0,
+        row=0,
+        expected_values=("1",),
```
cyrusdobbs, line 114: Do we know what the actual reporting period will be?

bobbi-hbrown: There is a row in the Start tab which has a select dropdown with all the reporting periods - for example Q3 Oct - Dec 23/24. We can use this to validate against instead of the reporting round?

tests/integration_tests/test_ingest_component.py
cyrusdobbs: The pattern I suggested doesn't work with `module` scope fixtures because the file closes after one test has finished, which causes subsequent tests to fail. Resolving this comment.

core/validation/initial_validation/schemas.py
cyrusdobbs: My bad, mixed up reporting period and round - this is fine for now

openapi/api.yml
```diff
@@ -166,7 +166,7 @@ paths:
                 example:
                   detail: "Spreadsheet successfully validated and ingested"
                   metadata:
-                    programme: "Blackfriars - Northern City Centre"
```
cyrusdobbs, line 169: this should be reverted

cyrusdobbs (commented):
just one thing to revert and then happy to approve

tests/integration_tests/test_ingest_component_pathfinders.py
```diff
@@ -0,0 +1,140 @@
...
+                        "Fund Types": [
+                            "Pathfinders",
+                        ],
+                    }
+                ),
+                "do_load": False,
+            },
+        )
+
+    assert e.value.args == ("There are no IngestDependencies for Pathfinders round 2",)
```
cyrusdobbs, line 140: Can we assert against a 500 response here instead of catching an `Exception` as these tests should be asserting against the response rather than anything at the Python level

core/validation/initial_validation/schemas.py
```diff
@@ -106,3 +106,30 @@
         '"Future_High_Street_Fund".',
     ),
 ]
+# TODO replace the placeholder error messages with those from design
```
nsnape, line 109: are there Jira tickets tracking these TODO items?

cyrusdobbs: There is a ticket that Kate is owning at the moment to write up all the messages and that will eventually get passed to us to implement - so yes for this

tests/integration_tests/test_ingest_component_pathfinders.py
cyrusdobbs: Thinking about this more.. I don't think it makes sense to return a 500 error here. We should be returning a 400 (bad request) response instead. 

It would be best to change `ingest_dependencies_factory` to return `None` in the case that fund and reporting round are invalid here: 
https://github.com/communitiesuk/funding-service-design-post-award-data-store/blob/682cce7c645b2fa6cd5118f819da4b4976daaded/core/controllers/ingest_dependencies.py#L88-L89

and then `controllers.ingest.ingest` can handle the `None` by returning a 400 response here:
https://github.com/communitiesuk/funding-service-design-post-award-data-store/blob/682cce7c645b2fa6cd5118f819da4b4976daaded/core/controllers/ingest.py#L78

cyrusdobbs (approved):
LGTM

PR #450, opened 2024-02-16
[#FMD-116]: Event data to JSONB fields

Joel-Pearce:
As part of our quest to make event data be more dynamic for other funds, we have decided to move as much event data as possible into a json blob field. 

This PR:
- Migrates all of the data in the relevant fields in the relevant tables to a json_blob column called ```event_data_blob```, and subsequently drops the original columns.
- Removed ```test_historical_ingest.py``` as it created conflicts with the code changes herein and is to be deprecated anyway.
- Changes ingest and serialisation such that incoming data is ingested into ```event_data_blob``` where appropriate but all data serialised out into download files remains the same from the perspective of the user.

Ticket: https://dluhcdigital.atlassian.net/browse/FMD-116


core/util.py
```diff
@@ -157,3 +175,30 @@ def custom_serialiser(obj: Any) -> str:
...
+    new_cols = list(set(data_columns).intersection(cols_to_jsonb))
+    df_with_cols_to_move = data[new_cols]
+    json_blob_col = []
+
+    for row in df_with_cols_to_move.iterrows():
+        json_blob_row = {}
+        # row[0] is the index for the row in the DataFrame, which can be discarded
+        for col_name, col_values in row[1].iteritems():
+            json_blob_row[col_name] = col_values
+        json_blob_col.append(json.dumps(json_blob_row))
```
cyrusdobbs, line 199: could simplify this to
```
json_blob = [
    json.dumps(row._asdict())
    for row in df_with_cols_to_move.itertuples(index=False)
]
```

db/migrations/versions/011_event_data_to_jsonb.py
```diff
@@ -0,0 +1,169 @@
+"""empty message
```
nsnape, line 1: there's a migration in feature production by Marc, so check the number/id on this before committing

PR #452, opened 2024-02-16
SMD-618 - Mock extracted validated data

wjrm500:
WIP - based off [SMD-553 branch](https://github.com/communitiesuk/funding-service-design-post-award-data-store/tree/SMD-553_reorganise-extraction) which has not yet been merged into main.

### Change description
Dataframes outputted from [scripts/pathfinders_validation.py](https://github.com/communitiesuk/funding-service-design-post-award-data-store/pull/432/files#diff-b29718a107d62113221c900f7ef117f2bc5f08c8e841d971c73a78f61bb78668) created, along with programme -> project ID -> project name mapping, currently sat in the tests folder and ready to be used during development of Pathfinders transformation function.


tests/transformation_tests/pathfinders/resources/extracted_validated_data.py
```diff
@@ -0,0 +1,180 @@
...
+        {
+            "ID": "PF-BOL-008",
+            "Project Name": "Streets for All - Farnworth",
+        },
+        {
+            "ID": "PF-BOL-009",
+            "Project Name": "Bolton College of Medical Science",
+        },
+    ]
+}
```
cyrusdobbs, line 180: This may be most useful its just a dictionary mapping project name to ID

tests/transformation_tests/pathfinders/resources/extracted_validated_data.py
```diff
@@ -0,0 +1,180 @@
+import numpy as np
+import pandas as pd
+
+financial_completion_date = pd.DataFrame({"Pathfinder Financial Completion Date": [pd.Timestamp("2001-01-01")]})
```
cyrusdobbs, line 4: With the current flow all data will be `str` type as we don't do any type "coercion"

cyrusdobbs: Perhaps we should still add types to the `pa.Column` definitions and coerce the data.. we could run the fund specific validation on the non-coerced data so that we know we know we won't be missing any un-coercible data. Then, if all checks pass then we can return the coerced data to transformation. I presume it would be useful to have the data correctly typed for some transformations (particularly for `datetime`s?)

wjrm500:
Superseded by [SMD-532 - Pathfinders transformation pipeline](https://github.com/communitiesuk/funding-service-design-post-award-data-store/pull/464)

PR #453, opened 2024-02-20
FMD-227: Remove all submission files save/retrieve in db logic

MarcUsher:
### Change description
As part of FMD-227, this PR removes all logic for saving and retrieving submission files to/from the database, including redundant endpoint and unneeded script. Preserving saving of excel filename to db in its own function. 

DB model update and migration will be done in a separate PR.

### How to test


core/controllers/ingest.py
```diff
@@ -371,17 +367,14 @@ def populate_db_historical_data(transformed_data: dict[str, pd.DataFrame], mappi
...
-# TODO: [FMD-227] Remove submission files from db
-def save_submission_file_db(excel_file: FileStorage, submission_id: str):
-    """Saves the submission Excel file.
+def save_submission_file_name(excel_file: FileStorage, submission_id: str):
+    """Saves the submission Excel filename.
 
     :param excel_file: The Excel file to save.
     :param submission_id: The ID of the submission to be updated.
     """
     submission = Submission.query.filter_by(submission_id=submission_id).first()
```
gidsg, line 376: Could we do a Get here on the ID rather than filter? i.e. as per: https://docs.sqlalchemy.org/en/14/orm/session_api.html#sqlalchemy.orm.Session.get ?

MarcUsher: @gids Actually not sure if that would work in this instance - it looks like the Get requires you to use the primary key which we don't have at this point (only have the `submission_id`), so would need an extra query to get that which seems redundant as at that point we'd have identified the submission already?

gidsg (approved):
LGTM 👍

gidsg (approved):

PR #456, opened 2024-02-20
[#FMD-116]: event data to json blob col

Joel-Pearce:
As part of our quest to make event data be more dynamic for other funds, we have decided to move as much event data as possible into a json blob field.

This PR:

- Migrates all of the data in the relevant fields in the relevant tables to a json_blob column called ```event_data_blob```, and subsequently drops the original columns.
- Removed ```test_historical_ingest.py``` as it created conflicts with the code changes herein and is to be deprecated anyway.
- Changes ingest and serialisation such that incoming data is ingested into ```event_data_blob``` where appropriate but all data serialised out into download files remains the same from the perspective of the user.

### How to test


tests/integration_tests/test_ingest_component.py
```diff
@@ -908,14 +908,6 @@ def test_ingest_same_programme_different_rounds(
...
 
     assert len(Project.query.filter(Project.project_id == "HS-WRC-01").all()) == 2
-    assert (
-        Project.query.filter(Project.project_id == "HS-WRC-01").all()[0].programme_junction.submission.reporting_round
-        == 3
-    )
-    assert (
-        Project.query.filter(Project.project_id == "HS-WRC-01").all()[1].programme_junction.submission.reporting_round
-        == 4
-    )
```
Joel-Pearce, line 918: To pre-empt questions of why I removed this: this particular piece of code is not relevant to the actual purpose of the test, and is poorly written as there's no guarantee the 3rd round's project will always be above the 4th round.

core/util.py
```diff
@@ -129,6 +131,100 @@ def load_example_data():
         table_df = pd.read_csv(resources / f"{table}.csv")
         if table == "project_dim":
             table_df["postcodes"] = table_df["postcodes"].str.split(",")
+        if table == "project_progress":
```
nsnape, line 134: could this be a good contender for using switch statement match-case ?

core/util.py
```diff
@@ -129,6 +131,100 @@ def load_example_data():
...
+            table_df = table_df.replace(np.nan, None)
```
nsnape, line 135: this line is repeated just prior to every time the `move_event_data_to_json_blob `function is called, so can it just be moved insidethe function to avoid this duplication?

core/util.py
```diff
@@ -157,3 +253,23 @@ def custom_serialiser(obj: Any) -> str:
         return obj.isoformat()
     else:
         raise TypeError(f"Cannot serialise object of type {type(obj)}")
+
+
+def move_event_data_to_json_blob(
```
nsnape, line 258: this could have a more generic name? It doesn't need to be specific to "event data" since it actually just moves any dataframe columns into a dict? And something outside this func dumps into the JSON? Probs need to update the signature and the docstring to reflect this, and maybe explain the structure of the json/dict (ie columns/values moved to be key/value pairs in the new dict column)?

nsnape (changes requested):

core/util.py
Joel-Pearce: Moved inside which function?

Joel-Pearce: The issue is it's better to have the ```cols_to_jsonb``` instantiated inside the mapping for the sake of clarity and consistency, but we can't call the mapping here because it creates a circular import if we import it into ```util```.

Joel-Pearce: Although perhaps this function can go in ```conftest``` as that seems somewhat more appropriate.

Joel-Pearce: But that would also create a circular import...

nsnape: to first line of `move_data_to_json_blov`

nsnape: I'm specifically talking about the line 
`table_df = table_df.replace(np.nan, None)`

Joel-Pearce: Ah ok, got it

core/util.py
```diff
@@ -157,3 +224,29 @@ def custom_serialiser(obj: Any) -> str:
         return obj.isoformat()
     else:
         raise TypeError(f"Cannot serialise object of type {type(obj)}")
+
+
+def move_data_to_json_blob(
```
nsnape, line 229: we should probably be careful with naming conventions here. JSONB does not actually stand for JSON Blob (it's actually JSON Binary), but we are using it as "blob storage".

Probably a better name would be jsonb_blob, just to be clear?

core/util.py
```diff
@@ -157,3 +224,29 @@ def custom_serialiser(obj: Any) -> str:
...
+    data: pd.DataFrame,
+    cols_to_jsonb: list[str],
+) -> pd.DataFrame:
+    """Move specified columns into a json_blob field.
+
+    This function moves specified columns in a DataFrame into a json_blob field.
```
nsnape, line 235: so can probably replace instances of "json_blob" in this doctring, as the snake case there kind of implies that "json_blob" is some kind of reference. In the docstring itself we should probably refer to is as JSONB Blob, or JSON-Binary blob storage. Something like that

nsnape (changes requested):

core/util.py
```diff
@@ -157,3 +224,29 @@ def custom_serialiser(obj: Any) -> str:
...
+def move_data_to_jsonb_blob(
+    data: pd.DataFrame,
+    cols_to_jsonb: list[str],
+) -> pd.DataFrame:
+    """Move specified columns into a JSONB blob field.
+
+    This function moves specified columns in a DataFrame into a JSONB blob field.
+    This field is a dictionary, with the keys as the column names and the values
+    as the values in each row of the columns.
+    Afterward, the columns from which the data has been moved into a json_blob are dropped.
```
nsnape, line 238: json_blob

core/util.py
```diff
@@ -157,3 +224,29 @@ def custom_serialiser(obj: Any) -> str:
...
+
+    :param data: data for a given table
+    :param cols_to_jsonb: columns to be moved into a json blob
```
nsnape, line 241: json_blob

core/util.py
```diff
@@ -157,3 +224,29 @@ def custom_serialiser(obj: Any) -> str:
...
+    :return: a DataFrame with specified columns moved into a JSONB blob
+    """
+    data_columns = data.columns.tolist()
+    new_cols = list(set(data_columns).intersection(cols_to_jsonb))
+    df_with_cols_to_move = data[new_cols]
+    json_blob_col = [row._asdict() for row in df_with_cols_to_move.itertuples(index=False)]
```
nsnape, line 247: json_blob

nsnape (changes requested):

PR #462, opened 2024-02-26
[#FMD-116]: make Funding table Project & Programme level

Joel-Pearce:
Adds a ForeignKey for ```Programme_Junction_ID``` to the Funding table.


core/db/queries.py
```diff
@@ -107,16 +107,34 @@ def funding_query(base_query: Query) -> Query:
     :param base_query: SQLAlchemy Query of core tables with filters applied.
     :return: updated query.
     """
+    conditional_expression_project_id = case((ents.Funding.project_id.is_(None), None), else_=ents.Project.project_id)
```
nsnape, line 110: as discussed, please adda bit to the docstring about these conditionals

nsnape (changes requested):

core/controllers/mappings.py
```diff
@@ -68,11 +68,13 @@ def map_data_to_models(self, data: pd.DataFrame) -> list[db.Model]:
                     parent_lookup_1, parent_lookup_2 = parent_lookup
                     child_lookup_1, child_lookup_2 = child_lookup_column
                     lookups = {
-                        parent_lookup_1: self.get_row_id(ents.Programme, {parent_lookup_1: row[child_lookup_1]}),
-                        parent_lookup_2: self.get_row_id(ents.Submission, {parent_lookup_2: row[child_lookup_2]}),
+                        parent_lookup_1: self.get_row_id(ents.Programme, {parent_lookup_1: row.get(child_lookup_1)}),
```
nsnape, line 71: I know it's not strictly speaking part of this PR, but could we rename anything post-foxed with a _1 or _2 to be more meaningful? I have literally no idea from this code alone what they are :)

db/migrations/versions/013_all_event_tables_prog_j_id.py
```diff
@@ -0,0 +1,44 @@
...
+            ["programme_junction_id"],
+            ["id"],
+            ondelete="CASCADE",
+        )
+
+    # ### end Alembic commands ###
+
+
+def downgrade():
+    with op.batch_alter_table("funding", schema=None) as batch_op:
```
nsnape, line 38: we're in a weird situation here, whereby this downgrade script will work ok until we add any programme-level data, then it will crash because you won't be able to make the project_id column non-nullable again until you delete out the rows that have no project_id.

It's very unlikely that we would downgrade once we've started ingesting data, so it's probably not an issue. But 1 thing we could do is DROP the rows that would break it before changing it back to non-nullable - It's not ideal as that goes against what a data migration should do (preserve data), so I guess I'm leaning towards leaving it and letting the programmer that reverts the code deal with the implications of dropping those rows (rather than it being automatic)

nsnape: i.e. deal with the ensuing DB error traceback

nsnape (changes requested):

db/migrations/versions/013_all_event_tables_prog_j_id.py
Joel-Pearce: I contend we should leave as-is or else someone in the future, unaware of the minutiae of each migration script, might rollback pre-013 for whatever reason and inadvertently wipe data without knowing about it, and never subsequently re-ingest it. At least this way the error will prompt an understanding of what is happening, and enable future contributors to decide the best course of action.

I'll leave a comment in the migration script to make this clear.

nsnape (approved):

PR #464, opened 2024-02-26
SMD-532 - Pathfinders transformation pipeline

wjrm500:
### Ticket
https://dluhcdigital.atlassian.net/browse/SMD-532

### Description
Implementing transformation between extracted validated data coming from new Pathfinders Reporting Template spreadsheet to internal data model ready for ingestion into database.


wjrm500:
Hey @nsnape and @cyrusdobbs, would be good to get your eyes on this...

The idea behind the work on this PR is to perfectly recreate the structure of TF transformed DataFrames, with the aim of minimising problems on integration. This means for the time being keeping all fields, even those that aren't relevant to Pathfinders - in these cases we fill the values as `pd.NA`. In cases where data is now programme-level not project-level (e.g., "output_data", "funding"), we also return "Programme ID" in the transformed DF and add an `fk_relation` in `mappings.py` to allow `programme_junction_id` to be populated. We also have a new table "Project Finance Changes" to support.

Regarding how this work syncs up with upstream work (extraction, validation, mapping retrieval), the transformation function has as dependencies the extracted, validated data (a dictionary of DataFrames, each representing an extracted, validated table from the spreadsheet e.g., "Organisation Name") and mappings (e.g., programme name -> programme ID). Currently I understand that the extraction and validation process returns a dictionary of **lists** of DataFrames, owing to the support for extraction of multiple tables for a single tag (this may be outdated information), so there is a mismatch here that warrants discussion. Given that we want to continue to support multiple table extraction, a couple of options might be:

- Adjusting the extraction and validation function (assuming this is encapsulated into a single function) to return single DataFrames e.g., by concatenating tables together after processing (with `TableProcessor`) but before validation (with `TableValidator`), or after validation but before returning from the function
- Adjusting the transformation function in this PR to concatenate tables together at the beginning, before any transformation takes place

Either way I think it definitely makes sense to be working with single DataFrames during transformation, not lists of DataFrames.

Regarding how this work syncs up with downstream work (database changes), the code in this PR presupposes the upcoming database change by Joel to [add programme_junction_id to output_data and funding tables](https://github.com/communitiesuk/funding-service-design-post-award-data-store/pull/462) as well as the upcoming [addition of the project finance change table](https://dluhcdigital.atlassian.net/browse/FMD-234). The docstrings in this PR also presuppose the upcoming change to [standardise the outputs and outcomes tables](https://dluhcdigital.atlassian.net/browse/FMD-229). One thing the code in this PR does yet presuppose is the potential change to the "Submission" table (addition of `event_data_blob` field) which would support ingestion of "Sign-off" data.

TODOs:
- Mapping retrieval and exposure (ticket here: https://dluhcdigital.atlassian.net/browse/SMD-652)
- Finalise database / entities / mappings changes (see above ticket links) and ensure transformation logic and documentation aligns
- Integrate transformation logic into `/ingest` endpoint, which as I understand is an ongoing piece of work already begun by Cyrus (unsure if ticket already exists?)
- Adjust data serialisation to display new project finance changes table. Other changes to data serialisation may also be necessary although my understanding is that not too much will need to be changed here owing to the tight mapping that we've done from PF to TF fields, the fact that significant differences in the data we collect in the "Admin" tab are abstracted away behind question and answer columns, etc. One issue is that we're going to have lots of empty columns in the data download, but this can be considered later (unsure if ticket already exists?)

Potential TODOs further down the road:
- Explore returning Pydantic models not DataFrames from transformation for better code design and validation on transformed data

core/transformation/pathfinders/version_1.py
```diff
@@ -0,0 +1,491 @@
...
+                                      transformed DF "Outcome_Data"
+            start_date              - from "Start_Date" in the transformed DF "Outcome_Data"
+            end_date                - from "End_Date" in the transformed DF "Outcome_Data"
+            event_data_blob         - includes "Unit of Measurement", "Amount" and "Actual/Forecast" from the
+                                      transformed DF "Outcome_Data"
+    """
+    organisation_name = df_dict["Organisation Name"].iloc[0, 0]
+    programme_id = programme_name_to_id_mapping[organisation_name]
+    outcomes = df_dict["Outcomes"]["Outcome"]
+    outcome_categories = outcomes.map(OUTCOME_CATEGORIES)
```
cyrusdobbs, line 403: I don't think we'll be able to use these mappings because they don't include all the values from the PF template. We need to confirm mappings with PF as mentioned here https://dluhcdigital.atlassian.net/browse/SMD-655.

I think it will probably be best to store them in the template, that way we can still display the IT to the user without letting them select it. I think we should also mock out those mappings for now as well, instead of using the existing ones used for TF. 

Once confirmed and added to the spreadsheet we can also update the work done in https://github.com/communitiesuk/funding-service-design-post-award-data-store/pull/468 to include these mappings. What do you think?

core/transformation/pathfinders/version_1.py
```diff
@@ -0,0 +1,491 @@
...
+            "Consequences": [pd.NA] * len(risks),
+            "Pre-mitigatedImpact": risks["Impact score"],
+            "Pre-mitigatedLikelihood": risks["Likelihood score"],
+            "Mitigations": risks["Mitigations"],
+            "PostMitigatedImpact": [pd.NA] * len(risks),
+            "PostMitigatedLikelihood": [pd.NA] * len(risks),
+            "Proximity": [pd.NA] * len(risks),
+            "RiskOwnerRole": [pd.NA] * len(risks),
+        }
+    )
```
cyrusdobbs, line 472: As an extension to my point above about hardcoding empty columns.. This seems even less necessary, as all these fields are to be stored in a JSONB field. I think we should let the serialisation handle missing values in the `/download` extract - what do you think?

cyrusdobbs (commented):
That is some clean transformation code, nice work.

My main point is around whether we want to return empty columns for all TF data points that are not included in PF, or if we want to just let the db models and serialisation handle missing data. I imagine @nsnape may have thoughts on this as well.

A good check against the output of the transformation will be to try and load it into the db. I would have said that it'd be good to set up a test as part of this PR that does that so we know the data will go in, but I think there a fair few other changes that will need to be made for that to work using the existing load functionality. So to avoid scope creep and a huge PR its probably best to avoid and just get this in first..

Other than that, thoughts on: 

- the outstanding PF outcome/output mapping which we spoke a lot about on the two away days which still needs to be cleared up with PF (see https://dluhcdigital.atlassian.net/browse/SMD-655) but for now we could mock that out as we have with the other mappings
- extending the test data so it covers a list of postcodes being entered in Project Location
- moving the new constants in `core/const.py` to somewhere more specific as it's not shared used over the whole codebase and is PF specific

db/migrations/versions/013_add_programme_junction_to_funding_and_output_data_tables.py
```diff
@@ -0,0 +1,40 @@
...
+# revision identifiers, used by Alembic.
+revision = "013_add_programme_junction_to_tables"
+down_revision = "012_event_data_to_jsonb"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column("output_data", sa.Column("programme_junction_id", sa.GUID(), nullable=False))
```
nsnape, line 21: also, as a side note - this line would never work on a pre-populated table, since you can never just add a non-nullable table (as all existing lines would instantly break this constraint). The usual work around is to update this script to first of all make the column nullable, then migrate in the correct data, then update the column to be non-nullable. All in one atomic transaction.

Same goes for the downgrade, you need to reverse the data migration done above or the DB will not downgrade (as running downgrade will hit errors and rollback)

tests/transformation_tests/pathfinders/resources/extracted_validated_data.py
```diff
@@ -0,0 +1,166 @@
+import pandas as pd
+
+financial_completion_date = pd.DataFrame({"Pathfinder financial completion date": [pd.Timestamp("2001-01-01")]})
```
nsnape, line 3: These are "globally" or at least module level variables and as such should be named in CAPS_CASE

nsnape (dismissed):

core/transformation/pathfinders/version_1.py
```diff
@@ -0,0 +1,491 @@
...
+    project_ids = df_dict["Project Location"]["Project name"].map(project_name_to_id_mapping)
+    location_multiplicities = df_dict["Project Location"]["Location"].map(
+        lambda x: "Multiple" if "," in x else "Single"
+    )
+    locations = df_dict["Project Location"]["Location"].map(lambda x: x.split(","))
+    postcodes = df_dict["Project Location"]["Location"].map(extract_postcodes)
+    return pd.DataFrame(
+        {
+            "Project Name": df_dict["Project Location"]["Project name"],
+            "Primary Intervention Theme": [pd.NA] * len(project_ids),
```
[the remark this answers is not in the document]
cyrusdobbs: ah didn't see those changes because this branch doesn't have them

wjrm500:
Hey @cyrusdobbs and @nsnape thanks for your feedback.

In light of your review, I have made changes this morning including:
- Removed NA columns from transformation output in cases where those columns just end up shoved into the event_data_blob table anyway
- Added output_intervention_theme_mapping and outcome_intervention_theme_mapping as dependencies to the transformation function in a similar fashion to programme_name_to_id_mapping, with idea being that existing OUTPUT_CATEGORIES and OUTCOME_CATEGORIES consts are not valid for PF and mappings may be extracted from spreadsheet
- Appended a new data point to project location test data containing multiple postcodes to enhance test
- Moved the PF-specific transformation constants out of `core/const.py` and into `core/transformation/pathfinders/consts.py`
- Removed duplicate migration and changes to mappings and entities around making `funding` and `output_data` tables project and programme-level, as this was covered in a separate PR by Joel that has now been merged into this one
- Renamed `core/transformation/pathfinders/version_1.py` to `core/transformation/pathfinders/transform.py`
- Created a Jira ticket link for the "Sign off" data TODO and added to comment
- Removed underscores as prefixes from method names
- Switched to capital casing for test data variable names

I have chosen to retain the function docstrings detailing the link between fields in the transformation output and columns in the database, because I feel it is helpful to include this context especially given the complexity of the mapping / load process that sits between the transformed data and the database, as well as the fact we will be onboarding new developers soon. I do definitely understand the point around these comments not focusing specifically on their methods, and the plan will be to remove them once the downstream process has been simplified.

The other thing to note is that I have not removed NA columns from the transformation output where those columns will be mapped to actual database columns rather than to fields in the `event_data_blob` JSON blob. For example, I have retained `"Primary Intervention Theme": [pd.NA] * len(project_ids)` in the transformed output that sits in the "Project Details" DataFrame, because `primary_intervention_theme` remains a column in the `project_dim` table. This was just because I haven't double checked the effect that removing these columns might have, so I thought would take the conservative approach initially.

cyrusdobbs (approved):
I'm actually in favour of using underscores as a convention for private function and method names. I think it helps to clearly show the interface into that module of functionality. Although I do also agree that ideally our approach should be consistent, so maybe this should be applied across the repo at once if it's something we want to do.

If we're happy to leave in the empty rows of data for now then do we want a TODO and ticket to look at testing and possibly removing them if they're not needed? I imagine it will be easily forgotten about if we just leave it like this.

wjrm500:
> I'm actually in favour of using underscores as a convention for private function and method names. I think it helps to clearly show the interface into that module of functionality. Although I do also agree that ideally our approach should be consistent, so maybe this should be applied across the repo at once if it's something we want to do.
> 
> If we're happy to leave in the empty rows of data for now then do we want a TODO and ticket to look at testing and possibly removing them if they're not needed? I imagine it will be easily forgotten about if we just leave it like this.

@cyrusdobbs - fair enough, I would tend to agree with you regarding using underscores to demonstrate clearly the public interface of a module. I've removed them for now as per Noel's comment, but implementing this sort of pattern more broadly could be a good idea for the future. Regarding adding a TODO and creating a ticket for possibly removing from the transformation output any fields whose values are all NA, I will do this now 👍

PR #467, opened 2024-02-28
[BAU] pandas read XLSX without header

cyrusdobbs:
### Change description

`tables` and the PF flow more generally expects the workbook data to be loaded into dfs with the first row of the Excel file to be the first row of the DataFrame. However, TF dependencies have all been developed with the first row being extracted as the DataFrames' header. For now, `extract_data` has been changed to have different behaviour for TF than any other fund to account for this (pd.read_xlsx header=0 for TF, None otherwise) - this is hacky and not optimal. 

This PR aims to modify the TF flow to expect the extracted dfs without the first row as the header. This turned out to be a bigger task than I'd hoped and there's a decent risk that it introduces bugs (most likely with validation error message cell indexes I think). 

Changes include: 

- incrementing `iloc` indexing to account for the row that was previously the header now being the first row of the dfs
- reduce the increment to all indexes at the end of transformation and errors raised during transformation from + 2 to + 1 (+1 accounts for the Excel file being 1-indexed, the other +1 was to account for the first row being extracted as the header)
- incrementing the `Check.row` attributes in the initial validation schemas
- modifying the hardcoded indexes used during sign off validation
- modifying transformation test fixtures to read csvs with `header=None` inline with the change in `ingest.extract_data`
- incrementing indexes in tests - assertion data and ilocs/iat use

.

### How to test


Joel-Pearce (approved):
LGTM to me, but might be worth checking with someone less familiar with the code to see if the comments help to explain why we are manipulating the indexes in the way we are.

PR #468, opened 2024-02-28
SMD-652 - Extract mappings from the Pathfinders template

wjrm500:
### Ticket
https://dluhcdigital.atlassian.net/browse/SMD-652

### Change description
- Added a new EPV (Extract, Process, Validate) config dictionary to `core/table_configs/pathfinders/round_1.py` support extraction of tables containing required mappings data located in hidden sheets of Pathfinders Reporting Template
- Added a new function `create_mappings` in `core/transformation/pathfinders/round_1/mappings.py` to take extracted tables and pull out the following mappings: programme name → programme ID, project name → project ID, programme name → project names, programme name → allowed bespoke outputs, programme name -> allowed outcomes, and lists of allowed standard outputs and allowed standard outcomes (not technically mappings)

Idea will be to call the `extract_process_validate_tables` method in `ingest.py` with `workbook_data` and the new EPV config defined in this PR, and then pass the returned extracted and validated tables to the `create_mappings` function. We can then pass the mappings to the transformation function (further integration work required to integrate transformation). If we need mappings earlier on (at validation stage, not transformation), we just need to shift this extraction up higher in the execution flow so that the mappings dependencies are available downstream wherever they are needed.

N.B. A separate PR will be required to extract intervention theme -> output and intervention theme -> outcome mappings, which AFAIK do not currently exist in the spreadsheet


wjrm500:
Output -> Intervention Theme and Outcome -> Intervention Theme mappings also need to be extracted.

core/controllers/ingest.py
```diff
@@ -464,3 +464,38 @@ def parse_auth(body: dict) -> dict | None:
         except (JSONDecodeError, TypeError) as err:
             abort(400, "Invalid auth JSON", err)
     return auth
+
+
+def create_pathfinders_mappings(extracted_tables: dict[str, list[pd.DataFrame]]) -> dict[str, dict | list[str]]:
```
cyrusdobbs, line 469: As this is PF specific it doesn't feel right to keep it here. We could keep it with the transformation code as it transforms the extracted mapping tables into more usable mappings?

core/table_configs/pathfinders/round_1_mappings.py
```diff
@@ -0,0 +1,42 @@
+PF_ROUND_1_MAPPINGS_EPV_CONFIG = {
```
cyrusdobbs, line 1: Is the plan to combine this with the existing mappings?

core/controllers/ingest.py
```diff
@@ -464,3 +464,38 @@ def parse_auth(body: dict) -> dict | None:
...
+    standard_outcomes_df = extracted_tables["Standard outcomes"][0]
+    allowed_standard_outcomes = [row["Standardised Outcomes"] for _, row in standard_outcomes_df.iterrows()]
+    return {
+        "programme_name_to_id": programme_name_to_id,
+        "project_name_to_id": project_name_to_id,
+        "programme_id_to_project_ids": programme_id_to_project_ids,
+        "programme_id_to_allowed_bespoke_outputs": programme_id_to_allowed_bespoke_outputs,
+        "programme_id_to_allowed_bespoke_outcomes": programme_id_to_allowed_bespoke_outcomes,
+        "allowed_standard_outputs": allowed_standard_outputs,
+        "allowed_standard_outcomes": allowed_standard_outcomes,
```
cyrusdobbs, line 500: Where do you anticipate us using these mappings?

wjrm500:
Hey @cyrusdobbs I've just updated this PR to address your comments and also to improve elsewhere. Specifically...

In relation to your comments:
> As this is PF specific it doesn't feel right to keep it here. We could keep it with the transformation code as it transforms the extracted mapping tables into more usable mappings?
This was in regards to having `create_pathfinders_mappings` in `ingest.py`. I have restructured so that this function now sits in a new module: `core/transformation/pathfinders/round_1/mappings.py`.

> Is the plan to combine this with the existing mappings?
This was in regards to the new EPV config for mappings tables. I have moved this config into the same module as the "user data" extraction config: `core/table_configs/pathfinders/round_1.py` and renamed to `PF_MAPPINGS_CONFIG`. `PF_TABLE_CONFIG` is a separate dict in the same module.

> Where do you anticipate us using these mappings?
This was in regards to the allowed standard outputs and allowed standard outcomes mappings. I figured that we could these for validation for the standard outputs and standard outcomes tables, rather than having those lists hard-coded our side. Even if we eventually decided not to use these for validation though, feels useful to have them available to us at this stage?

Other changes made include:
- After moving `create_mappings` to a new module, I decided to split out the processing into sub-functions similar to how we do for `pathfinders_transform`. I have prefixed private methods with an underscore as discussed in that other PR, and also took the opportunity to add the underscore prefix to the sub-functions in the transform module for consistency. Slightly out of scope but saves having another ticket / PR for something quite trivial
- I added mock data for the extracted mappings tables and a series of unit tests that use this mock data

cyrusdobbs (approved):
LGTM

PR #469, opened 2024-03-04
[FMD-234]: Add Project Finance Change table

Joel-Pearce:
- Adds the ```ProjectFinanceChange``` table.
- Updates serialisation & tests for this table.
- Adds mapping for this table.


core/db/entities.py
```diff
@@ -507,3 +508,23 @@ class RiskRegister(BaseModel):
...
         ),
     )
+
+
+class ProjectFinanceChange(BaseModel):
+    """Stores Project Finance Change data for projects."""
+
+    __tablename__ = "project_finance_change"
+
+    project_id: Mapped[GUID] = sqla.orm.mapped_column(
```
nsnape, line 518: as discussed, potentially map to programme level?

tests/serialisation_tests/test_serialisation.py
```diff
@@ -221,6 +222,23 @@ def test_serialise_download_data_no_filters(seeded_test_client, additional_test_
         "Place",
         "OrganisationName",
     ]
+    assert list(test_serialised_data["ProjectFinanceChange"][0].keys()) == [
```
nsnape, line 225: out of interest, if we don't add PFC data in the conftest, does this return here as an empty table, or do we just drop empty tables?

nsnape (changes requested):

tests/serialisation_tests/test_serialisation.py
Joel-Pearce: It returns an empty table.

The current behaviour is that we serialise out empty tabs where there's no information for a given table relative to a set of queries. I presumed this would remain the case for empty tables, or else we'll have to write conditionals for tables that we don't want to serialise out if empty.

core/db/entities.py
```diff
@@ -503,3 +506,23 @@ class RiskRegister(BaseModel):
...
+
+class ProjectFinanceChange(BaseModel):
+    """Stores Project Finance Change data for projects."""
+
+    __tablename__ = "project_finance_change"
+
+    programme_junction_id: Mapped[GUID] = sqla.orm.mapped_column(
+        sqla.ForeignKey("programme_junction.id", ondelete="CASCADE"), nullable=False
+    )
+    event_data_blob = sqla.Column(JSONB, nullable=True)
```
nsnape, line 519: I guess just make sure this is in-line with standard blob column naming

core/db/queries.py
```diff
@@ -533,6 +533,33 @@ def risk_register_query(base_query: Query) -> Query:
...
+    """
+    extended_query = (
+        base_query.join(
+            ents.ProjectFinanceChange, ents.ProjectFinanceChange.programme_junction_id == ents.ProgrammeJunction.id
+        )
+        .with_entities(
+            ents.Submission.submission_id,
+            ents.Programme.programme_id,
+            ents.ProjectFinanceChange.data_blob,
+            ents.Project.project_name,
```
nsnape, line 553: I don't think this field makes sense now (Project.project_name). The query should pull the PFC project names from the jsonb and differentiate them

nsnape (changes requested):

core/db/queries.py
Joel-Pearce: Makes sense

nsnape (approved):

PR #470, opened 2024-03-05
[FMD-242]: add jsonb col to submission dim

Joel-Pearce:
Adds a column ```data_blob``` to ```submission_dim```.


MarcUsher (approved):
LGTM :thumbsup:

core/db/entities.py
```diff
@@ -32,6 +32,7 @@ class Submission(BaseModel):
     reporting_period_end = sqla.Column(sqla.DateTime(), nullable=False)
     reporting_round = sqla.Column(sqla.Integer(), nullable=False)
     submission_filename = sqla.Column(sqla.String(), nullable=True)
+    event_data_blob = sqla.Column(JSONB, nullable=True)
```
nsnape, line 35: not sure the label of "event data" is strictly speaking accurate here - we should probably just call it `meta_data_blob` or similar?

nsnape (changes requested):

nsnape (approved):

core/db/entities.py
MarcUsher: Does the jsonb blob functionality not assume that it will be called `event_data_blob`?

db/migrations/versions/012_event_data_to_jsonb.py
```diff
@@ -22,12 +22,12 @@ def upgrade():
 
     # PROJECT PROGRESS #
     with op.batch_alter_table("project_progress", schema=None) as batch_op:
-        batch_op.add_column(sa.Column("event_data_blob", postgresql.JSONB(astext_type=sa.Text()), nullable=True))
+        batch_op.add_column(sa.Column("data_blob", postgresql.JSONB(astext_type=sa.Text()), nullable=True))
```
nsnape, line 25: Please dont alter historical (already applied migrations) as it won't reflect actual system changes and also its easy to get the dB in a bad state

nsnape: I think it requires a new migration yo rename the existing cols, sorry

nsnape (changes requested):

db/migrations/versions/016_rename_event_data_blob.py
```diff
@@ -0,0 +1,117 @@
...
+down_revision = "015_add_jsonb_to_submission"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    with op.batch_alter_table("funding", schema=None) as batch_op:
+        batch_op.add_column(sa.Column("data_blob", postgresql.JSONB(astext_type=sa.Text()), nullable=True))
+        batch_op.drop_column("event_data_blob")
```
MarcUsher, line 24: Would this result in us losing whatever data was in `event_data_blob`? Does it need a step in between to set the contents of `data_blob` to be the same as `event_data_blob` before it's dropped, similar to what you did when setting up the `event_data_blob` columns originally?

Joel-Pearce: Yes, it should be just renamed

nsnape (approved):
Lgtm. Just probs worth a check on dev that both migrations get applied. I presume they will

PR #472, opened 2024-03-06
BAU: remove orm on submission dim and replace with raw sql

Joel-Pearce:
Removes an ORM query, ```session.query(Submission)```, which was incorrectly selecting attributes belonging to a later migration for ```Submission```, and replaces it with raw SQL.


gidsg (approved):
LGTM (have tested on dev re-running this migration) 👍

db/migrations/versions/010_programme_junction.py
```diff
@@ -91,9 +91,15 @@ def upgrade():
...
     # and add this as FK to corresponding (new) row in junction table.
-    for submission in session.query(Submission):
+    sql_query = text(
+        """
+        SELECT id from submission_dim;
+        """
+    )
+    result = connection.execute(sql_query)
+    for row in result:
         programme_junction_row = ProgrammeJunction(
```
nsnape, line 101: should we also change this ORM class?

nsnape (changes requested):

db/migrations/versions/010_programme_junction.py
Joel-Pearce: I think ```ProgrammeJunction``` is very likely to be fine because we're probably not going to add an extra column (although thinking about it not sure I can say that), and because this doesn't use SQLAlchemy's ```query``` function (which does a ```SELECY *``` and was causing the headache).

I think we should however just remove all of the ORM stuff that could ever lead to any migration breaking in the future.

MarcUsher: There is a Jira ticket in the refinement bucket look at potentially adding more columns to Programme Junction (FMD-240) but yeah it was the `SELECT *` behaviour that was the issue so hopefully that wouldn't happen in this instance.

nsnape: It's more if the KWARGs for pushing to that Model change in future, this could break. But it's a small risk, and should be obvious how to fix if it occurs, so I'll approve.

nsnape (approved):

PR #473, opened 2024-03-07
SMD_671 - Integrate PF mappings extraction into pipeline

wjrm500:
### Ticket
[SMD_671 - Integrate PF mappings extraction into pipeline](https://dluhcdigital.atlassian.net/browse/SMD-671)


core/controllers/ingest.py
```diff
@@ -84,9 +89,10 @@ def ingest(body: dict, excel_file: FileStorage) -> tuple[dict, int]:
         else:
             # TODO https://dluhcdigital.atlassian.net/browse/SMD-653: replace hardcoded dependencies with dependency
             #   injection
-            _ = extract_process_validate_tables(workbook_data, PF_TABLE_CONFIG)
+            user_tables = extract_process_validate_tables(workbook_data, PF_USER_DATA_TABLE_CONFIG)  # noqa: F841
+            control_tables = extract_process_validate_tables(workbook_data, PF_CONTROL_TABLE_CONFIG)
+            control_mappings = create_control_mappings(control_tables)  # noqa: F841
```
Joel-Pearce, line 94: Can we update the docstring to mention what these are and what they do, or leave an in-line comment?

Joel-Pearce (changes requested):

core/controllers/ingest.py
```diff
@@ -130,7 +136,7 @@ def parse_body(body: dict) -> tuple[str, int, dict | None, bool]:
 
 def extract_process_validate_tables(
     workbook_data: dict[str, pd.DataFrame], tables_config: dict[str, dict]
-) -> dict[str, Table]:
+) -> dict[str, pd.DataFrame]:
```
cyrusdobbs, line 139: We need to preserve the `Table` objects because they enable errors found during cross-table validation to be mapped back to the original spreadsheet. Otherwise we'd need to hardcode indexes which I think we should definitely avoid.

core/transformation/pathfinders/round_1/control_mappings.py
```diff
@@ -0,0 +1,55 @@
+import pandas as pd
+
+
+def create_control_mappings(extracted_tables: dict[str, pd.DataFrame]) -> dict[str, dict | list[str]]:
```
cyrusdobbs, line 4: Would be good to have a docstring on at least the public method

cyrusdobbs (changes requested):

core/controllers/ingest.py
wjrm500: Noted

wjrm500:
Hey @cyrusdobbs I've added a docstring to the `create_control_mapping` function and also done a small refactor as per our conversation over Slack. The user data and control data config dictionaries have been combined into a single config dictionary, and this gets passed into `extract_process_validate_tables`. Inside that we will call `cross_validate` (I guess the callable will be injected) such that it has access to the `Table` objects for error messaging but we can keep a clean interface for `extract_process_validate_tables` of DataFrame dictionary in, DataFrame dictionary out. It can access the control data mappings by calling `create_control_mappings` and passing in the extracted tables from the merged config, which will include the control data tables necessary to create the mappings. Then downstream in transformation, we will call `create_control_mappings` again. So it's duplication but with the benefit of removing the need to inject `control_mappings` as a dependency or pass `Table` objects around at the level of the `ingest` module.

This means the actual integration of mapping extraction via a call to `create_control_mappings` will now be shifted to the PR I will shortly open for transformation integration, and we'll also need that call in the cross-validation PR you're working on.

Joel-Pearce (approved):

core/controllers/ingest.py
```diff
@@ -170,25 +167,24 @@ def extract_process_validate_tables(
                     f"{config['extract']['worksheet_name']} {error.cell.str_ref if error.cell else ''}:"
                     f" {error.message}"
                 )
+        tables[table_name] = table.df
```
cyrusdobbs, line 170: We'll still need to keep the `Table`s for now because cross-table validation will use them

wjrm500: Just had a thought - makes more sense for you to just make this change in your cross-table validation PR. All I'd be doing here would be to assign the `Table` not the `DataFrame` to the `tables` dict, but then to keep the interface of `extract_process_validate_tables` the same (`DataFrame`s returned), in the next line of code I'd need to add something like `tables = {table_name: table.df for table_name, table in tables.items()}` anyway

cyrusdobbs (approved):
LGTM

PR #474, opened 2024-03-07
FMD-241: Add GeospatialDim Table to DB

MarcUsher:
### Change description
Add new GeospatialDim table to db. 

Adds a migration which also seeds the initial reference data for the `geospatial_dim` table, including `postcode_prefix` and `itl1_region_code` and `itl1_region_name` mappings, correct as of the time of this migration. Adding the data as part of the migration means downgrades/upgrades will continue to work and the relevant reference data will always be present after the table is created.

A script to repopulate this table is also added, and the same reference data is used in testing for consistency. The script also accounts for future additions/mutations to the data by merging rows with the same id rather than dropping and replacing them to preserve future FK relationships.

### How to test


core/db/entities.py
```diff
@@ -527,3 +527,12 @@ class ProjectFinanceChange(BaseModel):
...
         ),
     )
+
+
+class GeospatialDim(BaseModel):
+    """Stores Geospatial information mapped to postcodes."""
+
+    __tablename__ = "geospatial_dim"
+
+    postcode_prefix = sqla.Column(sqla.String(), nullable=False, unique=True)
```
nsnape, line 537: probably could actually limit this in length? Maybe 4 chars? (check if there are any viable postcodes that start with 5 chars, but I don't think so)

core/db/entities.py
```diff
@@ -527,3 +527,12 @@ class ProjectFinanceChange(BaseModel):
...
+    itl_region = sqla.Column(sqla.String(), nullable=False, unique=False)
```
nsnape, line 538: oh, sorry this wasn't on the ticket, but just realised we could add separate columns for itl1_region_code and itl1_region_name. Could you update this col and add the extra one?

nsnape: hmm or do we add the names to another normalised table? Worth thinking about...

nsnape: or another option is keep them both as cols in this table, but add a unique together constraint for codes and names

nsnape: If you go with both fields, it will be worth considering which field will actually be used by the filter, and adding an index to that one.

Joel-Pearce (changes requested):
If you take a look at ```test/db_tests_test_entities.py``` there's a test called ```test_project_finance_change_table```. It's a seemingly somewhat pointless test - it merely asserts that we can put data into a given table and query it out in the way we expect to - but I thought it prudent because it made a database change without any concomitant testing.

I think this is probably a bit more worthy of testing because we're adding data in the actual migration. If that is the case I'd suggest writing a test to assert that the amount of data in the table is what we expect it to be, and to maybe query one row and take a look to ensure it's how it should be. That way if anything comes along and messes with the migration (like something that somehow drops a bunch of data for some reason) we'll have something to flag it.

Joel-Pearce:
This will need rebasing & updating the migration number BTW :)

db/migrations/versions/021_add_geospatial_dim_table.py
```diff
@@ -0,0 +1,44 @@
...
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+
+    op.create_table(
+        "geospatial_dim",
+        sa.Column("id", core.db.types.GUID(), nullable=False),
```
gidsg, line 27: Can we autogenerate the IDs instead e.g. 
`sa.Column("id", postgresql.UUID(as_uuid=True), nullable=False, primary_key=True, default=uuid.uuid4),`

MarcUsher: I think they're normally auto-generated by the base model in `entities.py` when data is added, but in the seed data I've actually already generated uuids for this reference data so that if we come to extend the table and add more mappings it'll match the ids and merge the rows so an FK relationships are preserved.

tests/db_tests/test_entities.py
```diff
@@ -76,3 +76,49 @@ def test_project_finance_change_table(seeded_test_client_rollback):
...
+        "TLL": "Wales",
+        "TLM": "Scotland",
+        "TLN": "Northern Ireland",
+        "TLZ": "Non-ITL1 Region",
+    }
+
+    for row in geospatial_data:
+        region_code = row.itl1_region_code
+        region_name = row.data_blob["itl1_region_name"]
+        assert (region_code, region_name) in itl1_region_pairs.items()
```
nsnape, line 124: this is fine and does exactly what you need, sorry to be nit-picky but would it be clearer (and slightly more performant) to do 
`assert itl_region_pairs[region_code] == region name`
or a similar assertion using .get method?

nsnape: or to deal with even smaller data structures, you could re-query the DB table, but selecting only region_code and region_name in the query, combined with a distinct, so you will iterate through less rows. Might not be worth the extra effort though :)

nsnape (changes requested):

scripts/db_seed_geospatial_dim_table.py
```diff
@@ -0,0 +1,74 @@
+from pathlib import Path
```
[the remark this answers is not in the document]
MarcUsher: I was thinking the same thing last night, seems unnecessary to have this as a separate script. Will update!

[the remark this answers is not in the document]
MarcUsher: As discussed, have removed the separate script and integrated this into the package in a new file (`reference_data.py` - thinking if we're adding more static reference data tables going forward then any associated methods can be added here). Updated the CLI commands as well.

core/reference_data.py
```diff
@@ -0,0 +1,56 @@
...
+    mapped_geospatial_data = geospatial_dim_mapping.map_data_to_models(geospatial_dim_df)
+
+    existing_geospatial_data = GeospatialDim.query.all()
+
+    if existing_geospatial_data is None:
+        db.session.add_all(mapped_geospatial_data)
+    else:
+        for geospatial_record in mapped_geospatial_data:
+            previous_record = next(
+                filter(lambda x: x.postcode_prefix == geospatial_record.postcode_prefix, existing_geospatial_data), None
```
nsnape, line 48: could this be simplified to a comprehension like such:
`next((record for record in existing_geospatial_data if record.postcode_prefix == geospatial_record.postcode_prefix), None)
`
Just because I find comprehensions nicer to read, and generally more "pythonic". Also, they can be faster (probably not a big enough dataset here to make any difference however)

README.md
```diff
@@ -121,10 +128,10 @@ flask seed-test
 ```
 
 ### drop
```
nsnape, line 130: could you update the header too please?

nsnape (changes requested):

nsnape (approved):

Joel-Pearce (approved):
LGTM

PR #475, opened 2024-03-07
BAU: fix pfc migration

Joel-Pearce:
Fixes a mistake where in programme_junction_id was nullable for ProjectFinanceChange. Also adds a test to ensure an appropriate error is flagged if the column become nullable.


db/migrations/versions/018_fix_pfc_table.py
```diff
@@ -0,0 +1,40 @@
...
+
+    # ### end Alembic commands ###
+
+
+def downgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    with op.batch_alter_table("project_finance_change", schema=None) as batch_op:
+        batch_op.drop_index("ix_project_finance_change_join_programme_junction")
+        batch_op.create_index(
+            "ix_project_finance_change_join_programme_junction", ["programme_junction_id"], unique=False
```
nsnape, line 36: I presume this is auto-generated? Any idea why it recreates what looks like exactly the same index?

nsnape: I'm also confused as to why it isn't re-created after the column is changed - this is why I presumed it was recreated

nsnape: ah, so this downgrade should actually set unique = True? And should be placed after the column is edited?

nsnape (approved):

PR #476, opened 2024-03-08
SMD-665 - Integrate PF transformation into pipeline

wjrm500:
### Ticket
[SMD-665 - Integrate PF transformation into pipeline](https://dluhcdigital.atlassian.net/browse/SMD-665)

### Summary of changes
- The actual integration of transformation itself is just one line of code in `ingest.py`: `data_to_load = pathfinders_transform(tables, reporting_round)` - this is the only "functional" change to the code made in this PR
- However adding this line causes a bunch of errors with existing tests, as changes have been made to the spreadsheet and schema while various components of the pipeline have been developed
- So the vast majority of changes in this PR are around bringing the extraction and validation config dict up to date with the spreadsheet, and then propagating changes to column names or whatever throughout the rest of the pipeline


core/transformation/pathfinders/round_1/transform.py
```diff
@@ -161,17 +161,19 @@ def _project_details(
...
-    project_ids = df_dict["Project Location"]["Project name"].map(project_name_to_id_mapping)
-    location_multiplicities = df_dict["Project Location"]["Location"].map(
+    project_ids = df_dict["Project location"]["Project name"].map(project_name_to_id_mapping)
+    location_multiplicities = df_dict["Project location"]['Project full postcode/postcodes (e.g., "AB1D 2EF")'].map(
         lambda x: "Multiple" if "," in x else "Single"
     )
-    locations = df_dict["Project Location"]["Location"].map(lambda x: list(map(str.strip, x.split(","))))
-    postcodes = df_dict["Project Location"]["Location"].map(extract_postcodes)
+    locations = df_dict["Project location"]['Project full postcode/postcodes (e.g., "AB1D 2EF")'].map(
+        lambda x: list(map(str.strip, x.split(",")))
```
Joel-Pearce, line 171: This is a bit of an unwieldy lambda, and I wonder if it might be more consistent for anything other than a very straightforward lambda to be a function. That way we can properly unit test it and avoid this difficult to read code.

Then the code will read something like:

```locations = df_dict["Project location"]['Project full postcode/postcodes (e.g., "AB1D 2EF")'].map(function)```

**Review by Joel-Pearce — CHANGES_REQUESTED** [review]
Apologies if this is done somewhere I can't see, but with regards to the column names, they should match what we find in the respective DataClass for each table in ```core/controllers/mappings.py```. If they don't match these, then the pipeline will break with a ```KeyError```.

File: core/transformation/pathfinders/round_1/transform.py (hunk 1 above)
Comment (wjrm500) [reply core/transformation/pathfinders/round_1/transform.py:171]: Hey @Joel-Pearce, while I agree with the principle that anything beyond a simple lambda should be a standalone function, I do think this lambda is quite simple - the lambda itself is just `list(map(str.strip, x.split(",")))` - split the string by commas, strip whitespace and convert to a list. I think the expression as a whole perhaps looks more complex at an initial glance because of the verbosity of the bit that comes before it - `df_dict["Project location"]['Project full postcode/postcodes (e.g., "AB1D 2EF")']`. The other reason I'd rather not put this into a standalone function is that currently in this module we have the one parent, public function `pathfinders_transform`, and then every other function is a "private sub-function" called once by `pathfinders_transform` - separating out this into its own function would break that pattern

**Comment by wjrm500** [comment]
> Apologies if this is done somewhere I can't see, but with regards to the column names, they should match what we find in the respective DataClass for each table in `core/controllers/mappings.py`. If they don't match these, then the pipeline will break with a `KeyError`.

Thanks @Joel-Pearce, could you clarify which column names in particular you are referring to?

File: core/transformation/pathfinders/round_1/transform.py (hunk 1 above)
Comment (nsnape) [reply core/transformation/pathfinders/round_1/transform.py:171]: It feels like something we should test though. Do we throw different strings at this in a test? Obvs in this case it will expensive as a unit test as it has all the extra setup / context

Comment (nsnape) [reply core/transformation/pathfinders/round_1/transform.py:171]: also, is this transforming this cell into a list? What is the reason for that as we just capture the original string in the DB for this column

**Review by nsnape — CHANGES_REQUESTED** [review]

File: core/transformation/pathfinders/consts.py [hunk 2]
```diff
@@ -8,27 +8,11 @@
 }
 
 PF_REPORTING_PERIOD_TO_DATES = {
```
nsnape, line 10: can you add an inline comment to detail exactly what this mapping is for (ie for translating dates present in the PF spreadsheet, I presume)

core/transformation/pathfinders/round_1/transform.py
```diff
@@ -334,7 +336,8 @@ def _outputs(
             amount                  - from "Amount" in the transformed DF "Output_Data"
             additional_information  - from "Additional Information" in the transformed DF "Output_Data"
     """
-    organisation_name = df_dict["Organisation Name"].iloc[0, 0]
+    # TODO: Handle bespoke outputs
```
nsnape: can we try and put a Jira ticket number on all TODO's going forward (action from last away day)

nsnape (changes requested):

Joel-Pearce (approved):
LGTM

nsnape (approved):

PR #479, opened 2024-03-13
SMD-667 - Fix cell reference string in case of ignored non-header rows

wjrm500:
### Ticket
[Fix cell reference string in case of ignored non-header rows](https://dluhcdigital.atlassian.net/browse/SMD-667)

### Change summary
By removing all index resetting during table processing, all table DataFrames contain the original indices from the spreadsheet, removing the need for additional logic to map back from table-scope row indices to original indices even in the case of arbitrary removal of rows. This allows us to dispense with the `first_row_idx` attribute of the `Table` class.

Because DataFrames now contain original indices and not reset ones...
- The `_remove_ignored_non_header_rows` method of `TableProcessor` had to be adjusted to remove rows based on `ignored_non_header_rows` indices matching the actual _position_ of rows in the DataFrame (the index values you'd get if you ran a `reset_index`), rather than the index (as those two things had diverged).
- The indices of expected DataFrames in the extraction tests had to be updated

### DIFF
--- a/tables/process.py
+++ b/tables/process.py
@@ -135,6 +135,7 @@ class TableProcessor:
         if rows := [idx for idx in self.ignored_non_header_rows if idx not in table.df.index]:
             raise TableProcessingError(f"Ignored non-header rows {rows} are out-of-bounds.")
         table.df = table.df.drop(self.ignored_non_header_rows)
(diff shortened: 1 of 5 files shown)


Joel-Pearce (approved):
LGTM if all of the cells flagged up by the integration tests are correct.

PR #482, opened 2024-03-14
FMD-250: add submission metadata to serialisation

gidsg:
### Change description
adds submission metadata to serialisation

### How to test
SubmissionRef tab should appear on download

### DIFF
--- a/core/const.py
+++ b/core/const.py
@@ -701,7 +701,7 @@ TABLE_SORT_ORDERS = {
     "OutcomeRef": ["OutcomeName"],
     "OutcomeData": ["SubmissionID", "ProjectID", "Outcome", "StartDate", "EndDate", "GeographyIndicator"],
     "RiskRegister": ["SubmissionID", "ProgrammeID", "ProjectID", "RiskName"],
-    "SubmissionRef": ["SubmissionID", "ReportingPeriodStart", "ReportingPeriodEnd", "ReportingRound"],
+    "SubmissionRef": ["SubmissionID", "ProgrammeID", "ReportingPeriodStart", "ReportingPeriodEnd", "ReportingRound"],
 }
 
 # map place to fund type associated with it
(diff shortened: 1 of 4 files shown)


core/const.py
```diff
@@ -701,6 +701,7 @@ class ITLRegion(StrEnum):
     "OutcomeRef": ["OutcomeName"],
     "OutcomeData": ["SubmissionID", "ProjectID", "Outcome", "StartDate", "EndDate", "GeographyIndicator"],
     "RiskRegister": ["SubmissionID", "ProgrammeID", "ProjectID", "RiskName"],
+    "SubmissionRef": ["SubmissionID", "ReportingPeriodStart", "ReportingPeriodEnd", "ReportingRound"],
```
nsnape: This doesn't need multiple values, as SubmissionID's are unique, and should only appear in this extract once each, so the other columns will never be sorted on. So can just keep "SubmissionID"

core/serialisation/data_serialiser.py
```diff
@@ -418,35 +454,15 @@ class Meta:
...
-    forecast_or_actual_change = fields.String(
-        attribute="data_blob.forecast_or_actual_change", data_key="ForecastOrActualChange"
-    )
-    reporting_period_change_takes_place = fields.String(
-        attribute="data_blob.reporting_period_change_takes_place", data_key="ReportingPeriodChangeTakesPlace"
-    )
-    programme_name = auto_field(model=Programme, data_key="Place")
-    organisation_name = auto_field(model=Organisation, data_key="OrganisationName")
+    reporting_period_start = auto_field(data_key="ReportingPeriodStart")
+    reporting_period_end = auto_field(data_key="ReportingPeriodEnd")
```
nsnape, line 467: Sorry, I forgot to mention, we serialise dates slightly differently, in order to get them into Excel as Excel Datetime objects (rather than just strings). See [this example ](https://github.com/communitiesuk/funding-service-design-post-award-data-store/blob/3ba0261aa36ef7d74dd94be5e198f0f9c5b5d587/core/serialisation/data_serialiser.py#L179C5-L179C50)

nsnape: Then the Pandas ExcelWriter function at core.controllers.download.data_to_excel handles the conversion automatically

nsnape (changes requested):

core/serialisation/data_serialiser.py
gidsg: fixed in 1239c6f

core/const.py
gidsg: fixed in 328ae2f

nsnape (approved):

PR #486, opened 2024-03-18
SMD-699 - Align indices in transformed DFs

wjrm500:
### Ticket
[SMD-699 - Review indexes](https://dluhcdigital.atlassian.net/browse/SMD-699)

### Summary of changes
Created new `create_dataframe` method in transformation utils to replace direct calls to `pd.DataFrame` constructor in each transformation function. In this method we reset all indices of input `pd.Series` to ensure indices align and columns merge correctly.

### DIFF
--- a/core/transformation/utils.py
+++ b/core/transformation/utils.py
@@ -1,6 +1,7 @@
 """Module for reusable DataFrame transformation functions."""
 
 import re
+from typing import Iterable
 
 import numpy as np
 import pandas as pd
@@ -114,3 +115,18 @@ def extract_postcodes(s: str) -> list[str] | None:
         if postcode_area_matches == []:
             return None
     return postcode_area_matches
+
+
+def create_dataframe(data: dict[str, Iterable]) -> pd.DataFrame:
+    """
+    Create a DataFrame from a dictionary of Series or lists, aligning the indices.
+
+    :param data: Dictionary of Series or lists
+    :return: DataFrame with aligned indices
+    """
+    return pd.DataFrame(
+        {
+            column: val.reset_index(drop=True) if isinstance(val, pd.Series) else pd.Series(val)
+            for column, val in data.items()
+        }
+    )
--- a/core/transformation/pathfinders/round_1/transform.py
+++ b/core/transformation/pathfinders/round_1/transform.py
@@ -227,9 +227,6 @@ def _project_progress(
     delivery_rags = df_dict["Project progress"]["Delivery RAG rating"].map(rag_to_integer_mapping)
     spend_rags = df_dict["Project progress"]["Spend RAG rating"].map(rag_to_integer_mapping)
     commentaries = df_dict["Project progress"]["Why have you given these ratings? Enter an explanation (100 words max)"]
-    # TODO: https://dluhcdigital.atlassian.net/browse/SMD-699
-    # set project_ids index to match other columns as is misaligned due to being taken from a different DataFrame
-    project_ids.index = delivery_rags.index
     return create_dataframe(
         {
             "Project ID": project_ids,
--- a/core/transformation/pathfinders/round_1/transform.py
+++ b/core/transformation/pathfinders/round_1/transform.py
@@ -10,7 +10,7 @@ from core.transformation.pathfinders.consts import (
 from core.transformation.pathfinders.round_1.control_mappings import (
     create_control_mappings,
 )
-from core.transformation.utils import extract_postcodes
+from core.transformation.utils import create_dataframe, extract_postcodes
 
 
 def pathfinders_transform(
@@ -62,7 +62,7 @@ def _submission_ref(
     sign_off_name = df_dict["Sign off name"].iloc[0, 0]
     sign_off_role = df_dict["Sign off role"].iloc[0, 0]
     sign_off_date = df_dict["Sign off date"].iloc[0, 0].isoformat()
-    return pd.DataFrame(
+    return create_dataframe(
         {
             "Submission Date": [datetime.now()],
             "Reporting Period Start": [PF_REPORTING_ROUND_TO_DATES[reporting_round]["start"]],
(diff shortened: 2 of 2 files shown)


core/transformation/utils.py
```diff
@@ -114,3 +115,18 @@ def extract_postcodes(s: str) -> list[str] | None:
         if postcode_area_matches == []:
             return None
     return postcode_area_matches
+
+
+def create_dataframe(data: dict[str, Iterable]) -> pd.DataFrame:
```
nsnape, line 120: is this type hinting too loose? Can this be union of the specific types expected rather than all Python iterables?

nsnape (changes requested):

core/transformation/utils.py
nsnape: Since this theoretically could take a string as a param, and give all sorts of weird results.

wjrm500: I'd originally considered `Union[pd.Series, List]` but then realised we have a tuple coming in as well, and that point I thought it best to simply type-hint `Iterable` rather than a three-way `Union` to be clean. I reckon it's okay to keep type hint as-is based on the [Pandas documentation for `pd.Series`](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.Series.html) which suggests that the constructor data argument can be "array-like, Iterable, dict, or scalar value"

Joel-Pearce (changes requested):
Can we have a unit test for ```create_dataframe```? Preferably one that simulates misaligned indexes in one of the columns getting correct aligned by the function

wjrm500:
> Can we have a unit test for `create_dataframe`? Preferably one that simulates misaligned indexes in one of the columns getting correctly aligned by the function

@Joel-Pearce have just added this unit test now

tests/transformation_tests/test_transformation_utils.py
```diff
@@ -99,3 +101,32 @@ def test_excel_datetime_conversion():
...
+    # Create a DataFrame from the dictionary using pd.DataFrame constructor directly to observe problem
+    result = pd.DataFrame(data)
+    expected_output = pd.DataFrame(
+        data={
+            "A": [1, 2, 3, np.nan, np.nan, np.nan],
+            "B": [np.nan, np.nan, np.nan, 4, 5, 6],
+        },
+        index=range(6),
+    )
+    assert_frame_equal(result, expected_output)
```
Joel-Pearce, line 121: Why do we need to do this?

tests/transformation_tests/test_transformation_utils.py
```diff
@@ -99,3 +101,32 @@ def test_excel_datetime_conversion():
...
+    # Create a DataFrame from the dictionary using the create_dataframe function to observe the fix
+    result = create_dataframe(data)
+    expected_output = pd.DataFrame(
+        {
+            "A": [1, 2, 3],
+            "B": [4, 5, 6],
+        },
+        index=[0, 1, 2],
+    )
+    assert_frame_equal(result, expected_output)
```
Joel-Pearce, line 132: How do we know we should prioritise the index of A over the index of B?

Joel-Pearce (changes requested):

tests/transformation_tests/test_transformation_utils.py
wjrm500: The index of A isn't being prioritised, the indices of both series are reset - I've just pushed a commit to change the index of A to `[1, 2, 3]` to make this clearer

Joel-Pearce: Ah of course, we don't even need the index post-transformation so can just reset

tests/transformation_tests/test_transformation_utils.py
wjrm500: Or better might be to move the example into the docstring so we're not actually executing the code here - will do this now

core/transformation/utils.py
```diff
@@ -114,3 +115,33 @@ def extract_postcodes(s: str) -> list[str] | None:
...
+        "B": pd.Series([4, 5, 6], index=[4, 5, 6]),
+    })
+    expected_output = pd.DataFrame(
+        data={
+            "A": [1, 2, 3, np.nan, np.nan, np.nan],
+            "B": [np.nan, np.nan, np.nan, 4, 5, 6],
+        },
+        index=[1, 2, 3, 4, 5, 6],
+    )
+    assert pandas.testing.assert_frame_equal(actual_output, expected_output)  # Returns True
```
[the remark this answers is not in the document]
wjrm500: Have just modified docstring

Joel-Pearce (approved):
LGTM

core/transformation/utils.py
nsnape: since python 3.10 you can union types with |

nsnape: so I _think_ it would look something like
`dict[str, pd.Series | List]`

nsnape: what is the tuple that comes in exactly?

wjrm500: I've just pushed a commit converting from `Union` syntax to pipe operator. I've kept tuple in the definition for now because the tuple examples come from the below code:

`questions, answers = zip(*[(q, a) for q, a in zip(questions, answers) if pd.notna(a)])
  return create_dataframe(
        {
            "Programme ID": [programme_id] * len(questions),
            "Question": questions,
            "Answer": answers,
        }
    )`

To convert back from tuples to lists, we'd have to call `list(questions)` and `list(answers)` and this doesn't feel worth it just for the sake of the type hint?

nsnape: In terms of clean code and re-useability, making your functions less variable/ambiguous should be a priority - It makes it easier to re-use / extend in future and reduces the risk of misuse. Also, if we used static type checker like MyPy it would probably complain about this. However, it's probably fine here as built as part of a bespoke pipeline. 

But, please do add it do the docstring description, otherwise it's not clear what kind of Tuple is actually expected.

wjrm500: I've parameterised the `list` and `tuple` type hints to make the type-hints self-documenting, rather than adding extra text to the docstring, such that the function interface now looks like:

```
def create_dataframe(
    data: dict[str, pd.Series | list[datetime | int | str] | tuple[datetime | int | str]]
) -> pd.DataFrame:
```

nsnape (approved):

PR #495, opened 2024-03-21
[#FMD-255]: e2e integration test

Joel-Pearce:
- Fixes a bug which tries to sort the column output of the extract on columns that don't exist for Pathfinders.
- Fixes a bug which caused a ```OutcomeRef``` data to only join at the project and not the programme level during serialisation.
- Updates the re-ingest integration test to make it more meaningful.
- Adds an end-to-end integration test.


tests/controller_tests/test_ingest_functions.py
```diff
@@ -16,9 +16,7 @@
 from core.controllers.load_functions import next_submission_id
 
 # flake8: noqa
```
nsnape, line 18: do we still need this noqa?

tests/integration_tests/test_ingest_component_towns_fund.py
```diff
@@ -217,6 +203,13 @@ def test_ingest_with_r4_file_success_with_load_re_ingest(
         },
     )
 
+    programme_junction_rows_first_ingest = ProgrammeJunction.query.all()
+    programme_id_first_ingest = programme_junction_rows_first_ingest[0].programme_id
+    submission_id_first_ingest = programme_junction_rows_first_ingest[0].submission_id
+    project_detail_rows_first_ingest = Project.query.all()
+
+    db.session.commit()
```
nsnape, line 211: is this commit superfluous?

nsnape (changes requested):

tests/integration_tests/test_ingest_component_towns_fund.py
Joel-Pearce: Have to include or else raises ```A transaction is already begun on this Session.```

nsnape: can you add an inline comment to state it's purpose?

tests/controller_tests/test_ingest_functions.py
Joel-Pearce: Yes

tests/integration_tests/test_ingest_component_all_funds.py
```diff
@@ -0,0 +1,363 @@
...
+
+
+def test_multiple_rounds_multiple_funds_end_to_end(
+    test_client_reset,
+    towns_fund_round_3_file_success,
+    towns_fund_round_4_file_success,
+    pathfinders_round_1_file_success,
+    test_buckets,
+):
+    """Tests that the ingestion of multiple rounds and funds will be loaded into the database
```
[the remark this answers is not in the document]
nsnape: with an emphasis on the bits that are checking certain behaviours?

core/serialisation/data_serialiser.py
```diff
@@ -160,10 +160,10 @@ class Meta:
 
     submission_id = auto_field(model=Submission, data_key="SubmissionID")
     programme_id = auto_field(model=Programme, data_key="ProgrammeID")
-    question = fields.String(attribute="data_blob.question", data_key="Question")
-    indicator = fields.String(attribute="data_blob.indicator", data_key="Indicator")
-    response = fields.String(attribute="data_blob.response", data_key="Answer")
-    guidance_notes = fields.String(attribute="data_blob.guidance_notes", data_key="GuidanceNotes")
+    question = fields.String(attribute="data_blob.question", data_key="Question", dump_default=None)
```
nsnape, line 163: I wonder, it may be cleaner to try and make a base class that extends fields.String (and for the other fields methods we use) and always sets dump_default=None. Then in the serialisation classes we just invoke our custom class rather than fields.String? Means we don't have to repeat that kwarg every time, and also easier to change the behaviour accross the board by juts changing in one place?

core/serialisation/data_serialiser.py
```diff
@@ -160,10 +160,10 @@ class Meta:
...
+    indicator = fields.String(attribute="data_blob.indicator", data_key="Indicator", dump_default=None)
+    response = fields.String(attribute="data_blob.response", data_key="Answer", dump_default=None)
+    guidance_notes = fields.String(attribute="data_blob.guidance_notes", data_key="GuidanceNotes", dump_default=None)
     programme_name = auto_field(model=Programme, data_key="Place")
```
nsnape, line 167: so the behaviour for the jsonb fields now defaults to empty cell if no data present, but does the same thing happen for auto_fields? We should check we have consistent behaviour between these 2 different methods of serialising.

nsnape (changes requested):

core/serialisation/data_serialiser.py
nsnape: something like:

```
from marshmallow import fields

class CustomFloatField(fields.Float):
    def __init__(self, *args, **kwargs):
        kwargs.setdefault("dump_default", None)
        super().__init__(*args, **kwargs)
```

I haven't tried the above, it is from bing chat, so might be garbage :D

core/serialisation/data_serialiser.py
Joel-Pearce: I'll have to doublecheck, but I don't think auto_fields has the ```dump_default``` param. The issue was that values in the JSONB are not guaranteed to be present between funds and rounds, and this causes a mismatch of column orderings depending on what data is ingested. All of the auto_field fields are either FKs, PKs, non-nullable, or datetimes, and so (apart from maybe a couple of edge cases in the latter) don't need a default value for the sake of ordering.

core/serialisation/data_serialiser.py
Joel-Pearce: I'll take a look at this

Joel-Pearce: Haven't thought about ```fields.Float``` or ```fields.Integer``` either. This needs to be QA'd as will effect data in extract

core/serialisation/data_serialiser.py
```diff
@@ -74,6 +74,30 @@ def _serialize(self, value, attr, obj, **kwargs):
         return ", ".join(value)
 
 
+class JSONBFloatField(fields.Float):
```
nsnape, line 77: could you also update the module docstring to include the JSONB stuff, including these new custom classes?

nsnape (changes requested):

core/serialisation/data_serialiser.py
```diff
@@ -88,6 +112,13 @@ def serialise_download_data(
     Each extended query and its corresponding schema should be added to the table_queries object in order to be
     serialised. Each additional query uses the base_query parameter as a starting point.
 
+    Fields serialised from a JSONB use custom serialisation classes in which dump_default=None.
```
nsnape, line 115: could you move this bit to the module docstring, and make sure the "guidance" in there is up to date?

nsnape (changes requested):

nsnape (approved):

PR #496, opened 2024-03-21
FMD-226 Add TF ProgrammeManagement DB table (Funding section B2)

nsnape:
https://dluhcdigital.atlassian.net/browse/FMD-226

### Change description
A new DB table definition + migration to store Section B2 from TF funding tab(specifically for Towns Deal returns)
Updated ingest pipelines/schemas/mappings, DB, serialisation, and tests.

### How to test
Run locally, ingest some TD and HS fund files. Extract should now show this data, but only for TD forms.


core/controllers/mappings.py
```diff
@@ -533,4 +533,30 @@ def get_row_id(model: Type[BaseModel], lookups: dict[str, Any]) -> str | None:
...
+        table="Programme Management",
+        model=ents.ProgrammeManagement,
+        column_mapping={
+            "Submission ID": "submission_id",
+            "Programme ID": "programme_id",
+            "Payment Type": "payment_type",
+            "Spend for Reporting Period": "spend_for_reporting_period",
+            "Start_Date": "start_date",
+            "End_Date": "end_date",
+            "Actual/Forecast": "forecast_or_actual",
```
Joel-Pearce, line 546: Can we standardise this to ```state```, as it appears in Outcomes and Outputs?

Joel-Pearce (changes requested):
"Programme Management" needs to be extracted out of TF R4 as well, right? I may be mistaken, but I think this only extracts the table for Round 3. Could we have some testing on both rounds to ensure that the table is now present when a sheet is ingested for either?

core/controllers/mappings.py
nsnape: yeah I was going to ask about that, this was my 50/50 guess 😆

core/serialisation/data_serialiser.py
```diff
@@ -306,6 +309,25 @@ class Meta:
...
+class ProgrammeManagementSchema(SQLAlchemySchema):
+    """Serialise ProgrammeManagement data"""
+
+    class Meta:
+        model = ProgrammeManagement
+
+    submission_id = auto_field(model=Submission, data_key="SubmissionID")
+    programme_id = auto_field(model=Programme, data_key="ProgrammeID")
+    payment_type = fields.String(attribute="data_blob.payment_type", data_key="PaymentType")
+    spend_for_reporting_period = fields.Integer(
```
Joel-Pearce, line 321: For other numerical fields we're serialising out of the data blob we use ```fields.Number```. I'm not sure if there's a meaningful difference, but we should probably standardise or specify why different?

Joel-Pearce (changes requested):

core/serialisation/data_serialiser.py
nsnape: good spot. I'll check the difference in the docs and choose one method for everything

nsnape: Changed this one to Float. Will fix the other existing ones in a separate PR

core/serialisation/data_serialiser.py
```diff
@@ -306,6 +309,25 @@ class Meta:
...
+class ProgrammeManagementSchema(SQLAlchemySchema):
+    """Serialise ProgrammeManagement data"""
+
+    class Meta:
+        model = ProgrammeManagement
+
+    submission_id = auto_field(model=Submission, data_key="SubmissionID")
+    programme_id = auto_field(model=Programme, data_key="ProgrammeID")
+    payment_type = fields.String(attribute="data_blob.payment_type", data_key="PaymentType")
+    spend_for_reporting_period = fields.Float(
```
Joel-Pearce, line 321: Can we change all of the numbers to ```fields.Float``` when serialised out of the blob now? Or else we have some as ```Number``` and it's not clear why

Joel-Pearce (changes requested):

core/serialisation/data_serialiser.py
nsnape: yeah, as above comment I'm fixing in a separate PR

Joel-Pearce (approved):
LGTM

PR #497, opened 2024-03-21
SMD-710 - Pathfinders - validate postcodes

wjrm500:
### Ticket

[SMD-710 - Pathfinders - validate postcodes](https://dluhcdigital.atlassian.net/browse/SMD-710)

### Change summary
Add validation to check for postcode or comma-separated list of postcodes in postcodes field in "Project location" table from new Pathfinder Reporting Template.

### DIFF
--- a/tables/checks.py
+++ b/tables/checks.py
@@ -17,6 +17,7 @@ However, there are two main disadvantages of schemas with inline custom checks:
     2. you can’t use them to synthesize data because the checks are not associated with a hypothesis strategy.
 """
 
+import re
 from datetime import datetime
 
 import pandas as pd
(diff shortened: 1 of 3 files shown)


Joel-Pearce (changes requested):

tables/checks.py
```diff
@@ -67,3 +68,21 @@ def max_word_count(element, *, max_words):
...
+@pa.extensions.register_check_method(check_type="element_wise")
+def postcode_list(element):
+    """Checks that a string can be split on commas and each element matches a basic UK postcode regex.
+
+    :param element: an element to check
+    :return: True if passes the check, else False
+    """
+    if not isinstance(element, str):
+        raise TypeError("Value must be a string")
+    postcode_regex = r"^[A-Z]{1,2}\d[A-Z\d]? ?\d[A-Z]{2}$"
```
nsnape, line 82: This is slightly different to the regex here https://github.com/communitiesuk/funding-service-design-post-award-data-store/blob/71b823014f2f02340da8c3271ee4cff69721c8a8/core/transformation/utils.py#L9

Should they be the same? Or use the same source?

nsnape (changes requested):

tables/checks.py
wjrm500: Existing postcode regex allows this postcode, which new regex marked as invalid: `SW1A 1AA1`. Based on the fact that this postcode is invalid, I will replace the existing POSTCODE_REGEX const with the new one and use this in `checks.py`.

image

wjrm500: Scratch that last message, changing `POSTCODE_REGEX` causes a bunch of tests to fail. I think best to revert to using existing regex regardless of whether or not it is objectively more accurate than the new one!

Joel-Pearce (approved):
LGTM

nsnape (approved):

PR #504, opened 2024-03-22
[BAU] update serialisation numerical field types

nsnape:
### Change description
Some fields serialised pout of JSONB were set to String, but these should be numerical types. Fixed on the schema so that they appear in Excel as the correct type (this could potentially be reverted if we stop using Excel)

### How to test


Joel-Pearce (approved):
LGTM

PR #506, opened 2024-03-25
FMD-241: Create Project-Geospatial M2M relationship

MarcUsher:
### Change description
Create many to many relationship between project_dim and geospatial_dim

- Migration to create association table between geospatial_dim and project_dim for many to many relationship
- Update CLI commands to re-add `flask drop` for completeness
- Update load behaviour to create relationship to `geospatial_dim` from projects with postcodes
- Update download query methods for region filter to now filter via the geospatial_dim relationship
- New util method to return a set of postcode prefixes from the list of project postcodes
- Update tests and add new unit tests covering new methods
- Remove now redundant itl_region hybrid property and static dictionaries in const.py

Additional data migration to populate association table with matches between existing projects and geospatial ref data will be added in a separate PR to avoid data loss during deployments. [See PR for Data Migration](https://github.com/communitiesuk/funding-service-design-post-award-data-store/pull/572)

### How to test


db/migrations/versions/028_project_geospatial_m2m.py
```diff
@@ -0,0 +1,76 @@
...
+            FROM geospatial_dim
+            WHERE LEFT(project_postcode.prefix, 1) = geospatial_dim.postcode_prefix
+            LIMIT 1
+        ) AS gd_shorter ON TRUE
+        WHERE gd_longer.id IS NOT NULL
+        """
+    )
+
+    connection.execute(sql_query)
+    session.commit()
```
Joel-Pearce, line 69: Can you just run ```op.execute``` on the SQL? It's what we've done for the JSONB stuff, but not sure if it's inconsistent elsewhere. Best to be consistent if it's what we've done elsewhere, otherwise at least it should save an import.

Joel-Pearce (changes requested):

db/migrations/versions/028_project_geospatial_m2m.py
```diff
@@ -0,0 +1,76 @@
...
+        ) AS gd_longer ON TRUE
+        LEFT JOIN LATERAL (
+            SELECT id
+            FROM geospatial_dim
+            WHERE LEFT(project_postcode.prefix, 1) = geospatial_dim.postcode_prefix
+            LIMIT 1
+        ) AS gd_shorter ON TRUE
+        WHERE gd_longer.id IS NOT NULL
+        """
+    )
```
samuelhwilliams, line 66: Do you have an estimate on the number of rows this is likely to create?

db/migrations/versions/028_project_geospatial_m2m.py
MarcUsher: I followed the method from the big migration that created the programme junction table which used this method, but as this is much simpler we can probably use `op.execute` as you suggest - will test and update!

core/controllers/get_filters.py
```diff
@@ -66,19 +66,27 @@ def get_outcome_categories():
...
     """
-    projects = Project.query.all()
-
-    itl_regions = set(region for project in projects for region in project.itl_regions)
+    geospatial_itl1_regions = (
+        GeospatialDim.query.order_by(GeospatialDim.data_blob["itl1_region_name"])
+        .distinct(GeospatialDim.itl1_region_code, GeospatialDim.data_blob["itl1_region_name"])
+        .with_entities(GeospatialDim.data_blob, GeospatialDim.itl1_region_code)
+        .where(GeospatialDim.projects != None)  # noqa
+        .all()
```
samuelhwilliams, line 79: Do you mind sharing some context around this data blob? Why do we have `itl1_region_code` as a column but `itl1_region_name` as a key inside a jsonb field? Do we expect to expand this slightly, have different data in the future, etc?

db/migrations/versions/028_project_geospatial_m2m.py
MarcUsher: No but likely a lot - it'll create a row for every distinct postcode prefix a project has in its postcode list, for every project. I'll try to check on test to do a count of how many that would be on test and report back.

openapi/api.yml
```diff
@@ -302,20 +302,29 @@ paths:
...
+                  properties:
+                    name:
+                      type: string
+                    id:
+                      type: string
+                example:
+                  - name: "North West"
+                    id: "TLD"
+                  - name: "Scotland"
+                    id: "TLM"
```
samuelhwilliams, line 327: I can't see anywhere yet where we call this API endpoint externally - so we don't need to worry about backwards compatibility here, right?

db/migrations/versions/028_project_geospatial_m2m.py
samuelhwilliams: It'd be useful to know. If it's semi significant then good practice would be to run it as a separate migration+deployment.

core/controllers/get_filters.py
MarcUsher: It's detailed in a bit more info on the ticket which clarified which should go into the data blob but it's part of moving a lot of the event/non-indexed data to data blobs that Noel and Joel have done over the last few months to make tables more extendable. Here, `itl1_region_code` is a column and indexed because it's what used when someone applies a region filter on Find, but the `itl1_region_name` can go in the data blob as it won't be searched in that way. We can also add more data to the data blob if and when we want more data that matches a postcode prefix.

openapi/api.yml
MarcUsher: No exactly - it existed before but wasn't being called by the Find front end, which just populated the Region checkboxes from a hard-coded list. This API update is to make that checkbox list dynamic based on what's in the database. There was a unit test for this endpoint as well which I've updated as part of this PR to match how this now returns the data.

core/controllers/get_filters.py
```diff
@@ -66,19 +66,27 @@ def get_outcome_categories():
...
 
-    :return: a list of itl region codes
+    :return: a list of ITL1 region codes and region names in dictionaries
     """
-    projects = Project.query.all()
-
-    itl_regions = set(region for project in projects for region in project.itl_regions)
+    geospatial_itl1_regions = (
+        GeospatialDim.query.order_by(GeospatialDim.data_blob["itl1_region_name"])
+        .distinct(GeospatialDim.itl1_region_code, GeospatialDim.data_blob["itl1_region_name"])
```
nsnape, line 76: I think we discussed this today, but just in case, I think we said, to that if we move itl_region_name to be a column rather than something hidden in the jsonb, then we can make this query much simpler / clearer (As won't need to specify which fields go in the distinct clause - it's also not very clear from this query what fields we use for what)? But I think we said to do in a separate PR / migration?

core/controllers/get_filters.py
```diff
@@ -66,19 +66,27 @@ def get_outcome_categories():
     return outcome_categories, 200
 
 
-def get_regions():
-    """Returns all unique itl region codes associated with projects.
+def get_geospatial_regions():
+    """Returns all unique ITL1 region codes associated with projects.
 
-    :return: a list of itl region codes
+    :return: a list of ITL1 region codes and region names in dictionaries
```
nsnape, line 72: I know this might be pedantic, but could you update the docstring to match https://github.com/communitiesuk/funding-service-design-post-award-data-store/blob/c17a8c7680f9a1469b94ea49ffb0337d8ef583ab/core/controllers/get_filters.py#L37 as really it does slightly more than just return a list of dicts.

core/controllers/get_filters.py
```diff
@@ -66,19 +66,27 @@ def get_outcome_categories():
...
+        .where(GeospatialDim.projects != None)  # noqa
+        .all()
+    )
 
-    if not itl_regions:
+    if not geospatial_itl1_regions:
         return abort(404, "No regions found.")
 
-    return list(itl_regions), 200
+    itl_regions = [
```
nsnape, line 85: is this supposed to return a list of simple one-line dicts? Like a JSON? Why not just a single dict containing all the lines? with key = name and val = code? or is this what the frontend code needs? If keeping the former, maybe update the docstring description to reflect this?

tests/db_tests/test_constraints.py
```diff
@@ -119,6 +121,20 @@ def test_output_constraint_project_xor_programme(seeded_test_client_rollback):
         db.session.commit()
 
 
+def test_project_geospatial_association(seeded_test_client_rollback):
```
nsnape, line 124: I think this is the wrong fixture? This one intended for non-committed changes, but in this case you do commit, so it's in danger of leaking into other test scopes.

nsnape (changes requested):

tests/db_tests/test_constraints.py
samuelhwilliams: Hm, this sounds inherently risky. Each test case should be fully isolated and rolled back by default. I don't think any test case should be able to commit data that lasts beyond the test run - unless we very explicitly expect the test database to be in a messy state and tests are built specifically to handle that.

db/migrations/versions/028_project_geospatial_m2m.py
MarcUsher: Running `op.execute` on the SQL works just fine so will update to make consistent with previous migrations.

core/db/entities.py
```diff
@@ -438,6 +473,9 @@ class Project(BaseModel):
     outcomes: Mapped[List["OutcomeData"]] = sqla.orm.relationship(back_populates="project")
     risks: Mapped[List["RiskRegister"]] = sqla.orm.relationship(back_populates="project")
     programme_junction: Mapped["ProgrammeJunction"] = sqla.orm.relationship(back_populates="projects")
+    geospatial: Mapped[List["GeospatialDim"]] = sqla.orm.relationship(
```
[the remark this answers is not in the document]
MarcUsher: Makes a lot of sense, thanks for the note Noel, will update and bear this in mind going forward - agree it's much more readable and contextual like this.

core/controllers/get_filters.py
MarcUsher: Yep a separate PR I think just for tidyness, I'll add a ticket for this and add a TODO referencing the ticket number

core/db/entities.py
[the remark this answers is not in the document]
samuelhwilliams: If I can pipe in I might go for `geospatial_dims` or `geospatial_dimensions` - I guess my preference is that it closely follows the name of the table/model. I think having consistent language makes it easier to intuit.

core/controllers/get_filters.py
MarcUsher: Yeah it returns a list of dictionaries with the ITL1 region code as id and region name as name, eg. `[{"id": "TLI", "name": "London"}, {"id": "TLN", "name": "Northern Ireland"}]` which is what the frontend needs to iterate through to create a checkbox for each element in the list and grab the ID when the user selects a checkbox.

Will update the docstring as per your other comment to make reflect this (I think this old get filter was never used and as it was implemented it would just return a list of the unique ITL1 region codes in use, but the front-end still used its own hardcoded list and this endpoint was never called).

tests/db_tests/test_constraints.py
MarcUsher: `test_client_reset` says in the docstring that it's intended for tests that need an empty DB and need to commit DB changes as part of their execution, and as it's function scoped it then resets all the DB tables to empty. I copied that setup for this new fixture and just added the geospatial data seeding at the top, so it should reset the db when the test ends.

`test_client_reset` is the one that many of the other integration tests that commit data to the DB used and needed to be updated to this fixture so they can make use of the seeded geospatial table, but as it's function scoped and resets any changes after the test there shouldn't be any leaking? Or have I misunderstood the fixtures (which is entirely possible!)?

core/util.py
```diff
@@ -13,10 +13,28 @@
...
+    # TODO: FMD-241: Do we need to handles regex failures here, if all postcodes have already passed validation?
+    postcodes_prefix_set = set()
+    for postcode in postcodes:
+        postcode = postcode.strip()
+        postcode_area_matches = re.match(POSTCODE_PREFIX_REGEX, postcode)
+        postcode_prefix = postcode_area_matches.group(0)
+        postcodes_prefix_set.add(postcode_prefix.upper())
+    return postcodes_prefix_set
+
+
```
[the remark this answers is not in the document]
MarcUsher: Will check the front-end won't fall over if they're removed before the front-end change is merged - if it's fine then I'll remove it all as part of this PR, unless you feel that's dangerous and we should only deprecate once the new functionality is deployed?

core/util.py
```diff
@@ -13,10 +13,28 @@
...
 
 
+def get_postcode_prefix_set(postcodes: list[str]) -> set[str]:
+    """
+    Transform a list of postcodes into set of distinct postcode prefixes.
+
+    :param postcodes: A list of strings representing a UK postcode.
+    :return: A set of distinct postcode prefixes.
+    """
+    # TODO: FMD-241: Do we need to handles regex failures here, if all postcodes have already passed validation?
```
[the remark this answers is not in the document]
MarcUsher: It throws an Attribute error when it hits `postcode_prefix = postcode_area_matches.group(0)` as `postcode_area_matches` is None at that point. At a function level I feel like it would be good to handle it, but in reality this function will only ever get passed a list of postcodes that have already passed validation and so have passed a much more stringent regex that checks they have the correct start (plus everything else), and the regex here is more just to grab the first one or two characters that are A-Z - not sure if adding a catch here is good for safety or over thinking it?

core/db/queries.py
```diff
@@ -24,6 +24,26 @@ def filter_on_regions(itl_regions: set[str]) -> list[UUID]:
...
+    Query DB and return all project UUIDs filtered on region filter params.
+
+    :param itl_regions: List of ITL regions to filter on.
+    :return: List of Project id's filtered on region params.
+    """
+    # TODO FMD-241 - can be removed if simpler filtering on base query works
+
+    results = (
+        ents.Project.query.filter(
+            any_(ents.Project.geospatial.itl1_region_code.in_(itl_regions) for geospatial in ents.Project.geospatial)
```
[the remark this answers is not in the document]
MarcUsher: Function will be removed as it's not needed anymore

db/migrations/versions/028_project_geospatial_m2m.py
nsnape: Is there a risk of race condition here if we separate the migration and deployment? ie the db migrates first and migrates it's existing relations. Meanwhile some submits go through the system before we update the codebase - then these will be lost as they will not map many to many until the code is up-to-date?
Or did I misunderstand the suggestion?

tests/db_tests/test_constraints.py
nsnape: Sam, I think we're on the same page here, I maybe just didn't explain clearly:
- We have test fixtures that completely reset the DB at a function level, to be used if any DB changes are committed in the test.
- We also have test fixtures that rollback uncommitted changes in a test - not intended to be used anywhere where commit is used. This is to prevent the complete reset of DB for every test where it's not always necassary.

In this case, my comment was saying that I think Marc used a test fixture that only rolls back uncommitted changes, but the test does a commit, the fixture #I could see in use was `seeded_test_client_rollback` but should have bee none at function scope for resetting the DB.

MarcUsher: Hey both, yep you're both right, I've used the wrong fixture here - will update it to use reset instead of rollback. Thanks for catching!

samuelhwilliams: I'm a bit torn. It feels like a chunk of mental overhead to have to work out whether the code under test is calling a `commit` at any point, and unless we have a very clear definition of when commits happen, feels again like something that could be refactored in the future and then change the semantics of the rest without anyone really spotting it.

Generally I'd probably want every test, by default, to rollback any changes that were made during the test - 'committed' or not.

If speed is a concern we can easily start looking at parallelising them (and always better to do that sooner rather than later so that you catch structural issues with concurrent test runes early).

samuelhwilliams (commented):
oops had some pending comms, all fairly mild

core/db/entities.py
```diff
@@ -107,6 +121,27 @@ class FundingQuestion(BaseModel):
...
 
 
+class GeospatialDim(BaseModel):
+    """Stores Geospatial information mapped to postcodes."""
+
+    __tablename__ = "geospatial_dim"
+
+    postcode_prefix = sqla.Column(sqla.String(length=4), nullable=False, unique=True)
+    itl1_region_code = sqla.Column(sqla.String(), nullable=False, unique=False)
+    data_blob = sqla.Column(JSONB, nullable=True)
```
samuelhwilliams, line 131: I don't think we ever expect this to be null, right? Should we enforce this?

core/db/entities.py
```diff
@@ -12,6 +12,20 @@
...
+    ),
+    db.Column(
+        "geospatial_id",
+        GUID(),
+        db.ForeignKey("geospatial_dim.id", ondelete="RESTRICT"),
+        nullable=False,
+        primary_key=True,
+    ),
+)
+
```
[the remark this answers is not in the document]
MarcUsher: Mainly because it wasn't a proper Model class like the others in the file so felt odd to stick it amongst the rest - but happy to move it to sit alphabetically in the list/sit it next to projects or geospatial if that's better practice?

core/db/entities.py
```diff
@@ -107,6 +121,27 @@ class FundingQuestion(BaseModel):
...
+    projects: Mapped[List["Project"]] = sqla.orm.relationship(
+        back_populates="geospatial_dims", secondary=project_geospatial_association
+    )
+
+    __table_args__ = (
+        sqla.Index(
+            "ix_geospatial_dim_filter_region",
+            "itl1_region_code",
+        ),
+    )
```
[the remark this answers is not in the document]
MarcUsher: I think the latter - Noel can maybe confirm but here I went for consistency with the other column indexes that are used for frontend filtering.

core/db/entities.py
MarcUsher: Hmmm yes this is a really good point, we should enforce it! Will update.

core/db/entities.py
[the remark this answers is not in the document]
samuelhwilliams: It's nice to co-locate things together, I think.

core/db/entities.py
MarcUsher: Oh I'm going to remove the data_blob anyway in a separate PR as we discussed with Noel, to move the `itl1_region_name` to its own column, but yes probably good practice to enforce this in this PR even if we're going to remove it

samuelhwilliams: oh right yep 😂

core/controllers/get_filters.py
MarcUsher: As discussed, `itl1_region_name` will be moved out of the datablob and into its own column as part of a separate ticket (FMD-275 in the backlog)

core/db/entities.py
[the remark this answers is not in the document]
MarcUsher: Have co-located to above `GeospatialDim` - the association table needs to be defined before it can be used in either `GeospatialDim` or `Project`.

tests/controller_tests/test_filters.py
```diff
@@ -197,16 +197,23 @@ def test_get_outcome_alphabetical_sorting(seeded_test_client):
...
     assert response.status_code == 200
     response_json = response.json
 
-    assert response_json
-    assert all(isinstance(region, str) for region in response_json)
+    assert all("name" in region for region in response_json)
+    assert all(isinstance(region["name"], str) for region in response_json)
+
+    assert all("id" in region for region in response_json)
+    assert all(isinstance(region["id"], str) for region in response_json)
```
samuelhwilliams, line 212: Somewhat redundant given we're doing a full assert on these things below 😅  but all good

samuelhwilliams (approved):
I think this is good to get in 🎉  Well done for sticking with this beast

Joel-Pearce (approved):

tests/serialisation_tests/test_serialisation.py
```diff
@@ -317,22 +317,21 @@ def test_serialise_data_region_filter(seeded_test_client, additional_test_data):
     when ITL region is passed, projects should be filtered by ITL region and any parent programmes with entirely
     filtered out child projects should not be returned.
     """
-    itl_regions = {ITLRegion.SouthWest}
-    test_query_region = download_data_base_query(itl_regions=itl_regions)
+    itl1_regions = [GeospatialDim.query.filter(GeospatialDim.postcode_prefix == "BS").one().itl1_region_code]
+    test_query_region = download_data_base_query(itl1_regions=itl1_regions)
```
nsnape, line 321: would this be clearer to read if you didn't instantly put the itl1_regions var into a one item list? As then you can set this param like `itl_regions=[itl_regions]`, and then later in the code you don't need to 0 index itl_regions to get the entity?

nsnape (changes requested):

nsnape (approved):

PR #510, opened 2024-03-27
SMD-702 - Validating reporting period change against 'Actual, forecast or cancelled' field

wjrm500:
### Ticket
[SMD-702 - Pathfinders - validate actual / forecast + reporting period in Project Finance Changes table](https://dluhcdigital.atlassian.net/browse/SMD-702)

### Change summary
- This PR is to add validation to check that the values for "Reporting period change takes place" and "Actual, forecast or cancelled" in the "Project finance changes" table align
- In order to do this, we need to reference the reporting period specified for the submission as a whole (newly extracted from the "Start" worksheet) to compare to the "Reporting period change takes place"
- For example, if the reporting period for the submission as a whole is "Q4 2023/24: Jan 2024 - Mar 2024" and we have a row in the "Project finance changes" table where the "Reporting period change takes place" is also "Q4 2023/24: Jan 2024 - Mar 2024" but "Actual, forecast or cancelled" is "Forecast", then this does not make sense and we should raise an error
- Or for another example, if the reporting period for the submission as a whole is "Q4 2023/24: Jan 2024 - Mar 2024" and we have a row in the "Project finance changes" table where the "Reporting period change takes place" is "Q1 2024/25: Apr 2024 - Jun 2024" but "Actual, forecast or cancelled" is "Actual", then this does not make sense and we should raise an error

- ALSO, not technically in scope of this ticket but I also renamed a couple of variables in `extracted_data.py` for clarity and to avoid name clashes


core/validation/specific_validation/pathfinders/round_1.py
```diff
@@ -186,3 +188,38 @@ def _check_credible_plan_fields(extracted_table_dfs: dict[str, pd.DataFrame]) ->
                         )
                     )
     return error_messages
+
+
+def _check_actual_forecast_reporting_period(extracted_table_dfs: dict[str, pd.DataFrame]) -> list[Message]:
+    reporting_period = extracted_table_dfs["Reporting period"].iloc[0, 0]
+    submission_reporting_period_start_date = PF_REPORTING_PERIOD_TO_DATES_2[reporting_period]["start"]
```
nsnape, line 195: do we not already know the submission start and end dates via this bit of code:

https://github.com/communitiesuk/funding-service-design-post-award-data-store/blob/9065c5107bc273ed8b5b86ccf8b0e6b81eeb5c41/core/transformation/pathfinders/round_1/transform.py#L68

nsnape: ie, the dates should be directly accessible from `extracted_table_dfs["Reporting Period"]` in the line above?

core/validation/specific_validation/pathfinders/round_1.py
```diff
@@ -186,3 +188,38 @@ def _check_credible_plan_fields(extracted_table_dfs: dict[str, pd.DataFrame]) ->
...
+    pfcs_df = extracted_table_dfs["Project finance changes"]
```
nsnape, line 196: is this specifically for PFC's tab? Do you know if we also do this validation (Actual/forecast) for the other tabs such as Funding / Outcomes? I'm not 100% sure if we even do need to validate these other tabs, but if we do, is there a separate ticket, or should we reuse this func to do it?

nsnape (changes requested):

core/validation/specific_validation/pathfinders/round_1.py
wjrm500: The bit of code you're referencing is part of transformation which comes after cross-table validation - `extracted_table_dfs` is just a dictionary containing tables extracted directly from the spreadsheet and minimally processed. The flow is:

- Initial validation (high-level checks on auth etc.)
- Table extraction (see `extract_process_validate_tables` in `ingest.py`)
- Table processing (see `extract_process_validate_tables` in `ingest.py`)
- Table validation (see `extract_process_validate_tables` in `ingest.py`)
- Cross-table validation (this file being commented on)
- Transformation
- Database load

core/transformation/pathfinders/consts.py
```diff
@@ -51,3 +52,43 @@
         "end": None,
     },
 }
+
+# Maps Pathfinder reporting periods from "Reporting period" and "Project finance changes" tables to start and end dates
+PF_REPORTING_PERIOD_TO_DATES_2 = {
```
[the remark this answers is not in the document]
wjrm500: Happy to change but not sure what do as both mappings are financial quarters to start and end dates, the financial quarters are just written in different ways e.g., "Financial year 2023 to 2024, (Jan to Mar)" in the first mapping versus "Q4 2023/24: Jan 2024 - Mar 2024" in the second mapping - any suggestions?

core/validation/specific_validation/pathfinders/round_1.py
wjrm500: Yes specifically for PFCs - no strong need to validate for "Outputs", "Outcomes" and "Forecast and actual spend" tables as "Actual" versus "Forecast" in the column headers is not specified by the user, rather hard-coded. Here we require validation because the user selects "Actual", "Forecast" or "Cancelled" for "Actual, forecast or cancelled" field, and then selects an available reporting period for "Reporting period change takes place", and we need to make sure those align (the spreadsheet does not do this for us)

core/transformation/pathfinders/consts.py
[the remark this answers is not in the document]
nsnape: ah, I didn't look at the values properly :D yeah okay, just 2 different ways of saying the same thing in the form there.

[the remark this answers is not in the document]
nsnape: Could we just bung them all in one dict?

core/validation/specific_validation/pathfinders/round_1.py
wjrm500: Thinking about this some more, because the "Actual" and "Forecast" values in the "Outputs", "Outcomes" and "Forecast and actual spend" tables actually form part of the column headers, they are effectively validated already - e.g., if the user hypothetically modified the value "Forecast" for "Financial year 2024/25, (Apr to Jun)" to "Actual", then the calculated column header expected for the template "Financial year 2024/25, (Apr to Jun), Forecast" would not appear (as we would end up with "Financial year 2024/25, (Apr to Jun), Actual" instead, and the validation would fail at `_check_columns` (see `TableValidator` class)

core/transformation/pathfinders/consts.py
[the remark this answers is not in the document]
nsnape: ok yeah, change as you suggested on slack :)

nsnape (approved):

PR #521, opened 2024-04-04
update ProgrrammeManagementFunding tab name in serialised extract

nsnape:
### Change description
Changing name in data extract for this table due to request from Towns Fund (currently they are they are the only user of this tab). Request copied from teams channel TF Hotline, (Renata Walton, 12:45, 04/04/2024):

_"Hi 
Gideon_
 
_Goldberg
, without fully checking the contents as I'll need to pull this into out dashboard. Thank you all for finally getting this information in there. It's actually ProgrammeManagementFunding, as it's a specific type of payment which is made to LAs for PM activity. So can the name be changed?"_

### DIFF
--- a/core/const.py
+++ b/core/const.py
@@ -700,7 +700,7 @@ TABLE_SORT_ORDERS = {
     "RiskRegister": ["SubmissionID", "ProgrammeID", "ProjectID", "RiskName"],
     "ProjectFinanceChange": ["SubmissionID", "ProgrammeID"],
     "SubmissionRef": ["SubmissionID"],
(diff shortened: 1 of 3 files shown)


samuelhwilliams (commented):
Do we want to do a rename on the `ProgrammeFundingManagement` class -> `ProgrammeManagementFunding` as well?

We don't necessarily have to change the db table name. Maybe it's less desirable to have them out of sync? Up to you. (I think I'd prefer renaming so that a naive search will show up the results. We explicitly set the table name so it's discoverably different)

Joel-Pearce (approved):

samuelhwilliams (approved):

PR #523, opened 2024-04-05
FMD-259: Add DB constraint to check date ranges

samuelhwilliams:
### Change description
Add constraints to our DB tables which have start/end dates, ensuring that end dates can't be before start dates.

Includes a couple of extra commits which tweak config so that:
* SQL is logged when migrations run, for information/records
* Migrations run in separate transactions (see 586359ca9d77cf746beadb309815035fd368573c for more info)

### How to test

### DIFF
--- a/db/migrations/versions/027_start_date_before_end.py
+++ b/db/migrations/versions/027_start_date_before_end.py
@@ -1,0 +1,56 @@
+"""require start_dates to be before end_dates, if both fields are present in a row
+
+Revision ID: 027_start_date_before_end
+Revises: 026_standardise_actual_forecast
+Create Date: 2024-04-04 13:59:15.747254
+
+"""
+
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision = "027_start_date_before_end"
+down_revision = "026_standardise_actual_forecast"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    for table in ["funding", "programme_funding_management", "project_progress"]:
+        op.create_check_constraint(
+            "start_before_end",  # gets prefixed with `ck_{table}`
+            table,
+            "start_date IS NULL OR end_date IS NULL OR (start_date <= end_date)",
+            postgresql_not_valid=True,
+        )
+
+    for table in ["outcome_data", "output_data"]:
+        op.create_check_constraint(
+            "start_before_end",  # gets prefixed with `ck_{table}`
+            table,
+            "(start_date <= end_date)",
+            postgresql_not_valid=True,
+        )
+
+    for table in ["submission_dim"]:
+        op.create_check_constraint(
+            "start_before_end",  # gets prefixed with `ck_{table}`
+            table,
+            "(reporting_period_start <= reporting_period_end)",
+            postgresql_not_valid=True,
+        )
+
+
+def downgrade():
+    for table in [
+        "funding",
+        "programme_funding_management",
+        "project_progress",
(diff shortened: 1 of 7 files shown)


db/migrations/versions/028_validate_check_constraints.py
```diff
@@ -0,0 +1,31 @@
...
+def upgrade():
+    for table in [
+        "funding",
+        "programme_funding_management",
+        "project_progress",
+        "outcome_data",
+        "output_data",
+        "submission_dim",
+    ]:
+        op.execute(f"ALTER TABLE {table} VALIDATE CONSTRAINT ck_{table}_start_before_end")
```
Joel-Pearce, line 27: Have you deployed this branch to the dev environment to see what happens when we run this? Dev is very close to Prod in terms of what's in the database so would be a good reflection of whether this will validate successfully or not

db/migrations/versions/027_start_date_before_end.py
```diff
@@ -0,0 +1,56 @@
...
+            postgresql_not_valid=True,
+        )
+
+    for table in ["submission_dim"]:
+        op.create_check_constraint(
+            "start_before_end",  # gets prefixed with `ck_{table}`
+            table,
+            "(reporting_period_start <= reporting_period_end)",
+            postgresql_not_valid=True,
+        )
```
Joel-Pearce, line 41: Just a thought, but won't this blow-up if the constraint is not met somewhere in our existing data? What's therefore the need to have a separate validate script for checking the constraint?

Joel-Pearce (changes requested):
Do we need a separate script for validating the constraint, or can we not just run that as bespoke SQL before making this change, or include it in the same migration file?

db/migrations/versions/028_validate_check_constraints.py
samuelhwilliams: Just did - and it blew up as you perhaps suspected ;)

Will chat with Gideon to work out how we find+fix the dodgy data

db/migrations/versions/027_start_date_before_end.py
samuelhwilliams: The `postgresql_not_valid=True` bits mean that the constraint goes in but ignores any data that already exists which isn't valid. The constraint is only validated on new data.

The second migration then validates the constraint and blows up on bad data. It being separate means:

* we keep the constraint which validates _new_ data
* each migration takes a less heavy-handed lock, keeping database queries less restricted during migration

db/migrations/versions/028_validate_check_constraints.py
nsnape: The main problem with putting  this into 2 migrations is that if on Prod the 3nd one fails, then we're no better off than if they were just one and it all failed, as basically our migrations will be out of sync with code. But if all in one migration, then it will all roll back atomically if there is an issue

nsnape (changes requested):

samuelhwilliams:
I've removed the second migration (which validated the constraints) from this for now, after discovering some bad data in our DB that wouldn't validate. The second migration is now in a separate PR (https://github.com/communitiesuk/funding-service-design-post-award-data-store/pull/524) and we'll merge that after fixing the bad data.

We can still merge this PR with just the first migration, which will add the 'NOT VALID' constraints, protecting us against more bad data entering those tables but doing nothing about existing data.

Joel-Pearce (approved):
LGTM

PR #525, opened 2024-04-05
SMD-748: Validate TF project dates

MarcUsher:
### Change description
- Add schema validation check to make sure project start dates are not later than project completion dates, for TF Rounds 3, 4 and 5
- Add new validation error message for invalid project dates
- Add unit tests for new validation function

### How to test
- Amend a mock TF Round 3 or 4 spreadsheet from the repo and set project start dates to something later than the project completion date (or vice versa)
- Upload the spreadsheet to submit
- See that validation fails and errors are returned to the frontend

### DIFF
--- a/core/controllers/ingest_dependencies.py
+++ b/core/controllers/ingest_dependencies.py
@@ -66,6 +66,7 @@ def ingest_dependencies_factory(fund: str, reporting_round: int) -> IngestDepend
                 transform_data=ingest_round_three_data_towns_fund,
                 validation_schema=TF_ROUND_3_VAL_SCHEMA,
                 initial_validation_schema=TF_ROUND_3_INIT_VAL_SCHEMA,
+                messenger=TFMessenger(),
                 table_to_load_function_mapping=get_table_to_load_function_mapping("Towns Fund"),
             )
         case ("Towns Fund", 4):
--- a/core/validation/schema_validation/validate.py
+++ b/core/validation/schema_validation/validate.py
@@ -10,6 +10,7 @@ from datetime import datetime
 import pandas as pd
 from numpy.typing import NDArray
 
+from core.messaging.tf_messaging import TFMessages as msgs
 from core.validation.failures import internal, user
 from core.validation.utils import is_blank, remove_duplicate_indexes
 
@@ -81,6 +82,7 @@ def validations(
         (validate_unique_composite_key, "composite_key"),
         (validate_enums, "enums"),
         (validate_nullable, "non-nullable"),
+        (validate_project_dates, "project_date_validation"),
     )
 
     validation_failures = []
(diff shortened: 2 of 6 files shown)


core/controllers/ingest_dependencies.py
```diff
@@ -66,6 +66,7 @@ def ingest_dependencies_factory(fund: str, reporting_round: int) -> IngestDepend
                 transform_data=ingest_round_three_data_towns_fund,
                 validation_schema=TF_ROUND_3_VAL_SCHEMA,
                 initial_validation_schema=TF_ROUND_3_INIT_VAL_SCHEMA,
+                messenger=TFMessenger(),
```
samuelhwilliams, line 69: Is this just for consistency or is it doing something more real than I imagine?

My naive understanding I suppose now is that rounds 3 and 4 are closed and done and dead and that these would never be hit? Am I just totally wrong there?

MarcUsher: Round 5 uses the same `validation_schema` as Round 4 (as the spreadsheet is the same one), and we sometimes reingest spreadsheets from old rounds in order to tidy up data issues or for data to fit into DB changes - in this case those few spreadsheets you, Gideon and Noel found that have start dates after completion dates will be reingested, so including the validation check for R3 is just to make sure we haven't missed any incorrect dates when those are reingested (this was specified in the JIRA ticket).

The reason for the messenger specifically being added to Round 3 here is that the `user.GenericFailure` validation error that the validation generates needs the messenger to generate the error messages - it's there in r4 and none of the r3 errors used this Class so needed to add the messenger here.

core/validation/schema_validation/validate.py
```diff
@@ -368,3 +370,36 @@ def validate_nullable(
                 )
 
     return non_nullable_constraint_failure
+
+
+def validate_project_dates(
+    data_dict: dict[str, pd.DataFrame], table: str, project_dates: list[str]
```
wjrm500, line 376: `project_date_cols` or something similar may be a more accurate argument name here

wjrm500 (approved):
Hey Marc, all looks good and testing manually via local API upload works as expected.

A couple of notes:

- Maybe this new project date validation would have been better placed in fund-specific validation, as it is currently sat alongside some very generic, universal validation logic (uniqueness, nullability etc.) where it feels slightly misplaced
- Using `@pytest.mark.parametrize` for the tests might have been more concise, allowing you to combine the test functions and pass a list of individually-evaluated scenarios

Approving anyway as these are just quibbles and don't affect functionality :)

core/controllers/ingest_dependencies.py
samuelhwilliams: Nice, thanks for explaining 👍

PR #526, opened 2024-04-08
SMD-711 - Pathfinders - update enums or move allowed value checks to cross-table validation

bobbi-hbrown:
### Ticket
[Pathfinders - update enums or move allowed value checks to cross-table validation](https://dluhcdigital.atlassian.net/browse/SMD-711)

### Change summary
- **Replacing hard-coded enum checks with dynamic cross-table checks** - At the moment there are lists of values that are used in the spreadsheet to populate the dropdowns that the user interacts with, and we effectively replicate those lists on the backend for the purposes of validation. However, this approach means that we have two conflicting sources of truth and can lead to misalignment if not maintained properly. This change removes the hard-coded checks that are performed at schema validation time, replacing the functionality of each with the addition of corresponding cross-table validation checks. These checks validate data against the values extracted from the hidden worksheet 'Dropdown Values' in the spreadsheet, creating it as a single source of a truth.

- **Updating the cross-table validation failures spreadsheet to demo new checks** - This change updates the test spreadsheet for cross-table validation errors to include errors for RAG rating, risk category and risk score. The corresponding integration test is updated accordingly.

- **Refactoring cross-table validation check _check_projects** - The new function _dropdown_error_messages can be used by _check_projects to make it more concise, as _check_projects currently contains bespoke code almost identical to the code used in _dropdown_error_messages, around getting the breaching names based on the indices. In order to achieve this, the control mapping _programme_id_to_project_ids has been replaced by _programme_to_projects, so that we don't need to worry about mapping back and forth and making things more convoluted. Control mappings only exist as a convenience for cross-table validation and transformation, so changing them to reduce complexity in those stages is fine.


core/validation/specific_validation/pathfinders/round_1.py
```diff
@@ -371,3 +376,92 @@ def _check_actual_forecast_reporting_period(extracted_table_dfs: dict[str, pd.Da
...
+    Check that values in a DataFrame column belong to a list of allowed values.
+    """
+    breaching_row_indices = df[~df[value_column].isin(allowed_values)].index.tolist()
+    breaching_values = df.loc[breaching_row_indices, value_column].tolist()
+    return [
+        _error_message(
+            sheet=sheet,
+            section=section,
+            description=error_description.format(value=value),
+        )
```
Joel-Pearce, line 394: I think we can return the  ```cell_index``` alongside this information given we already have the ```value_column``` and can therefore know the letter of the column in the Excel sheet, and all we need to do is increment the indices by one to get the row. Hopefully it's not more effort than that

Joel-Pearce (changes requested):

core/validation/specific_validation/pathfinders/round_1.py
bobbi-hbrown: Fair point, I was going to make a separate ticket for this but makes more sense to add it here if it's a simple change

core/table_configs/pathfinders/round_1.py
```diff
@@ -803,9 +777,7 @@ class PFErrors:
                 "Actual, forecast or cancelled": pa.Column(
                     str, pa.Check.isin(PFEnums.ACTUAL_FORECAST, error=PFErrors.ISIN)
                 ),
-                "Reporting period change takes place": pa.Column(
-                    str, pa.Check.isin(PFEnums.REPORTING_PERIOD, error=PFErrors.ISIN)
-                ),
+                "Reporting period change takes place": pa.Column(str),
```
wjrm500, line 780: The enum check has been removed but not replaced by any check in cross-table validation - although a reporting period check has been added there, it only checks the reporting period in the "Reporting period" table of the "Start" worksheet, not this one

tests/integration_tests/test_ingest_component_all_funds.py
```diff
@@ -118,7 +118,7 @@ def test_multiple_rounds_multiple_funds_end_to_end(
             ],
         }
     )
-
+    print(df_dict["ProgrammeRef"].to_string())
```
wjrm500, line 121: Reminder to remove when finalised

wjrm500 (changes requested):
Hey @bobbi-hbrown, from the code in the PR alone (forgetting the issues with the test spreadsheet), I can see we're almost there so well done :)

I've requested a few changes, some more important than others - only one or two critical things in there though. The other thing I noticed was that this changeset includes some changes unrelated to the task e.g., the addition of the `PHONE_NUMBER` constant in `PFErrors` - it'd be good to fix this up so we only see the true diff here

Let me know if you need any help and have a good evening!

core/table_configs/pathfinders/round_1.py
```diff
@@ -111,6 +65,13 @@ class PFErrors:
     CREDIBLE_PLAN_YES = "If you have selected 'Yes' for 'Credible Plan', you must answer Q2, Q3 and Q4."
     CREDIBLE_PLAN_NO = "If you have selected 'No' for 'Credible Plan', Q2, Q3 and Q4 must be left blank."
     INTERVENTION_THEME_NOT_ALLOWED = "Intervention theme '{intervention_theme}' is not allowed."
+    ACTUAL_REPORTING_PERIOD = "Reporting period must not be in future if 'Actual, forecast or cancelled' is 'Actual'."
```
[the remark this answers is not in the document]
MarcUsher: Same with `FORECAST_REPORTING_PERIOD` and `PHONE_NUMBER` - they both have their correct values on lines 78 and 81. I removed these duplicates in a small PR the other day so might just need a rebase.

wjrm500:
@samuelhwilliams pointed out this morning that users are able to bypass password protection on hidden sheets (including "Dropdown Values") by opening the workbook in Apple Numbers or OpenOffice Calc, which means that this approach to validation is too flimsy. Having a single source of truth between front and backend would be ideal, but the data would need to exist in a truly secure environment that we control.

PR #528, opened 2024-04-08
BAU: fix na as none issue in finances tab

Joel-Pearce:
Fixes a bug where in a user inputting "N/A" was being treated as a blank value and raising a validation error where one shouldn't occur.

Pandas' ```read_excel``` treats "N/A" as a NaN value, which is where the issue was occurring, but because of this it's difficult to write a unit test as we cannot easily mock an Excel file. As such, this PR changes one of the values in ```PF_Round_1_Success.xlsx``` to be "N/A" in a place where this should not be treated as blank.


samuelhwilliams (approved):
Change looks good but (as discussed) be good to get QA over this as the impact on pandas processing could be quite wide.

samuelhwilliams:
Noel would be sad `Merge branch 'main' into BAU/fix-na-bug-in-funding-template` 😂

PR #530, opened 2024-04-10
SMD-759: bug fix postcode validation

samuelhwilliams:
### Change description
Fixes two things:

1) When uploading a spreadsheet with null data and invalid postcodes in the postcode columns, only the null data is reported. (If each failure state is uploaded separately they are reported - but in combination only one is). Now both failures will be reported up front.
2) If entering invalid postcode data that is a non-string type (eg int, datetime), that would previously be accepted. Now it will be rejected correctly.

---

I would have liked to update an integration test spreadsheet to contain some int/datetime postcodes but I don't have excel, so ... what can you do 🤷  help please?

### How to test
Edit the `tests/integration_tests/mock_pf_returns/PF_Round_1_Success.xlsx` file on the `Project location` tab. Change some of the postcode lists to be numeric eg `1234`, or date eg `10/04/24`, or null eg ``. Upload it via submit and check that all errors are reportedly accurately and in one go.

### DIFF
--- a/tables/checks.py
+++ b/tables/checks.py
@@ -17,15 +17,36 @@ However, there are two main disadvantages of schemas with inline custom checks:
     2. you can’t use them to synthesize data because the checks are not associated with a hypothesis strategy.
 """
 
+import math
 import re
 from datetime import datetime
-from typing import Any
+from typing import Any, Literal
 
 import pandas as pd
 import pandera as pa
 from pandera.extensions import CheckType
 
 from core.transformation.utils import POSTCODE_REGEX
+
+
+def _should_fail_check_because_element_is_nan(element) -> Literal[True] | None:
+    """Detect nan values in pandera checks which should return false to bypass type errors.
+
+    When writing element-wise pandera checks, pandera will pass through any null/nan values rather than skipping over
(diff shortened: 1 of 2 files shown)


tables/checks.py
```diff
@@ -49,39 +71,57 @@ def is_float(element):
...
+def not_in_future(element: Any):
     """Checks that a datetime is not in the future.
 
     :param element: an element to check
     :return: True if passes the check, else False
     """
+    if _should_fail_check_because_element_is_nan(element):
+        return False
+
+    # If it's not come through here as a string, it's probably some other native data type like an int or string,
```
Joel-Pearce, line 83: "some other native data type like an int or **string**"

is this a typo?

Joel-Pearce (changes requested):
Looks good, seems like there's a small typo?

tables/checks.py
samuelhwilliams: oops yeah - we expect it here as a datetime - good spot

samuelhwilliams: fixed

Joel-Pearce (approved):
LGTM

PR #532, opened 2024-04-10
SMD-734 - Combining general and cross-table validation error messages

wjrm500:
### Ticket
[Pathfinders - combine table validation and cross-table validation errors](https://dluhcdigital.atlassian.net/browse/SMD-734)

### Change description
Currently we raise table validation errors and cross-table validation errors separately, meaning that if the user has both in their spreadsheet, they will be met with a new set of validation errors even after resolving all of the validation errors listed originally, which seems like a bad user experience.

This PR combines error messages. Implementation involves not raising `ValidationError` directly from inside `extract_process_validate_tables` or `cross_table_validation`, instead returning lists of error messages from those functions. The original design for implementation here (continuing to raise `ValidationError` in both functions, catching in `ingest.py`, taking error messages out of `ValidationError` objects and using these) does not work as `cross_table_validation` requires the tables themselves, and we wouldn't have access to these if we simply exited `extract_process_validate_tables` with an exception. As part of this work, `coerce_data` had to be pulled out of `extract_process_validate_tables`, because this blows up with validation errors.

I've added a new test spreadsheet containing both general and cross-table validation errors to demonstrate the new functionality working.

Open to suggestions for alternative implementations as aware that returning a tuple of tables and error messages from `extract_process_validate_tables` is perhaps not an ideal design, but please be aware that I will be off from this afternoon until PF go-live date on Monday 15th April, so might be best to consider refactor post-merge to get this functionality into `main` ASAP.


samuelhwilliams (approved):

samuelhwilliams:
I've stuck this in QA for @NarasimhaYanamala - I think chat to him to make sure he's happy with this before merging? Unless the QA/merge process is different than I imagine.

wjrm500:
Hey @samuelhwilliams, looks from the Slack thread like Nara ran some QA on this the other day, will seek confirmation from him now that he is happy to sign off

PR #540, opened 2024-04-15
SMD-773 -  short-term fix for Bespoke Outputs issue with unhandled dropdown placeholder

wjrm500:
### Ticket
[Pathfinders - short-term fix for issue around "You have no bespoke outputs to select"](https://dluhcdigital.atlassian.net/browse/SMD-773)

### Description
Add a simple method to the table processing step that checks the ID tag of the table being processed, and if it corresponds to the ID tag of the Bespoke Outputs table, drops any rows where the "Output" value is "You have no bespoke outputs to select". Does a similar thing for Bespoke Outcomes. Hacky but should prevent the issue generating support tickets from the affected LAs. Idea will be to deprecate this code in Pathfinders round two, as we'll be able to make changes on the client side in advance of that round to prevent this issue arriving to the backend (see SMD-772).


samuelhwilliams (approved):
Tested and can confirm it works 👍  But yeah, let's make sure we get rid of this ASAP... Maybe one to proactively rip out as soon as this pathfinders submission closes so that we know we can't rely on it for round 2?

PR #541, opened 2024-04-15
SMD-769 - Order error cell indexes logically

samuelhwilliams:
https://dluhcdigital.atlassian.net/browse/SMD-773

### Change description

### How to test
* Edit a valid spreadsheet and introduce multiple instances of a specific type of error across the same rows and/or columns.
* On `main` branch, upload the spreadsheet and confirm that the cell references as unordered.
* Checkout this branch, re-upload and confirm that the cell references are now ordered.

#### Before
image

#### After
image

### DIFF
--- a/core/messaging/__init__.py
+++ b/core/messaging/__init__.py
@@ -1,13 +1,19 @@
+import math
+import re
 from abc import ABC, abstractmethod
 
 
 class Message:
     """Generic Message class that defines the components of a validation error message"""
 
+    cell_index_pattern = re.compile(r"^([a-zA-Z]+)(\d+)?$")
+
     def __init__(self, sheet, section, cell_indexes: tuple[str, ...] | None, description, error_type):
         self.sheet = sheet
         self.section = section
-        self.__cell_indexes: None | tuple[str] = cell_indexes
+        # If `cell_indexes` is set, it should be a tuple containing strings that describe either an Excel column
+        # (eg `A`) or an Excel cell (eg `A1`) only.
+        self.__cell_indexes: tuple[str, ...] | None = self.__clean_cell_indexes(cell_indexes)
         self.description = description
         self.error_type = error_type
 
@@ -32,23 +38,51 @@ class Message:
         return getattr(self, item)
 
     @property
-    def cell_indexes(self) -> str:
-        return ", ".join(self.__cell_indexes)
+    def cell_indexes(self) -> tuple[str, ...] | None:
+        return self.__cell_indexes
+
+    @classmethod
+    def __clean_cell_indexes(cls, cell_indexes) -> tuple[str, ...] | None:
+        """Remove any duplicate cell references, sorts them in column-then-row order, and ensures uppercase columns"""
+        if cell_indexes is None:
+            return None
+        if None in cell_indexes:
+            raise ValueError("`cell_indexes` cannot contain None elements; must be None directly")
+        try:
+            # Split cell indexes into column and row parts, ie 'A1' becomes ('A', '1')
+            split_cell_indexes = [cls.cell_index_pattern.match(cell_index).groups() for cell_index in set(cell_indexes)]
+        except AttributeError:
+            # Regex failed to match somehow - returning unsorted is better than dying
+            return cell_indexes
+
+        # Sort cell indexes by column (Excel-wise) then by row (numerically) - being careful that there may not
+        # be a row number.
+        # Sorting columns Excel-wise means that AA comes after Z, not between A and B.
+        sorted_cell_indexes = sorted(
+            split_cell_indexes,
+            key=lambda cell_index: (
+                len(cell_index[0]),
+                cell_index[0].upper(),
+                int(cell_index[1]) if cell_index[1] else -math.inf,
+            ),
+        )
+
+        return tuple(f"{parts[0].upper()}{parts[1] if parts[1] else ''}" for parts in sorted_cell_indexes)
 
     def combine(self, other: "Message"):
         if not isinstance(other, Message):
-            raise NotImplementedError("`other` must be another instance of Message")
+            raise ValueError("`other` must be another instance of Message")
 
         if self.__cell_indexes is None or other.__cell_indexes is None:
-            raise NotImplementedError("Can only combine Message instances if both reference cell indexes")
+            raise ValueError("Can only combine Message instances if both reference cell indexes")
 
-        self.__cell_indexes = self.__cell_indexes + other.__cell_indexes
+        self.__cell_indexes = self.__clean_cell_indexes(self.__cell_indexes + other.__cell_indexes)
 
     def to_dict(self):
         return {
             "sheet": self.sheet,
             "section": self.section,
-            "cell_index": self.cell_indexes,
+            "cell_index": ", ".join(self.cell_indexes) if self.cell_indexes else None,
             "description": self.description,
             "error_type": self.error_type,
         }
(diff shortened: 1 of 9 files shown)


tests/integration_tests/test_ingest_component_towns_fund.py
```diff
@@ -645,7 +645,7 @@ def test_ingest_with_r4_file_hs_file_failure(test_client, towns_fund_round_4_fil
         "title": "Bad Request",
         "validation_errors": [
             {
-                "cell_index": "DNone",
```
samuelhwilliams, line 648: This feels like a strange way to represent just a column without a row number, so have also changed this to strip the "None" and just say "D".

core/messaging/__init__.py
```diff
@@ -4,15 +4,15 @@
 class Message:
     """Generic Message class that defines the components of a validation error message"""
 
-    def __init__(self, sheet, section, cell_index, description, error_type):
+    def __init__(self, sheet, section, cell_indexes: tuple[str, ...] | None, description, error_type):
         self.sheet = sheet
         self.section = section
-        self.cell_index = cell_index
+        self.__cell_indexes: None | tuple[str] = cell_indexes
```
wjrm500: Type hint here is `None | tuple[str]` while type hint for argument `cell_indexes` is `tuple[str, ...] | None`, can we make consistent?

core/messaging/__init__.py
```diff
@@ -1,13 +1,19 @@
+import math
+import re
 from abc import ABC, abstractmethod
 
 
 class Message:
     """Generic Message class that defines the components of a validation error message"""
 
+    cell_index_pattern = re.compile(r"^([a-zA-Z]+)(\d+)?$")
```
wjrm500, line 9: Do we need to allow lowercase letters?

core/messaging/__init__.py
```diff
@@ -32,8 +38,35 @@ def __getitem__(self, item):
         return getattr(self, item)
 
     @property
-    def cell_indexes(self) -> str:
-        return ", ".join(self.__cell_indexes)
+    def cell_indexes(self) -> tuple[str, ...] | None:
```
wjrm500, line 41: Now that `__cell_indexes` is the same as `cell_indexes`, maybe simpler just to have `cell_indexes` as a normal attribute with no `property` method?

tests/messaging_tests/test_messaging.py
```diff
@@ -271,3 +271,44 @@ def test_failures_to_message_group(test_messenger):
...
+
+
+@pytest.mark.parametrize(
+    "input_cell_indexes, expected_cell_indexes",
+    (
+        (("A1",), ("A1",)),
+        (("B1", "A1"), ("A1", "B1")),  # Sorts by column correctly
+        (("A2", "A1"), ("A1", "A2")),  # Sorts by row correctly
+        (("A2", "B1", "A1"), ("A1", "A2", "B1")),  # Sorts both correctly
+        (("A1", "B50", "B4", "A10"), ("A1", "A10", "B4", "B50")),  # Sorts row numbers numerically not lexicographically
```
wjrm500: This test case doesn't prove numeric over lexicographic as expected result would be the same either way

core/messaging/__init__.py
```diff
@@ -32,8 +38,35 @@ def __getitem__(self, item):
...
+        return self.__cell_indexes
+
+    @classmethod
+    def __clean_cell_indexes(cls, cell_indexes) -> tuple[str, ...] | None:
+        """Remove any duplicate cell references, sorts them in column-then-row order, and ensures uppercase columns"""
+        if cell_indexes is None:
+            return None
+        if None in cell_indexes:
+            raise NotImplementedError("`cell_indexes` cannot contain None elements; must be None directly")
```
wjrm500: Maybe `ValueError` would be more appropriate?

core/messaging/__init__.py
```diff
@@ -32,8 +38,35 @@ def __getitem__(self, item):
...
+        try:
+            cell_indexes = [cls.cell_index_pattern.match(cell_index).groups() for cell_index in set(cell_indexes)]
```
wjrm500: Maybe rename to `cell_index_groups` or something so that we don't have the variable `cell_indexes` changing type / format throughout the function e.g., currently `cell_indexes` might enter the function as `('A1',)` and then be replaced by `[('A', '1')]` on this line, which is slightly confusing

wjrm500 (commented):
Hey @samuelhwilliams, I reviewed on a commit by commit basis and it looks as though you must have read my mind as a lot of my comments were addressed in more recent commits this morning - were you able to see the review as I was writing it? I always figured that it only became visible after you clicked "Submit review". Anyway, I'll double check everything again now but I think you've covered most of the stuff I mentioned already!

core/messaging/__init__.py
samuelhwilliams: Good spot 👍  Flagged this with Gids last week - we defo need to add something that will actually enforce the type hints otherwise it's fairly pointless. I'll make a ticket for that and maybe just look at doing it ...

core/messaging/__init__.py
samuelhwilliams: Can't tell if this was before or after I allowed lowercase - do you think we should, or shouldn't?

tests/messaging_tests/test_messaging.py
samuelhwilliams: yep good spot, fixed 😓

core/messaging/__init__.py
samuelhwilliams: ha yes, also in agreement, fixed 👍

core/messaging/__init__.py
samuelhwilliams: I've changed to `split_cell_indexes`, which I think is hopefully close enough 👍  so fixed

core/messaging/__init__.py
samuelhwilliams: I'm mildly of two minds on removing the property. Having it as a 'protected' property means that it can't (easily) be set to an invalid value (see below) - forcing to either instantiate a new Message, which we know calls the `clean` method, or via combining messages, which also calls `clean`. I think it's a defensive design - but if you'd rather remove it and leave it up to the developer to do the right thing, I could probably be persuaded.

```python
>>> m.cell_indexes = 'foo'
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AttributeError: property 'cell_indexes' of 'Message' object has no setter
```

core/messaging/__init__.py
wjrm500: I was going to suggest we shouldn't, assuming this doesn't break anything existing

core/messaging/__init__.py
wjrm500: Makes sense - gives us a firm guarantee that cell indexes will be correctly formatted and sorted - happy to leave as is and resolve then

wjrm500 (approved):
Approving as all comments addressed and whether or not cell index regex allows lowercase letters is not a blocker

core/messaging/__init__.py
samuelhwilliams: don't think it breaks anything - I'll remove it

wjrm500 (approved):
Latest change to handle "and" or "or" looks good - only thing to note is that it looks like the second cell (after the "and" or "or") isn't taken into account during sorting, meaning that e.g., "A1 or C1" could appear before "A1 or B1". This seems like a real edge case though, so not too bothered

PR #545, opened 2024-04-17
BAU: return error for invalid excel submission

Joel-Pearce:
Adds a new ```SheetCheck``` to return useful error message if admin sheet not present for specified fund.

### DIFF
--- a/core/validation/initial_validation/schemas.py
+++ b/core/validation/initial_validation/schemas.py
@@ -13,9 +13,19 @@ from core.validation.initial_validation.checks import (
     AuthorisationCheck,
     BasicCheck,
     ConflictingCheck,
+    SheetCheck,
 )
 
 TF_ROUND_4_INIT_VAL_SCHEMA = [
+    SheetCheck(
+        sheet="1 - Start Here",
+        error_message=(
+            "The data return template you are submitting is not valid. Please make sure you are submitting a valid "
+            "template for Towns Fund. "
+            'If you have selected the wrong fund type, you can change the fund by returning to the "Submit monitoring '
+            'and evaluation data dashboard" and changing the fund type to Towns Fund before continuing.'
+        ),
+    ),
     BasicCheck(
         sheet="1 - Start Here",
         row=7,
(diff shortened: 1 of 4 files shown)


core/validation/initial_validation/schemas.py
```diff
@@ -13,9 +13,19 @@
...
     ConflictingCheck,
+    SheetCheck,
 )
 
 TF_ROUND_4_INIT_VAL_SCHEMA = [
+    SheetCheck(
+        sheet="1 - Start Here",
+        error_message=(
+            "The data return template you are submitting is not valid. Please make sure you are submitting a valid "
+            "template for Towns Fund."
```
samuelhwilliams, line 24:
```suggestion
            "template for Towns Fund. "
```

nit: could use a space after the period in all of these (sorry 🙈)

core/validation/initial_validation/schemas.py
```diff
@@ -13,9 +13,19 @@
...
+            "If you have selected the wrong fund type, you can change the fund by returning to the 'Submit monitoring "
+            "and evaluation data dashboard' and changing the fund type to Towns Fund before continuing."
```
samuelhwilliams, line 26: Also not sure how picky our content/design people are, but possibly we want fancy quotes here? `“”` instead of `''`? If we aren't generally consistent though then disregard this.

core/validation/initial_validation/schemas.py
Joel-Pearce: Would it be better to be consistent as we're currently not doing this for every error message?

Joel-Pearce: for any error messages * I should say

core/validation/initial_validation/schemas.py
Joel-Pearce: I copy and pasted from a Slack message but looks like we are using fancy quotes, amended.

Joel-Pearce:
> This does look to work nicely - the error message may be too long though. Might be one to get some quick content design on - when on the page it looks a bit overwhelming
> 
> image
> I will leave that to you though and not block on this - as even with this error it's much better than just a 500. Could be one we tweak as a followup (or not 🤷)

I got this from design, but will doublecheck with them

samuelhwilliams:
Nice 👍  If content/design are happy with that error message then I'm happy enough with the code side of it.

I guess this might break down if we add more funds - at the moment I guess we're just saying that Pathfinders has a sheet called x, and Towns Funds called y, and they are exclusively owned by the fund. That's fine for now as long as we really do move away from spreadsheets for future funds 🤞

Joel-Pearce:
> Nice 👍 If content/design are happy with that error message then I'm happy enough with the code side of it.
> 
> I guess this might break down if we add more funds - at the moment I guess we're just saying that Pathfinders has a sheet called x, and Towns Funds called y, and they are exclusively owned by the fund. That's fine for now as long as we really do move away from spreadsheets for future funds 🤞

This check is really only for TF vs. other funds as all other funds should now use the same sheet, and a separate fund check should catch this one. Only an issue with TF because the different tab names were causing a ```KeyError```.

PR #549, opened 2024-04-24
Stop logging with eagerly-formatted strings / move to `ruff`

samuelhwilliams:
https://dluhcdigital.atlassian.net/browse/SMD-812

### Change description
Stop eagerly formatting log messages, which ends up causing some log messages to be double-formatted. When the initial format is injecting a dictionary into the string, the second format (by fsd_utils.logging) can error because the dict looks like a format injection point, but is incorrectly structured.

I've also switched black+flake8+isort out for [ruff](https://docs.astral.sh/ruff/), which is a fast modern tool that combines a bunch of formatting and linting rules. One of these will raise linting issues for log call that have eagerly-formatted log lines.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -14,5 +14,6 @@ repos:
   hooks:
     # Run the linter.
     - id: ruff
+      args: [ --fix ]
     # Run the formatter.
     - id: ruff-format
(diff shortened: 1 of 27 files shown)


gidsg (approved):
👍

PR #551, opened 2024-04-25
Postcodes extract regex SMD-735

sc15zs:
Changes the regex we use to find postcodes in strings. 
Checks that there are no extra characters before or after a valid postcode. 
Allows for new line characters before the postcode.

https://dluhcdigital.atlassian.net/browse/SMD-735

Testing:
pytest tests/test_postcodes.py

### DIFF
--- a/tests/test_postcodes.py
+++ b/tests/test_postcodes.py
@@ -58,7 +58,10 @@ def test_extract_postcodes_list_of_matches():
     assert postcodes == ["BN9 0DF", "BN9 9BP", "BN9 9BN", "BN9 9QD", "BN9 0AS", "BN9 9PA", "BN9 9PD", "BN9 0DF"]
 
 
-def test_extract_postcodes_list_of_matches_2():
+def test_extract_postcodes_list_of_matches_prepended_and_appended_chars():
+    """Tests that those postcodes are not returned where there are additional characters before or after a
+    valid postcode (except for new line characters before that).  e.g. shouldn't return BN9 0DFA, because
+    the 'A' makes the valid postcode invalid"""
     postcode_string = (
         "1. Pedestrian Gateway \nBN9 0DFA (Nr Station); \n2. Wayfinding\n\nBN9 9BP (Riverside); \nBBN9 9BN "
         "(Denton Island); \nBN9 9QD\n (Huggetts Green); \nBN9 0AS (Near railway station); \nBN9 9PA "


tests/test_postcodes.py
```diff
@@ -58,6 +58,19 @@ def test_extract_postcodes_list_of_matches():
     assert postcodes == ["BN9 0DF", "BN9 9BP", "BN9 9BN", "BN9 9QD", "BN9 0AS", "BN9 9PA", "BN9 9PD", "BN9 0DF"]
 
 
+def test_extract_postcodes_list_of_matches_2():
```
gidsg, line 61: Can we give this test a more descriptive name? i.e. describe how it is different from the test above?

sc15zs: Changed the name of the test and added a description below.

samuelhwilliams (commented):
This postcode regex is used in two places:

* the `extract_postcodes` function you've updated here
* `tables.checks.postcode_list`, which validates postcode entry fields to flag errors to the user.

Do you think that we should make the internal behaviour of these checks the same?

`postcode_list` splits the input on `,`, strips whitespace around each element, and then rejects the entire input if any of the elements don't match the postcode regex.
`extract_postcodes` checks if the input is non-empty/non-nan, and then applies the regex to the entire string (ie not split).

I think it'd be nice to process these both in the same way. It probably does make sense to split them on commas, strip the elements, and apply a tightly-bounded regex with start and end anchors `^` and `$`. I'm not sure we need to allow arbitrary text/newlines as if they're full addresses.

What do you think?

tests/test_postcodes.py
sc15zs: @samuelhwilliams 
Sam, for some reason I can't reply to your message, that's why i'm writing here.
As I see it, both the extract_postcodes and postcode_list uses the same regex. Yes, it would be better to use ^ and $ anchors to check for the beginning and the end of a postcode, but it will not work with the extract_postcodes function.
In cases when we have short strings which should contain only one single postcode, like in the case of postcode_list, and we check these one-by-one, I think it would be the best and the easiest check to use. 

But in case of the extract_postcodes function it would not work. This function can have long strings as argument and it has to find and return all postcodes from this string. 
In tests/test_postcodes.py/test_extract_postcodes_list_of_matches there's a long string with several postcodes, each starting in a new line. In this case we can't use ^ and & to check for boundaries, that's why I used /b checking for word boundaries and added the part to the regex to allow for the new line characters. 

If we want to use the same regex for both functions, I don't know how we could use tightly-bound checking in it.
Maybe having two separate regular expressions, one with ^ and &, and the other with word boundary checking /b.
Then in postcode_list to use the tightly-bound expression directly, and in extract_postcodes first extract postcodes with the word boundary checking first, and then check again with the tightly-bound regex, but in this case it's a double work to do.

samuelhwilliams (approved):

PR #552, opened 2024-04-25
[BAU] Fix fixture scope misuse

nsnape:
### Change description
The DB test changed as per this PR was using the wrong fixture. This test was making uncommitted DB changes, but never rolled them back so they leaked into the integration tests, causing seemingly random errors that only happened when running tests in certain orders.

### How to test
run full test suite locally


MarcUsher (approved):
Good catch! Approved

PR #554, opened 2024-04-25
[FMD-265]: normalise fund reference data

Joel-Pearce:
Adds a table, ```fund_dim```, which contains reference data relating to fund-specific metadata.

Changes ```fund_type_id``` in ```programme_dim``` to be an FK pointing to said table.

### DIFF
--- a/core/cli.py
+++ b/core/cli.py
@@ -63,6 +63,7 @@ def create_cli(app):
             db.drop_all()
             db.create_all()
             seed_geospatial_dim_table()
+            seed_fund()
 
         print("Database reset and geospatial data re-seeded.")
 
--- a/core/cli.py
+++ b/core/cli.py
@@ -4,7 +4,7 @@ import requests
 from flask import current_app
 
 from core.db import db
-from core.reference_data import seed_geospatial_dim_table
+from core.reference_data import seed_fund_table, seed_geospatial_dim_table
 from core.util import load_example_data
 
 resources = Path(__file__).parent / ".." / "tests" / "resources"
@@ -77,3 +77,14 @@ def create_cli(app):
         with current_app.app_context():
             seed_geospatial_dim_table()
             print("Geospatial data re-seeded.")
+
+    @app.cli.command("seed-fund")
+    def seed_fund():
+        """CLI command to seed (or re-seed) the fund reference table in isolation.
+        Example usage:
+            flask seed-fund
+        """
+
+        with current_app.app_context():
+            seed_fund_table()
+            print("Fund data re-seeded.")
--- a/core/controllers/get_filters.py
+++ b/core/controllers/get_filters.py
@@ -32,16 +32,16 @@ def get_organisation_names():
 
 def get_funds():
     """
-    Fetches all funds sorted alphabetically by fund_type_id.
+    Fetches all funds sorted alphabetically by fund_code.
 
     :return: A tuple - list of funds and status code 200. If no funds found, aborts with 404 error.
     """
-    funds = Fund.query.order_by(Fund.fund_type_id).with_entities(Fund.fund_type_id).all()
+    funds = Fund.query.order_by(Fund.fund_code).with_entities(Fund.fund_code).all()
 
     if not funds:
         return abort(404, "No funds found.")
 
-    fund_list = [{"name": FUND_ID_TO_NAME[row.fund_type_id], "id": row.fund_type_id} for row in funds]
+    fund_list = [{"name": FUND_ID_TO_NAME[row.fund_code], "id": row.fund_code} for row in funds]
 
     return fund_list, 200
 
--- a/core/controllers/get_filters.py
+++ b/core/controllers/get_filters.py
@@ -5,7 +5,7 @@ from core.const import FUND_ID_TO_NAME
 from core.db import db
 
 # isort: off
-from core.db.entities import Organisation, OutcomeDim, Programme, Project, Submission
+from core.db.entities import Organisation, OutcomeDim, Programme, Project, Submission, Fund
 
 
 # isort: on
@@ -32,20 +32,18 @@ def get_organisation_names():
 
 def get_funds():
     """
-    Fetches all unique funds sorted alphabetically by fund_type_id.
+    Fetches all funds sorted alphabetically by fund_type_id.
 
-    :return: A tuple - list of unique funds and status code 200. If no funds found, aborts with 404 error.
+    :return: A tuple - list of funds and status code 200. If no funds found, aborts with 404 error.
     """
-    programmes = Programme.query.order_by(Programme.fund_type_id).with_entities(Programme.fund_type_id).distinct().all()
+    funds = Fund.query.order_by(Fund.fund_type_id).with_entities(Fund.fund_type_id).all()
 
-    if not programmes:
+    if not funds:
         return abort(404, "No funds found.")
 
-    funds = [
-        {"name": FUND_ID_TO_NAME[programme.fund_type_id], "id": programme.fund_type_id} for programme in programmes
-    ]
+    fund_list = [{"name": FUND_ID_TO_NAME[row.fund_type_id], "id": row.fund_type_id} for row in funds]
 
-    return funds, 200
+    return fund_list, 200
 
 
 def get_outcome_categories():
--- a/core/db/entities.py
+++ b/core/db/entities.py
@@ -24,7 +24,7 @@ class Fund(BaseModel):
 
     __tablename__ = "fund_dim"
 
-    fund_type_id = sqla.Column(sqla.String(), nullable=False, unique=True)
+    fund_code = sqla.Column(sqla.String(), nullable=False, unique=True)
 
     programmes: Mapped[List["Programme"]] = sqla.orm.relationship(back_populates="fund")
 
--- a/core/db/entities.py
+++ b/core/db/entities.py
@@ -19,6 +19,16 @@ class BaseModel(db.Model):
     id: Mapped[GUID] = sqla.orm.mapped_column(GUID(), default=uuid.uuid4, primary_key=True)
 
 
+class Fund(BaseModel):
+    """Stores Fund Entities."""
+
+    __tablename__ = "fund_dim"
+
+    fund_type_id = sqla.Column(sqla.String(), nullable=False, unique=True)
+
+    programmes: Mapped[List["Programme"]] = sqla.orm.relationship(back_populates="fund")
+
+
 class Funding(BaseModel):
     """Stores Funding Entities."""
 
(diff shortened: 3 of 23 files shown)


MarcUsher (commented):
A few comments on things I spotted that were different/better than how I'd approached in the geospatial ticket!

tests/db_tests/test_entities.py
```diff
@@ -127,3 +134,14 @@ def test_geospatial_dim_table(seeded_test_client_rollback):
         region_code = row.itl1_region_code
         region_name = row.data_blob["itl1_region_name"]
         assert itl1_region_pairs[region_code] == region_name
+
+
+def test_fund_dim_unique_constraint(test_client_rollback):
```
MarcUsher, line 139: Should this be in `test_constraints.py`?

tests/integration_tests/test_ingest_db_transactions.py
```diff
@@ -601,7 +610,7 @@ def test_next_submission_numpy_type(test_client_rollback):
     prog = Programme(
         programme_id="HS-ROW",
         programme_name="TEST-PROGRAMME-NAME1",
-        fund_type_id="HS",
+        fund_type_id=Fund.query.first().id,
```
samuelhwilliams, line 613: It'd probably be nice for readability to pin this to specifically be the HS fund PK? Probably as a static value if we can define one for it.

tests/conftest.py
```diff
@@ -141,6 +145,7 @@ def seeded_test_client(test_client: FlaskClient) -> FlaskClient:
     :param test_client: a Flask test client
     :yield: a flask test client with application context and seeded db.
     """
+    seed_fund_table()
```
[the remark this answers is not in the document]
Joel-Pearce: I'd end up having to put a try/catch around the specific bit of code in ```load_example_data``` that loaded data into ```fund_dim``` - or writing something verbose that did a merge/uspert type thing like in your PR. It felt a lot more straightforward to keep the seeding of reference data, or rather data that must be pre-seeded for anything to work, and data tied to specific submissions that is not required for the application to work.

core/reference_data.py
```diff
@@ -59,3 +61,22 @@ def seed_geospatial_dim_table():
...
+def seed_fund_table():
+    """
+    Input seed data to fund_dim table using the mappings in /tests/resources/fund_dim.csv.
+    """
+
+    resources = Path(__file__).parent / ".." / "tests" / "resources"
+    fund_df = pd.read_csv(resources / "fund_dim.csv")
+
+    try:
+        fund_df.to_sql("fund_dim", con=db.session.connection(), index=False, index_label="id", if_exists="append")
```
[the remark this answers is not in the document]
Joel-Pearce: What I've written is a bit of an "anti-pattern" in that there could be an ```IntegrityError``` that we would want to be raised. I'll look at a rewrite. I think the idea is we'll just do a direct SQL insertion every time we onboard a new fund - at least we haven't thought about it further than that - so I was envisioning this function being only used for testing.

core/controllers/get_filters.py
```diff
@@ -32,20 +32,18 @@ def get_organisation_names():
...
+    funds = Fund.query.order_by(Fund.fund_type_id).with_entities(Fund.fund_type_id).all()
 
-    if not programmes:
+    if not funds:
         return abort(404, "No funds found.")
 
-    funds = [
-        {"name": FUND_ID_TO_NAME[programme.fund_type_id], "id": programme.fund_type_id} for programme in programmes
-    ]
+    fund_list = [{"name": FUND_ID_TO_NAME[row.fund_type_id], "id": row.fund_type_id} for row in funds]
```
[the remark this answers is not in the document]
Joel-Pearce: No specific reason, I've only dealt with the bare bones of fund reference data for this PR because I imagined we'd flesh it out going forward, I can add a ticket to the backlog to look at extra data to add to the column

core/reference_data.py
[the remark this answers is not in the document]
MarcUsher: Yeah mine was written with a bit of ambiguity, not knowing if we'd ever have to reseed, but agree it's likely if a new fund is onboarded or new/different geospatial data is added we'd do it as a SQL insertion in a migration rather than using any internal function or script, so if these functions are only used for testing they can be a lot simpler. Interested to see where you go with a rewrite if you do one - if I update my function to be closer to yours I could remove how it's called from `load_example_data`.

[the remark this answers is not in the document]
Joel-Pearce: I updated mine to be closer to yours...we could write something like:
```
except IntegrityError as e:
    db.session.rollback()
    error_message = str(e.orig)

    if isinstance(e.orig, postgresql.exc.UniqueViolation):
        current_app.logger.info(
            "UniqueKey violation occurred when inserting data into the fund_dim table."
            "Please remove any duplicate rows before calling this helper function."
        )
    else:
        current_app.logger.error(f"IntegrityError occurred: {error_message}")
    db.session.close()
```
But it is a bit of a hacky anti-pattern. If what you've written works we can stick with that, and then it's easier to update the reference tables.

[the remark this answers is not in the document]
MarcUsher: Sounds good - I think the only difference between our functions now is that I had `geospatial_dim_df["id"] = None` after creating the `geospatial_dim_df` so that a 'real' DB wouldn't use the hardcoded UUIDs in the csv, but as we discussed if these will only be used for tests and any DB updates will be done via migrations than that's not an issue and I can remove that line from mine.

core/cli.py
```diff
@@ -77,3 +77,14 @@ def seed_geospatial():
         with current_app.app_context():
             seed_geospatial_dim_table()
             print("Geospatial data re-seeded.")
+
+    @app.cli.command("seed-fund")
```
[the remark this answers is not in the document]
MarcUsher: Is it worth adding `seed_fund_table()` to the `reset` method above as I have with `seed_geospatial_dim_table()`? Not critical though, Gideon and I were just talking about these seed methods and how they're probably not being used at all, and if people need to drop the DB entirely they'd probably use the script in the docker runner repo or just rebuild the container?

db/migrations/versions/028_create_fund_ref_table.py
```diff
@@ -0,0 +1,85 @@
...
+        sa.UniqueConstraint("fund_type_id", name=op.f("uq_fund_dim_fund_type_id")),
+    )
+
+    op.execute(
+        """
+        INSERT INTO fund_dim (id, fund_type_id)
+        VALUES
+        ('4a6e9f7d-fc9d-4c12-b1b6-89e784c310e1', 'HS'),
+        ('9fde58b2-8a89-4b2c-af7d-1f968b03c7b9', 'TD'),
+        ('e8c7c1c8-90d3-4b2d-aa50-4a2d4091d4f3', 'PF');
```
MarcUsher, line 36: I remember for the geospatial seeding, Noel had concerns about hard-coding the UUID that will be in the actual DBs into the migration?

Joel-Pearce: @samuelhwilliams requested above that the UUIDs be static so that they be consistent across environments. What are the concerns with hard-coding the values?

samuelhwilliams: If we've decided before to not, I'm fine to not, but be interested to hear Noel's reasons there too. It simplifies a few things in my head (eg testing), but it's not really the end of the world either way.

I've seen and used static IDs in migrations for 'reference'y data before and it's been fine.

nsnape: TBH it's probably fine, the main reasons for not doing probably don't really apply to our use case (ie not exposing DB info on public repo) so I'm happy either way, as long those keys were generated correctly and are actually unique

core/db/entities.py
```diff
@@ -19,6 +19,16 @@ class BaseModel(db.Model):
     id: Mapped[GUID] = sqla.orm.mapped_column(GUID(), default=uuid.uuid4, primary_key=True)
 
 
+class Fund(BaseModel):
+    """Stores Fund Entities."""
+
+    __tablename__ = "fund_dim"
+
+    fund_type_id = sqla.Column(sqla.String(), nullable=False, unique=True)
```
samuelhwilliams, line 27: What do you think of just calling this `code` rather than `fund_type_id`?

I guess my thoughts around this are:
* In our DB, `id` often means a UUID.
* Adding `type` possibly suggests that there are groups of funds that are somehow related, and this identifies that, but it doesn't.

core/db/entities.py
```diff
@@ -296,11 +306,12 @@ class Programme(BaseModel):
     programme_id = sqla.Column(sqla.String(), nullable=False, unique=True)
 
     programme_name = sqla.Column(sqla.String(), nullable=False)
-    fund_type_id = sqla.Column(sqla.String(), nullable=False)
     organisation_id = sqla.Column(GUID(), sqla.ForeignKey("organisation_dim.id"), nullable=False)
+    fund_type_id = sqla.Column(GUID(), sqla.ForeignKey("fund_dim.id"), nullable=False)
```
samuelhwilliams, line 310: This gets particularly confusing because now the FK related fields are called `fund_type_id`, and connect to the `fund_dim` table which has a `fund_type_id` field - but that isn't the field we're talking about.

I think I'd either call these `fund_id` or `fund_dim_id`? Possibly updating the `Fund` python model to `FundDim` so that we're using consistent terminology.

core/reference_data.py
```diff
@@ -59,3 +59,41 @@ def seed_geospatial_dim_table():
                 db.session.add(geospatial_record)
 
     db.session.commit()
+
+
+def seed_fund_table():
+    """
+    Input seed data to fund_dim table using the mappings in /tests/resources/fund_dim.csv.
```
samuelhwilliams, line 66: Maybe something in here saying that this is for local dev only. Obviously not just this function, but it looks slightly weird that this is in the `core` of the app but then reaches into some test data fixture to insert into the DB. I think just being super clear this isn't intended to touch any of the deployed envs would be nice.

samuelhwilliams (commented):
A few comments just around naming. Make your own call for those, but otherwise happy 👍 Will give a final ✅  after you've replied.

core/db/entities.py
Joel-Pearce: Makes sense, I've changed to ```fund_code```. We'll probably update this table in the future to have ```full_name_of_fund``` or some other column too.

core/db/entities.py
Joel-Pearce: I've changed the name of the column in ```fund_dim```, I'd propose leaving this as ```fund_type_id``` for now in the Programme table as during ingestion/transformation it is ```FundType_ID```, and is then serialised out as ```FundTypeID``` so may just add extra confusion if the 'type' bit disappears between these two processes. We can of course talk to the business about serialising out the name as something different in the future.

core/db/entities.py
samuelhwilliams: yeah that makes sense to have for sure. I almost recommended it but it seems an easy thing to carve out as something separate 👍

core/db/entities.py
samuelhwilliams: Ah interesting - good context - thanks.

PR #556, opened 2024-05-01
BAU - Remove load_data from ingest.py module

wjrm500:
### Ticket

No ticket - BAU change.

### Change summary

- Remove `load_data` method from `ingest.py` module and call `populate_db` directly from `ingest`
- Believe that `load_data` used to contain additional logic inc. possibility of calling `populate_db_with_historical_data`, but this was removed


samuelhwilliams (approved):
Looks sensible 👀

PR #557, opened 2024-05-01
BAU: improve mapping docstring

Joel-Pearce:
Improve ```map_data_to_models``` docstring.

### DIFF
--- a/core/controllers/mappings.py
+++ b/core/controllers/mappings.py
@@ -47,7 +47,25 @@ class DataMapping:
     def map_data_to_models(self, data: pd.DataFrame) -> list[db.Model]:
         """Maps the given data to a list of database models.
 
-        :param data: The data to map.
+        - Renames columns such that they match those in the data model.
+        - Replaces the human-readable values in selected FKs to match the UUIDs for the rows in the parent
+        tables corresponding to those values.
+        - Moves all specified columns into a JSONB blob where needed.
+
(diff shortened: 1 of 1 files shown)


core/controllers/mappings.py
```diff
@@ -47,7 +47,20 @@ class DataMapping:
...
+        tables, and then using these values to get the UUID for the row corresponding to both in the ProgrammeJunction
+        table.
+
+        FK look-ups for project_id require the context of the submission id for the current ingestion,
+        which we can derive from the ProgrammeJunction table once the appropriate FK look-ups have been done.
+
+        :param data: The data to map, arranged into DataFrames corresponding to tables in the db.
         :return: A list of database models.
         """
         global CURRENT_SUBMISSION_ID
```
samuelhwilliams, line 66: If it's not too much to ask, I would _love_ a paragraph that goes through the use of this - because it's fairly unintuitive and scary. There's no reference to this in the global scope anywhere, so this looks like it can be undefined, and I think the code relies heavily on some implicit ordering to trust that it will be written before it's ever read. It's fairly dodgy looking.

It also relies fairly heavily on us being quite wasteful with our server resource, in that each worker can only ever be processing one request at a time - so we cannot use threads/greenlets to process requests while we have this bit of code.

Joel-Pearce: Did my best :)

samuelhwilliams (approved):
Nice one 👍

PR #558, opened 2024-05-01
Allow turning on SQL logging

samuelhwilliams:
SQLAlchemy has a simple config option that enables logging of all queries send to the DB. By adding this to dev config we can make it easy (easier) to turn on in local.


Joel-Pearce (approved):
LGTM

PR #559, opened 2024-05-01
SMD-819: Make current underspend optional

gidsg:
### Change description
Make current underspend optional as it is not required in Q4 according to the spreadsheet

(Added a TODO to make this a cross-table validation against the reporting period, but not implemented this yet so we can expedite this fix)

- [ ] Unit tests and other appropriate tests added or updated
- [ ] README and other documentation has been updated / added (if needed)
- [x] Commit messages are meaningful and follow good commit message guidelines (e.g. "FS-XXXX: Add margin to nav items preventing overlapping of logo")

### How to test
Leaving current underspend blank should allow successful submission

PR #561, opened 2024-05-02
SMD-819 - Pathfinders - current underspend cross-table validation

wjrm500:
### Ticket

[Validation on ingest for PF current underspend question](https://dluhcdigital.atlassian.net/browse/SMD-819)

### Change summary
- [A separate PR](https://github.com/communitiesuk/funding-service-design-post-award-data-store/pull/559) makes current underspend optional (nullable), allowing users to leave this field blank, which is fine as we are currently in Q4 (quarter 4)
- Outside of Q4 we want to validate that current underspend has been entered
- This requires cross-table validation as we need to check reporting period alongside current underspend
- This PR adds that cross-table validation check

### DIFF
--- a/tests/validation_tests/specific_validation_tests/pathfinders/test_round_1.py
+++ b/tests/validation_tests/specific_validation_tests/pathfinders/test_round_1.py
@@ -8,6 +8,7 @@ from core.validation.specific_validation.pathfinders.round_1 import (
     _check_actual_forecast_reporting_period,
     _check_bespoke_outputs,
     _check_credible_plan_fields,
+    _check_current_underspend,
     _check_intervention_themes_in_pfcs,
     _check_projects,
     _check_standard_outcomes,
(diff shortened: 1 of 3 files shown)


wjrm500:
Hey @gidsg I've assigned you as a reviewer on this as you did the original fix PR, let me know if you want to chat cross-table validation on a Slack huddle :)

tests/validation_tests/specific_validation_tests/pathfinders/test_round_1.py
```diff
@@ -218,6 +219,52 @@ def test__check_credible_plan_fields_fails(mock_df_dict):
...
+            section="Current underspend",
+            cell_indexes=("B1",),
+            description="Current underspend must be filled in if the reporting period is not Q4.",
+            error_type=None,
+        )
+    ]
+
+    # Reset the mock data
+    mock_df_dict["Reporting period"].iloc[0, 0] = original_reporting_period
+    mock_df_dict["Current underspend"].iloc[0, 0] = original_underspend
```
gidsg, line 265: Could we use a copy of this dict with the changes rather than mutating/resetting it here?

samuelhwilliams: `mock_df_dict` should probably just be function-scoped so that we don't even have to think about changes to it being carried between tests. That seems too painful to want to think about. Any reason we can't do this?

wjrm500: Have just tried switching the fixture from module scope to function scope and all tests pass, so seems like a no-brainer. As this pattern of capturing original values from the mock_df_dict, mutating, running the function and then resetting to those values is prevalent throughout this entire module, not just these new functions, what do you think to using Gideon's copy technique on this PR in particular to keep the scope limited to the ticket, but then for me to create a separate ticket and PR to change the scope and remove all the capturing original values / resetting to those values logic?

samuelhwilliams: Whichever you prefer is fine with me :)

wjrm500:
@gidsg @samuelhwilliams - I've used `copy.deepcopy` to create a new version of `mock_df_dict` in each of the two new test functions, avoiding mutating the fixture object. Shallow copying doesn't work as we need to re-create the `DataFrame` objects inside the dict, not just the dict. Although it's not ideal having to add a new import here, it keeps the changeset small. Originally I had hoped to simply change the scope of the `mock_df_dict` fixture from `module` to `function` and  remove the mutation from the new test functions in this PR, and then remove the mutation from the other test functions in a separate PR; however, re-scoping the fixture alone does not have the desired effect, as it simply recycles the reference to the `EXTRACTED_DATA` object, which is created elsewhere. Thus, for the fixture re-scoping change, we will need to create the mock data inside the function - basically everything inside `extracted_data.py` will need to be defined inside a function that can be called from the `mock_df_dict` fixture. This is fine but it dramatically increases the number of lines in the changeset, so I figured it would be best to leave that to the secondary PR mentioned above.

samuelhwilliams (approved):

PR #562, opened 2024-05-02
SMD-822 - Factor out control mappings

wjrm500:
### Ticket

[Factor out control mappings](https://dluhcdigital.atlassian.net/browse/SMD-822)

### Change summary
- **Replacing programme_id_to_project_ids with programme_to_projects**
This allows simplification within cross-table validation as we don't need to run lookups.

- **Using programme names not IDs for bespoke outputs and outcomes lookups**
This allows simplification within cross-table validation as we don't need to run lookups.

- **Removing control mappings dependency from cross-table validation**
The original purpose of having a control mappings module was to share functionality across cross-table validation and transformation, and share functionality across functions within those modules too - however, in practice many of the control mappings were used in one single function in either cross-table validation and transformation, which meant the abstraction became more confusing than helpful. In this change, we remove the external dependency on control mappings from cross-table validation, generating mappings within the functions they are used.

- **Updating docstrings in cross-table validation**
Docstrings have been updated so that all include parameter and return type definitions and descriptions. Also transferred duplicate message about adding 1 to cell index to module docstring.

- **Removing control mappings including transformation dependency**
The only place remaining in the code where control mappings are used is transformation, as the dependency in cross-table validation was previously removed. By generating the required mappings in the transformation module, control mappings become redundant and we can remove the module and its corresponding test module.


samuelhwilliams (approved):
This all looks and runs good for me, so I'm happy. Feels a lot cleaner. You may want to get another pair of experienced eyes over this though if you want some extra certainty - up to you ...

wjrm500:
> This all looks and runs good for me, so I'm happy. Feels a lot cleaner. You may want to get another pair of experienced eyes over this though if you want some extra certainty - up to you ...

Thanks @samuelhwilliams - I've just examined the changeset line-by-line and I'm happy to go ahead without secondary review, will merge now

PR #563, opened 2024-05-02
Persist submitting user metadata

sc15zs:
Ticket:
https://dluhcdigital.atlassian.net/browse/FMD-251

Description:
Adds two new fields to Submission model in data-store.
Saves submitting user's account_id and email to these fields passed in from post-award-submit.
PR for this:
https://github.com/communitiesuk/funding-service-design-post-award-submit/pull/160

Test:
Tested locally and in dev environment.


Joel-Pearce (changes requested):
This functionality is going to require a unit test in which a mocked json containing this metadata from the frontend hits the /ingest endpoint. 

At the very least, ```save_submission_file_name_and_user_metadata``` is going to require a unit test.

core/controllers/ingest.py
```diff
@@ -152,10 +152,14 @@ def parse_body(body: dict) -> tuple[str, int, dict | None, bool]:
...
     auth = parse_auth(body)
     do_load = body.get("do_load", True)  # defaults to True, if False then do not load to database
+    submitting_account_id = body.get("submitting_account_id", None)
+    submitting_user_email = body.get("submitting_user_email", None)
 
     # Set these values for reporting sentry metrics via `core.metrics:capture_ingest_metrics`
     g.fund_name = fund
     g.reporting_round = reporting_round
+    g.submitting_account_id = submitting_account_id
+    g.submitting_user_email = submitting_user_email
```
samuelhwilliams, line 162: I think we can avoid using `g` here by returning this metadata directly, and passing it through to the `populate_db` function.

I think it's nice to keep the use of `g` as small as possible because it's a global namespace and has all of the traditional issues with global namespaces. This one's reasonably inoffensive because we're only using these values in one place, but I also think that we can just pass them through the function call chain quite easily as well (unlike the other things that we're sticking on `g` here).

db/migrations/versions/2024_05_01_1700-4a0c70f02342_added_submitting_account_id_submitting_.py
```diff
@@ -0,0 +1,34 @@
+"""Added submitting_account_id & submitting_user_email fields to Submission model.
```
nsnape, line 1: Can you rename this migration file to follow the naming convention of the others? You can then update the revision ID referenced both in the docstring and as the revision var, to match the filename (see previous ones for examples)

nsnape (changes requested):

sc15zs:
> This functionality is going to require a unit test in which a mocked json containing this metadata from the frontend hits the /ingest endpoint.
> 
> At the very least, `save_submission_file_name_and_user_metadata` is going to require a unit test.

Unit test has been added to make sure that new field values are saved to submission model from submission json.

core/controllers/ingest.py
sc15zs: Removed the new fields from g object and passed them into functions.

db/migrations/versions/2024_05_01_1700-4a0c70f02342_added_submitting_account_id_submitting_.py
sc15zs: Migration filename has been changed to match naming convention.

core/controllers/ingest.py
```diff
@@ -394,6 +398,8 @@ def populate_db(
     mappings: tuple[DataMapping],
     excel_file: FileStorage,
     load_mapping: dict[str, Callable],
+    submitting_account_id=None,
+    submitting_user_email=None,
```
samuelhwilliams, line 402: Could you add type hints to these please? Both are `str | None`. If you can cascade these types down as needed to other functions too, that'd be grand, thanks.

core/db/entities.py
```diff
@@ -572,6 +572,8 @@ class Submission(BaseModel):
     reporting_round = sqla.Column(sqla.Integer(), nullable=False)
     submission_filename = sqla.Column(sqla.String(), nullable=True)
     data_blob = sqla.Column(JSONB, nullable=True)
+    submitting_account_id = sqla.Column(sqla.String())
+    submitting_user_email = sqla.Column(sqla.String())
```
samuelhwilliams, line 576: We may want to think about whether we can backfill these columns and make them non-nullable in the future. Could you make a ticket for this and put it in the refinement bucket of one of the boards please (probably submit's board)? :)

samuelhwilliams (approved):

tests/integration_tests/test_ingest_component_pathfinders.py
```diff
@@ -505,3 +509,66 @@ def test_ingest_pf_r1_general_and_cross_table_validation_errors(
...
+
+    save_submission_file_name_and_user_metadata(
+        excel_file=file,
+        submission_id="S-PF-R01-1",
+        submitting_account_id="0000-1111-2222-3333-4444",
+        submitting_user_email="testuser@levellingup.gov.uk",
+    )
+
+    # Check that submitting_account_id and submitting_user_email are saved on the Submission
+    sub = Submission.query.filter_by(submission_id="S-PF-R01-1").first()
```
nsnape, line 572: consider changing first() to one() - it's better practice as:

- If you are only expecting one row, one() explicitly gets this, or throws an error.
- If something changes in the future and there ends up being more than one row returned by that query (due to test fixture leakage, or badly defined test setup, or some regressive code change) then first() may give arbitrary, difficult to debug error messages, whereas one() will explicitly say it only expects one result

tests/integration_tests/test_ingest_component_pathfinders.py
```diff
@@ -505,3 +509,66 @@ def test_ingest_pf_r1_general_and_cross_table_validation_errors(
...
+            ),
+            "do_load": True,
+            "submitting_account_id": "0000-1111-2222-3333-4444",
+            "submitting_user_email": "testuser@levellingup.gov.uk",
+        },
+    )
+
+    assert response.status_code == 200, f"{response.json}"
+
+    sub = Submission.query.filter_by(submission_id="S-PF-R01-1").first()
```
nsnape, line 543: as per other comment, switch to use one()

nsnape (changes requested):

tests/integration_tests/test_ingest_component_pathfinders.py
```diff
@@ -505,3 +509,66 @@ def test_ingest_pf_r1_general_and_cross_table_validation_errors(
...
+    )
+
+    assert response.status_code == 200, f"{response.json}"
+
+    sub = Submission.query.filter_by(submission_id="S-PF-R01-1").one()
+    assert sub.submitting_account_id == "0000-1111-2222-3333-4444"
+    assert sub.submitting_user_email == "testuser@levellingup.gov.uk"
+
+
+def test_save_submission_file_name_and_user_metadata():
```
nsnape, line 548: this needs to use a test fixture that resets the changes, made within the test, otherwise there's a big risk of the data leaking into another test (probably in a separate module, or if someone extends this one). I think in this case the changes are not committed to the DB (from what I can tell by quickly reading it) so the appropriate fixture is probably seeded_test_client_rollback

nsnape (changes requested):

PR #564, opened 2024-05-03
SMD-823 - Re-scope PF mock data fixture

wjrm500:
### Ticket
[Re-scope Pathfinders mock data fixture](https://dluhcdigital.atlassian.net/browse/SMD-823)

### Change summary
- **Refactor PF mock data creation into function for use in function-scoped fixture**
By moving all mock data creation into a function, we ensure that when this function is called from a Pytest function-scoped fixture, fresh mock data is created rather than the fixture simply recycling a reference to existing mock data.
- **Re-scoping PF mock data fixture from module to function**
Now that mock data is generated within a function that is called by the fixture, re-scoping the fixture from module to function level will work as intended and ensure that tests that use the fixture get a fresh set of mock data, minimising unintended side effects.
- **Removing mutation reversals to PF mock data fixture**
Pathfinders mock data fixture mock_df_dict is now function-scoped and so capturing original values and resetting to these post-test is a redundant operation. This change removes all code relating to that redundancy, with the exception of test_check_actual_forecast_reporting_period, which requires refactoring with parameterisation.
- **Refactoring unit test for check_actual_forecast_reporting_period**
Split unit test into two - passes and fails; added a new pass case; parameterised; and removed redundant capturing of original values and resetting.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -521,7 +521,7 @@ def towns_fund_td_round_3_submission_data(test_client_reset):
     db.session.commit()
 
 
-@pytest.fixture(scope="module")
+@pytest.fixture(scope="function")
 def mock_df_dict() -> dict[str, pd.DataFrame]:
     return get_extracted_data()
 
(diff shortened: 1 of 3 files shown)


samuelhwilliams (approved):
Nice changes here - this is a lot cleaner and more reliable 👍  Nice logical commits+messages too.

PR #566, opened 2024-05-03
[FMD-278]: consolidate funding data

Joel-Pearce:
Consolidates funding data as per: https://dluhcdigital.atlassian.net/browse/FMD-278


Joel-Pearce:
deployment failed

notes:

> 
the issue is that for some reason data_blob is being set as null on some rows when all that should be happening is some values in some keys should be being moved around, the syntax is wrong but it's quite a tricky one because it's JSONB manipulation with raw SQL, maybe just be easier to use ORM IMO

albertkol:
Will the pipeline pass now? :popcorn:

db/migrations/versions/038_correct_fund_data.py
```diff
@@ -0,0 +1,119 @@
+"""empty message
```
samuelhwilliams, line 1: 😢

gidsg (approved):
LGTM, we'll want to make sure these changes are communicated to the Fund Teams/end users 👍

db/migrations/versions/038_correct_fund_data.py
albertkol: I changed the message! I will not put up with this emoji passive aggressive feedback!

PR #571, opened 2024-05-08
Start tracking reporting round on programme junction

samuelhwilliams:
https://dluhcdigital.atlassian.net/jira/software/c/projects/FMD/boards/139?selectedIssue=FMD-260

### Change description
Adds a new column `reporting_round` to the `programme_junction` table,
which will eventually replace the `submission.reporting_round` column.
This patch only starts populating the data and doesn't deal with
backfilling existing nulls. That will be done in a follow-up PR.

### How to test
* Upload some spreadsheets
* Check the DB table `programme_junction` and see that the reporting round has been recorded.

### DIFF
--- a/core/db/entities.py
+++ b/core/db/entities.py
@@ -373,6 +373,7 @@ class ProgrammeJunction(BaseModel):
         sqla.ForeignKey("submission_dim.id", ondelete="CASCADE"), nullable=False
     )
     programme_id: Mapped[GUID] = sqla.orm.mapped_column(sqla.ForeignKey("programme_dim.id"), nullable=False)
+    reporting_round = sqla.Column(sqla.Integer, nullable=True)
 
     # parent relationships
     submission: Mapped["Submission"] = sqla.orm.relationship(back_populates="programme_junction", single_parent=True)
@@ -402,6 +403,11 @@ class ProgrammeJunction(BaseModel):
         sqla.Index(
             "ix_programme_junction_join_programme",
             "programme_id",
+        ),
+        sqla.UniqueConstraint(
+            "programme_id",
+            "reporting_round",
+            name="uq_programme_junction_unique_submission_per_round",
         ),
     )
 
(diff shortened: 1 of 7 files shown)


core/db/entities.py
```diff
@@ -373,6 +373,7 @@ class ProgrammeJunction(BaseModel):
         sqla.ForeignKey("submission_dim.id", ondelete="CASCADE"), nullable=False
     )
     programme_id: Mapped[GUID] = sqla.orm.mapped_column(sqla.ForeignKey("programme_dim.id"), nullable=False)
+    reporting_round = sqla.Column(sqla.Integer, nullable=True)
```
nsnape, line 376: Should this be non-nullable? Or is that a future PR change?

nsnape: Oh, I guess you're not populating it yet, so it has to be nullable

tests/integration_tests/test_ingest_component_pathfinders.py
```diff
@@ -134,6 +134,16 @@ def test_ingest_pf_r1_file_success_with_tf_data_already_in(
     # check submission id correctly generated
     assert len(Submission.query.filter(Submission.submission_id == "S-PF-R01-1").all()) == 1
 
+    # TODO FMD-260: remove after this is enforced via DB constraint and tested elsewhere
+    assert (
+        ProgrammeJunction.query.join(Submission, Submission.id == ProgrammeJunction.submission_id)
+        .join(Programme, Programme.id == ProgrammeJunction.programme_id)
+        .filter(Submission.submission_id == "S-PF-R01-1", Programme.programme_id == "PF-BOL")
+        .first()
```
nsnape, line 142: should we use one() here instead? Non a massive deal, but we've had a few instances now where we use first() when really we only expect one row, but then due to some later misuse of test fixtures elsewhere, we get test leakage and more than one row returning, meaning first() is no longer necessarily getting the row you want, and when it doesn't the error message can be confusing. I think if we expect one row, always best to use one()

nsnape (changes requested):

core/db/entities.py
samuelhwilliams: Yep exactly

tests/integration_tests/test_ingest_component_pathfinders.py
samuelhwilliams: Sure - although this should only be around for one or two PRs

nsnape (approved):

PR #575, opened 2024-05-09
SMD-842 - Re-structuring data-store repo

wjrm500:
### Ticket
https://dluhcdigital.atlassian.net/browse/SMD-842

### Change summary
- **Moving 'tables' under 'core' directory and moving 'tables/tests' under 'tests' directory**
Code under 'tables' can definitely be considered part of the core application at this point. I believe Cyrus originally created a separate directory outside of 'core' to house this stuff just to keep it isolated during development, but it's time to move it where it belongs, to improve the logical layout of the repo as a whole.
- **Moving 'table_configs' under 'tables'**
- **Re-naming 'tables' to 'table_extraction' for clarity** 
Felt that the name 'tables' wasn't very meaningful.
- **Re-arranging 'core/validation' into TF and PF-specific directories**
This follows the same pattern we use in 'core/transformation' and helps keep TF and PF-specific code more obviously isolated, which will be especially important when we port PF schema validation from 'core/table_extraction' to 'core/validation' shortly.
- **Moving table validation under 'core/validation'**
Although table validation for Pathfinders is strongly linked to the table extraction process (they are configured within the same schema, and share the use of the Table class), I feel that it should be moved to the 'core/validation' directory, as that is where you would expect to find all validation logic, not just for TF validation logic. Moving the validation-specific modules 'core/table_extraction/checks.py' and 'core/table_extraction/validate.py' is not the end of the story, as we then need to adjust 'core/table_extraction/__init__.py' to remove the imports from those modules, as it does not make sense for an '__init__.py' file to be importing from a completely different module. Further, 'core/table_extraction/exceptions.py' needs to be broken in two, to allow the validation-specific 'TableValidationError' and 'TableValidationErrors' to move alongside the 'checks' and 'validate' modules into 'core/validation'. Then, the full pipeline test needs to be moved one level up the test directory hierarchy, as it now concerns not only code in 'core/table_extraction' but also code in 'core/validation'; the E2E CSV resources likewise need to move, and given that we are creating a new 'tests/resources/pathfinders' folder, it makes sense also to move 'tests/resources/extracted_data.py' into it while we're at it. Finally, we needed to adjust the basename of the migrating 'test_validate.py' file, to avoid a conflict error when tests are run.
- **Re-naming validate and transform module basenames to differentiate**
It's a pain having multiple files with basenames likes 'round_1.py' when developing, making navigating the repo more difficult. This change gives such modules more specific basenames.
- **Consistent interface across TF and PF transformation**
By using the method name 'transform' and aliasing where ambiguity might arise, we create a consistent interface for the transformation stage of the pipeline, which is analogous in Towns Fund and Pathfinders.


core/controllers/ingest.py
```diff
@@ -39,7 +39,7 @@
 from core.messaging.messaging import failures_to_messages, group_validation_messages
 from core.metrics import capture_ingest_metrics
 from core.table_extraction.config.pf_r1_config import PF_TABLE_CONFIG
-from core.transformation.pathfinders.pf_transform_r1 import pathfinders_transform
+from core.transformation.pathfinders.pf_transform_r1 import transform as pf_r1_transform
```
samuelhwilliams, line 42: This reads maybe slightly weird to me - we have a module called `pf_transform_r1` and then a function aliased as `pf_r1_transform`. The overlap and inconsistency feels weird just as I'm reading this. I don't really have any strong objections or suggestions though.

samuelhwilliams (approved):
I don't have any objections to this - generally all feels fairly reasonable. However I don't have strong opinions on the code structure yet, so you may want to get the opinions of someone else who's been working with it longer for a second opinion :) up to you

core/controllers/ingest.py
wjrm500: My logic was basically that a function name should end with a verb, whereas a module name needn't - however I agree it does read more like I've just got confused haha

gidsg (approved):
Looks sensible to me!

PR #577, opened 2024-05-10
SMD-503 - Tie Pandera check functions to error messages

wjrm500:
### Ticket
[Pathfinders - use direct Check instantiation instead of using pandera.register_check_method](https://dluhcdigital.atlassian.net/browse/SMD-503)

### Change summary
Currently we register custom Pandera check functions (used in schema validation for Pathfinders) through Pandera's `register_check_method` decorator. The problem is that we cannot specify an error message that should be raised if the check fails, and that means that every time we instantiate a check in our code, we have to remember to pass the error message that corresponds to the check as a separate argument, which is error-prone and verbose. For example, we add a float datatype check to the validation schema using `pa.Check.is_float(error=PFErrors.IS_FLOAT)`.

As check functions are very much tied to error messages in the real world, for the sake of a consistent user experience, it makes sense to tie them together formally in the backend, instead of simply by convention. Thus, this change is to remove the practice of registering check functions via a decorator and handing over the process of actually instantiating these checks to Pandera, and instead instantiate checks in our own application code, facilitating the passing of error messages as arguments. As we also need to support arguments, we cannot instantiate them directly in the `checks.py` module and instead need to offer wrapper functions or closures. What we end up with is the ability to add a float datatype check to the validation schema using `checks.float_check()`.

To facilitate this tying together of check functions and error messages for all checks used during Pandera schema validation for Pathfinders, checks that do not use custom logic have also been assigned closures in the `checks.py` module (e.g., the email regex check, which uses Pandera's built-in `str_matches` check function).

As checks are now tied to error messages, tests had to be updated to reflect this and ensure that the real error messages (from `PFErrors`) were being verified against.

Finally, the check function `exactly_five_rows` was replaced by the more generic `exactly_x_rows`, slightly out of scope but definitely the right thing to do.

### DIFF
--- a/core/table_extraction/config/pf_r1_config.py
+++ b/core/table_extraction/config/pf_r1_config.py
@@ -7,13 +7,7 @@ from datetime import datetime
 
 import pandera as pa
 
-
-class PFRegex:
-    BASIC_EMAIL = r"^[a-zA-Z0-9_.+-]+@[a-zA-Z0-9-]+\.[a-zA-Z0-9-.]+$"  # will catch most common errors
-    BASIC_TELEPHONE = (
-        r"^(((\+44\s?\d{4}|\(?0\d{4}\)?)\s?\d{3}\s?\d{3})|((\+44\s?\d{3}|\(?0\d{3}\)?)\s?\d{3}"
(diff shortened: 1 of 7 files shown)


wjrm500:
Superseded by https://github.com/communitiesuk/funding-service-design-post-award-data-store/pull/578

wjrm500:
Superseded by https://github.com/communitiesuk/funding-service-design-post-award-data-store/pull/578

PR #578, opened 2024-05-13
SMD-503 - Tie Pandera check functions to error messages

wjrm500:
### Ticket
[Pathfinders - use direct Check instantiation instead of using pandera.register_check_method](https://dluhcdigital.atlassian.net/browse/SMD-503)

### Change summary
- **Moving const classes from config module to new consts module under core/validation**
The const / enum classes PFErrors, PFRegex and PFEnums belong in a separate module to the round-specific config module they are currently found in. All are specifically tied to the schema validation process, for which we already have modules including checks.py and exceptions.py under core/validation. A new consts.py should be created in the same location as those other modules to house these const / enum classes.
- **Replacing 'exactly five rows' check and error const with more generic 'exactly X rows' equivalent**
- **Replacing PF 'Amount moved' validation errors with more generic alternatives**
We can use greater than or less than error messages with placeholders for formatting, to promote re-use.
- **Adding docstrings and alphabetising PF enum classes**
- **Adding specific PF validation error message for integers not floats**
Currently the integer check uses the same error message const 'IS_FLOAT' as the float check, which seems confusing.
- **PF validation - tying together check functions and error messages for existing custom checks**
Currently we register custom Pandera check functions (used in schema validation for Pathfinders) through Pandera's register_check_method decorator. The problem is that we cannot specify an error message that should be raised if the check fails, and that means that every time we instantiate a check in our code, we have to remember to pass the error message that corresponds to the check as a separate argument, which is error-prone and verbose. As check functions are very much tied to error messages in the real world, for the sake of a consistent user experience, it makes sense to tie them together formally in the backend, instead of simply by convention. Thus, this change is to remove the practice of registering check functions via a decorator and handing over the process of actually instantiating these checks to Pandera, and instead instantiate checks in our own application code, facilitating the passing of error messages as arguments. As we also need to support arguments, we cannot instantiate them directly in the 'checks.py' module and instead need to offer wrapper functions or closures. This commit specifically targets only checks that have existing custom check functions.
- **PF validation - tying together check functions and error messages for Pandera built-in check functions**
These built-in check functions include things like str_matches, greater_than and is_in. To support error message tying, we need to create new custom wrapper functions in the 'checks.py' module. This change completes the main objective of the workon this ticket and as such removes the need to import error messages (PFErrors) in the PF round 1 config module. Removing this import means we then have to reimport PFErrors from a different location in cross-table validation.
- **Replacing 'positive' PF error message const with more generic 'greater than or equal to' const**
- **Removing remaining references to pa.Check**
Searched the repo for all remaining uses of pa.Check and replaced where necessary. Only place to leave it in is 'table_level_check_schema' Pytest fixture which has some custom check only used in tests, not in the application itself.


wjrm500:
@samuelhwilliams - this is basically the same as https://github.com/communitiesuk/funding-service-design-post-award-data-store/pull/577 but with changes broken down into more specific commits

tests/validation_tests/pathfinders/schema_validation_tests/test_pf_schema_validate.py
```diff
@@ -80,9 +80,7 @@ def joint_uniqueness_schema():
 def greater_than_5_schema():
     return dict(
         columns={
-            "Column1": pa.Column(
-                checks=[pa.Check.is_float(error="is_float"), pa.Check.greater_than(5, error="greater_than")]
-            ),
+            "Column1": pa.Column(checks=[checks.is_float(), pa.Check.greater_than(5, error="greater_than")]),
```
samuelhwilliams, line 83: Is this pa.check intentional?

samuelhwilliams: Same with the one in `greater_than_5_schema` fixture

samuelhwilliams (approved):

PR #579, opened 2024-05-13
SMD-653 - Dependency injection for Pathfinders

wjrm500:
### Ticket
[Pathfinders - replace hard-coded dependencies with dependency injection](https://dluhcdigital.atlassian.net/browse/SMD-653)

### Change summary
Currently we have an IngestDependencies class that is used by Pathfinders but only for initial validation schema and table to load function mapping, as the other dependencies don't map directly between Towns Fund and Pathfinders. For the dependencies that Towns Fund has but not Pathfinders, we specify None in the constructor for the IngestDependencies object when doing a Pathfinders run, while for the dependencies that Pathfinders has but not Towns Fund, we hard-code them in the main 'ingest' pipeline flow.

This change is to break apart IngestDependencies into three classes - a base IngestDependencies, containing shared dependencies; and then specific child classes for Towns Fund and Pathfinders, containing fund-specific dependencies. This allows us to share the same dependency factory method but to return IngestDependency objects specific to funds. We can then leverage this change to replace the hard-coding of Pathfinders-specific dependencies with dependency injection, similar to what Towns Fund enjoys now, which will help when we move to Pathfinders round 2.

### DIFF
--- a/core/controllers/ingest.py
+++ b/core/controllers/ingest.py
@@ -20,6 +20,8 @@ from core.aws import upload_file
 from core.const import DATETIME_ISO_8601, EXCEL_MIMETYPE, FAILED_FILE_S3_NAME_FORMAT
 from core.controllers.ingest_dependencies import (
     IngestDependencies,
+    PFIngestDependencies,
+    TFIngestDependencies,
     ingest_dependencies_factory,
 )
 from core.controllers.load_functions import (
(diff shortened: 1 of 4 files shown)


core/controllers/ingest.py
```diff
@@ -98,7 +96,7 @@ def ingest(body: dict, excel_file: FileStorage) -> tuple[dict, int]:
     try:
         initial_validate(workbook_data, ingest_dependencies.initial_validation_schema, auth)
         if fund == "Towns Fund":
-            # TODO https://dluhcdigital.atlassian.net/browse/SMD-660: use tables to extract, process and validate TF
+            assert isinstance(ingest_dependencies, TFIngestDependencies)
```
samuelhwilliams, line 99: I'm never keen on assertions in production code flows particularly. If we think we need this check, can we swap this for a standard conditional and raise a ValueError or similar instead?

wjrm500: Fair enough, have just made that change now

core/controllers/ingest.py
```diff
@@ -107,15 +106,17 @@ def ingest(body: dict, excel_file: FileStorage) -> tuple[dict, int]:
...
+            tables, p_error_messages = extract_process_validate_tables(
+                workbook_data, ingest_dependencies.extract_process_validate_schema
+            )
+            ct_error_messages = ingest_dependencies.cross_table_validate(tables)
             error_messages = p_error_messages + ct_error_messages
             if error_messages:
                 raise ValidationError(error_messages)
             coerce_data(tables, PF_TABLE_CONFIG)
-            transformed_data = pf_r1_transform(tables, reporting_round)
+            transformed_data = ingest_dependencies.transform(tables, reporting_round)
```
samuelhwilliams, line 119: On the TF side this is called `transform_data`. Can we rationalise the interface - probably in the abstract base class?

wjrm500: Good shout

PR #582, opened 2024-05-14
Stop reading `reporting_round` from submission table

samuelhwilliams:
### Change description
Now that we have the `reporting_round` tracked on the
`programme_junction` table, lets switch all reads from
`submission.reporting_round` over to
`programme_junction.reporting_round`.

As well as changing DB reads, we also stop reading the `Submission_Ref`
dataframe that eventually populates the `submission` table. Instead, we
rely on the `reporting_round` variable passed into `ingest`, and we
cascade this through to `populate_db` so that it can avoid checking the
`Submission_Ref` data frame. This is because we'll shortly stop writing
`submission.reporting_round`, and remove it from the data frame, so
we're just preparing for that.

### DIFF
--- a/core/controllers/ingest.py
+++ b/core/controllers/ingest.py
@@ -140,6 +140,7 @@ def ingest(body: dict, excel_file: FileStorage) -> tuple[dict, int]:
     clean_data(transformed_data)
     if do_load:
         populate_db(
+            reporting_round=reporting_round,
             transformed_data=transformed_data,
             mappings=INGEST_MAPPINGS,
             excel_file=excel_file,
@@ -396,6 +397,7 @@ def get_metadata(transformed_data: dict[str, pd.DataFrame]) -> dict:
 
 @transaction_retry_wrapper(max_retries=5, sleep_duration=0.6, error_type=exc.IntegrityError)
 def populate_db(
+    reporting_round: int,
     transformed_data: dict[str, pd.DataFrame],
     mappings: tuple[DataMapping],
     excel_file: FileStorage,
@@ -417,7 +419,6 @@ def populate_db(
     :param submitting_user_email: The email address of the submitting user.
     :return: None
     """
-    reporting_round = int(transformed_data["Submission_Ref"]["Reporting Round"].iloc[0])
     programme_id = transformed_data["Programme_Ref"]["Programme ID"].iloc[0]
     fund_id = transformed_data["Programme_Ref"]["FundType_ID"].iloc[0]
     programme_exists_previous_round = get_programme_by_id_and_previous_round(programme_id, reporting_round)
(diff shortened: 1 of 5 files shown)


core/controllers/ingest.py
```diff
@@ -396,6 +397,8 @@ def get_metadata(transformed_data: dict[str, pd.DataFrame]) -> dict:
 
 @transaction_retry_wrapper(max_retries=5, sleep_duration=0.6, error_type=exc.IntegrityError)
 def populate_db(
+    *,
```
gidsg, line 400: What's this wildcard doing?

samuelhwilliams: This forces everything following it to be passed in as [keyword-only arguments](https://peps.python.org/pep-3102/).

It's similar to writing `def func(*args, **kwargs)`. In that signature, all positional parameters go into the `args` argument as a tuple.

In `def func(*, **kwargs)` we're saying that we accept no positional arguments, so everything has to go into `kwargs` and be passed as keyword args.

You can also have the best of both worlds, eg `def func(x, y, *, keyword_z)` for example. Here, I think there's no particularly sensible positional args, so I'm forcing everything to be keyword args.

wjrm500 (approved):
Have added one small comment and like Gideon question the asterisk added as the first arg to `populate_db`, but other than that looks good. Approved anyway.

One question though: why are we keeping reporting round as a column on the submission table - would it not be best to drop that column completely?

tests/integration_tests/test_ingest_db_transactions.py
```diff
@@ -737,13 +767,19 @@ def test_delete_existing_submission(test_client_reset, mock_r3_data_dict, mock_e
...
-    populate_db(mock_r3_data_dict, INGEST_MAPPINGS, mock_excel_file, get_table_to_load_function_mapping("Towns Fund"))
+    populate_db(
+        reporting_round=3,
+        transformed_data=mock_r3_data_dict,
+        mappings=INGEST_MAPPINGS,
+        excel_file=mock_excel_file,
+        load_mapping=get_table_to_load_function_mapping("Towns Fund"),
+    )
     # add new round of identical data
     mock_r3_data_dict["Submission_Ref"]["Reporting Round"].iloc[0] = 4
```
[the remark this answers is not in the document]
samuelhwilliams: Yep - we can defo remove this. I'm planning to do the cleanup on `Submission_Ref` in a follow-up PR though, where we stop writing to that dataframe.

samuelhwilliams:
> One question though: why are we keeping reporting round as a column on the submission table - would it not be best to drop that column completely?

Yep, we will be dropping it. I'm chunking this work out into small PRs to keep them easy to review and low line count. I've got the follow up PR basically ready to go once this is merged in that makes the field nullable and stops writing it. Then we can drop the column altogether after that.

core/controllers/ingest.py
wjrm500: Interesting!

PR #584, opened 2024-05-14
FMD-275: Move itl1_region_name to new column

MarcUsher:
### Change description
- Move `itl1_region_name` from the JSONB data_blob to a new `itl1_region_name` column in `GeospatialDim` as the property is accessed by the `get_geospatial_regions filter`.
- Update `itl1_region_name` of Yorkshire to Yorkshire and The Humber to match official ITL1 Region name.
- Update tests and test seed functions to reflect the new column.

### How to test


albertkol (approved):
:+1: Looking good.. Ship it!

db/migrations/versions/034_geospatial_itl1_region_name.py
```diff
@@ -0,0 +1,68 @@
...
+        """
+            UPDATE geospatial_dim
+            SET itl1_region_name = 'Yorkshire and The Humber'
+            WHERE itl1_region_name = 'Yorkshire';
+        """
+    )
+
+    with op.batch_alter_table("geospatial_dim", schema=None) as batch_op:
+        batch_op.alter_column("itl1_region_name", nullable=False)
+        batch_op.drop_column("data_blob")
```
samuelhwilliams, line 43: Please can we split this out into a separate PR/migration. It could cause errors because the existing app deployment will still be trying to read/write `data_blob` until the code changes in this PR are released.

We could probably get away with deploying this, but that's only because our transaction volume is low. It would be a good habit to get into, to always try to create migrations that would still provide zero-downtime/zero-error guarantees for a higher volume service.

To test locally, essentially run the migration from this PR and then switch back to the main branch. If you get errors with the migration applied and the old code running, it's not backwards compatible.

samuelhwilliams (changes requested):

db/migrations/versions/034_geospatial_itl1_region_name.py
MarcUsher: @samuelhwilliams pushed an update to address this, will make a new PR to cover dropping the `data_blob` in `entities.py` and separate migration.

samuelhwilliams (approved):

PR #586, opened 2024-05-14
FMD-275: Drop GeospatialDim data_blob

MarcUsher:
### Change description
Drop the JSONB `data_blob` column from the GeospatialDim table, as the new `itl1_region_name` column now replaces the only field that was present in the `data_blob`.

### How to test


samuelhwilliams (approved):
I think I'm fine with this in this case, but technically could raise errors (probably just in find) during rollout as discussed.

db/migrations/versions/035_drop_geospatial_data_blob.py
```diff
@@ -0,0 +1,44 @@
...
+        batch_op.drop_column("data_blob")
+
+    # ### end Alembic commands ###
+
+
+def downgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    with op.batch_alter_table("geospatial_dim", schema=None) as batch_op:
+        batch_op.add_column(
+            sa.Column("data_blob", postgresql.JSONB(astext_type=sa.Text()), autoincrement=False, nullable=True)
```
wjrm500, line 32: Not sure it matters but original migration (no. 22) defined this column without `autoincrement` argument: `sa.Column("data_blob", postgresql.JSONB(astext_type=sa.Text()), nullable=True)`

wjrm500 (commented):
Not got much to add other than this!

db/migrations/versions/035_drop_geospatial_data_blob.py
MarcUsher: Hmmm good catch, that bit was auto-generated by the alembic migration! I'll try removing it and see if it thinks there are unapplied changes/anything breaks

PR #588, opened 2024-05-15
Split CI tests and db migrations check

samuelhwilliams:
### Change description
Split out the Flask DB migrations check from the unit tests. This will let us mark the unit tests check as required passing, while making the flask db migrations check an optional passing.

This is because in some situations we may intentionally expect a migration to be missing (eg if we're dropping a column, where the python model column definition should be removed and deployed before the migration to actually drop the column is deployed).


.github/workflows/test.yml
```diff
@@ -41,6 +42,45 @@ jobs:
       run: pip install -r requirements-dev.txt
     - name: Run tests
       run: pytest
+  check_db:
```
gidsg, line 45: I think to make this optional it'd need to be a separate job rather than a step. (i.e. move into a separate workflow file)

samuelhwilliams:
@gidsg any thoughts on this one? I'd like to get it in if so. Happy to chat if any concerns

.github/workflows/test.yml
```diff
@@ -41,6 +42,45 @@ jobs:
...
+        image: localstack/localstack:3.2
+        env:
+          # these match the values in config/unit_test.py
+          AWS_REGION: eu-central-1
+          AWS_ACCESS_KEY_ID: test
+          AWS_SECRET_ACCESS_KEY: test
+        ports:
+          - 4566:4566
+          - 4571:4571
+        options: --health-cmd="curl http://localhost:4566/health" --health-interval=10s --health-timeout=10s --health-retries=10
```
gidsg, line 75: Do we actually need localstack for this check?

samuelhwilliams: let's check 👀

gidsg (approved):
👍 LGTM, think I may have to use admin privileges to merge this as the required check won't run.

PR #602, opened 2024-05-29
SMD-732: Accept datetime for Funding Questions' Response

albertkol:
### Change description
Accept datetime for Funding Questions' response.

At the moment we crash in a 500 if the user provides a `datetime` when we expect a `string`, not quite sure why int/float did not.

Updated the excel sheet for the integration test with a situation where one of the Funding Questions' Respones is `datetime`. This will not throw an error. 

### How to test
_Ask for the file I tested with._


tests/integration_tests/mock_tf_returns/TF_Round_4_TD_Funding_Failure.xlsx
samuelhwilliams, line 1: This excel type of the cell in question here is coming through as 'Custom' - if you don't have excel access I may push a commit that makes it 'Date', which I think is more what we want?

samuelhwilliams (approved):

PR #609, opened 2024-06-13
New download endpoint

sc15zs:
https://dluhcdigital.atlassian.net/browse/FPASF-101

Add a new endpoint which emails the user a link to download the report


core/controllers/async_download.py
```diff
@@ -0,0 +1,112 @@
...
+    if file_format == "xlsx":
+        file_format_str = "Microsoft Excel spreadsheet"
+    elif file_format == "json":
+        file_format_str = "JSON file"
+
+    file_size_kb = round(len(response.data) / 1024, 1)
+    if file_size_kb < 1024:
+        file_size_str: str = f"{round(file_size_kb, 1)} KB"
+    else:
+        file_size_str = f"{round(file_size_kb / 1024, 1)} MB"
```
samuelhwilliams, line 46: These could probably be extracted to small helper functions: `get_file_format_from_extension` and `get_human_readable_file_size` or something similar?

core/controllers/async_download.py
```diff
@@ -0,0 +1,112 @@
...
+        # email the URL to the user
+        send_email(
+            email_address=email_address,
+            download_url=url,
+            find_service_url=find_service_base_url + "/download",
+            file_format=file_format_str,
+            file_size_str=file_size_str,
+        )
+
+    return None
```
samuelhwilliams, line 70:
```suggestion
```

I think we can drop this return statement?

core/controllers/async_download.py
```diff
@@ -0,0 +1,112 @@
...
+    file_obj = io.BytesIO(response.data)
+    bucket = Config.AWS_S3_BUCKET_FIND_DATA_FILES
+    find_service_base_url = Config.FIND_SERVICE_BASE_URL
+
+    if not bucket:
+        current_app.logger.error("AWS_S3_BUCKET_FIND_DATA_FILES is not set.")
+    elif not find_service_base_url:
+        current_app.logger.error("FIND_SERVICE_BASE_URL is not set.")
+    else:
+        file_name = f"{str(uuid.uuid4())}.{file_format}"
```
samuelhwilliams, line 58: Can we give this a nicer filename so that when the user downloads it, it's more meaningful?

Maybe just something like `fund-monitoring-data-YYYY-MM-DD-HH:MM:SS.xlsx`? Not the nicest, but better than a random selection of letters and numbers...

core/controllers/async_download.py
```diff
@@ -0,0 +1,112 @@
...
+    :param email_address: user's email address,
+    :param download_url: download URL,
+    :param find_service_url: URL of the FIND service,
+    """
+
+    notifications_client = NotificationsAPIClient(current_app.config["NOTIFY_FIND_API_KEY"])
+
+    notifications_client.send_email_notification(
+        email_address=email_address,
+        template_id="62124580-5d5e-4975-ab84-76d14be2a9ad",
```
samuelhwilliams, line 105: I think this template ID should probably be on our Config object with a sensible name, eg `NOTIFY_TEMPLATE_FIND_DOWNLOAD_REPORT` - what do you think?

core/controllers/async_download.py
```diff
@@ -0,0 +1,112 @@
...
+        file_name = f"{str(uuid.uuid4())}.{file_format}"
+        url = upload_file(file=file_obj, bucket=bucket, object_name=file_name, expiry=86400)
+
+        # email the URL to the user
+        send_email(
+            email_address=email_address,
+            download_url=url,
+            find_service_url=find_service_base_url + "/download",
+            file_format=file_format_str,
+            file_size_str=file_size_str,
```
samuelhwilliams, line 67: Nice work just adding this information immediately 👍  I split it out into a separate ticket but if you're already doing it, I'll get rid of it.

openapi/api.yml
```diff
@@ -85,6 +85,93 @@ paths:
...
+                type: object
+                properties:
+                  file_content:
+                    type: object
+                    description: The file content in json format.
+            application/octet-stream:
+              schema:
+                type: string
+                format: binary
+                description: The file content in a binary format (e.g. XLSX).
```
samuelhwilliams, line 173: I don't think this is right, is it? The `async_download` view doesn't return any data as far as I can see?

samuelhwilliams (commented):
This looks a really solid starting place, not many comments or objections on the code - well done :)

There's some good functionality here that we should add tests around, though. Can you add some tests to check things around:

* Storing a report in S3 in the right bucket
* Sending the email and including the expected personalisation
* 400s from the endpoint for bad requests (ie bad types on query parameters)

Let me know what you think ... I know we're going to change this implementation to use a background queue of some kind soon, so don't want you to test something and immediately throw that test away. We will need solid test coverage before we can release this to users, so maybe we can lay some groundwork here now?

If you've already thought this through and have a plan to manage it / anticipate writing most of the tests later, I can be brought on board though :)

core/controllers/async_download.py
sc15zs: Separated to helper functions.

core/controllers/async_download.py
sc15zs: return statements removed.

core/controllers/async_download.py
sc15zs: File name changed to:
current_datetime = datetime.now().strftime("%Y-%m-%d-%H:%M:%S")
file_name = f"fund-monitoring-data-{current_datetime}.{file_format}"

core/controllers/async_download.py
sc15zs: Template ID has been added to Config object.

openapi/api.yml
sc15zs: Not sure if it's the correct response, but changed it to:
responses:
        '204':
          description: Request has been completed successfully.

config/envs/default.py
```diff
@@ -20,5 +20,6 @@ class DefaultConfig(object):
 
     # Config variables for sending FIND-emails
     NOTIFY_FIND_API_KEY = os.getenv("NOTIFY_FIND_API_KEY")
+    NOTIFY_TEMPLATE_ID = os.getenv("NOTIFY_TEMPLATE_ID")
```
samuelhwilliams, line 23: Sorry, can we be more specific here? `NOTIFY_FIND_REPORT_DOWNLOAD_TEMPLATE_ID`? Just so you can see at a glance what it's used for. We may end up with multiple Notify template references, so just NOTIFY_TEMPLATE_ID is quite vague.

core/controllers/async_download.py
```diff
@@ -0,0 +1,142 @@
...
+        organisations=organisations,
+        regions=regions,
+        rp_start=rp_start,
+        rp_end=rp_end,
+        outcome_categories=outcome_categories,
+    )
+
+    if response.status_code != 200:
+        msg = f"Error downloading data from data-store/download endpoint. Status code: {response.status_code}"
+        current_app.logger.error(msg)
```
samuelhwilliams, line 38: For logging, we shouldn't format the error string up front, as it can mess with Sentry's ability to group errors. Feel free to see examples elsewhere in the app for how to structure log messages. Can I suggest something like this instead, using the `extra` dict from standard logging (note, `extra` dict can only contain basic data types, ie no full response object)?

```suggestion
        current_app.logger.error("Error downloading data from data-store/download endpoint. Status code: {response_status}", extra=dict(response_status=response.status_code))
```

sc15zs: Changed as requested.

core/controllers/async_download.py
```diff
@@ -0,0 +1,147 @@
...
+    bucket = Config.AWS_S3_BUCKET_FIND_DATA_FILES
+    find_service_base_url = Config.FIND_SERVICE_BASE_URL
+    email_template_id = Config.NOTIFY_FIND_REPORT_DOWNLOAD_TEMPLATE_ID
+
+    if not bucket:
+        current_app.logger.error("AWS_S3_BUCKET_FIND_DATA_FILES is not set.")
+    elif not find_service_base_url:
+        current_app.logger.error("FIND_SERVICE_BASE_URL is not set.")
+    elif not email_template_id:
+        current_app.logger.error("NOTIFY_TEMPLATE_ID is not set.")
```
samuelhwilliams, line 57: This is returning a 204 from the endpoint, indicating success. This is fine for now since no (end) users will see it, and we're about to turn this into a background job, but we'll need to consider these error states more thoroughly when this is running on the queue. It should probably fail the task and raise an error to be handled fully rather than just logging an error message.

samuelhwilliams:
Happy with the broad strokes of this code now - let me know when we've added tests and I'll re-review :)

tests/integration_tests/test_async_download.py
```diff
@@ -0,0 +1,93 @@
...
+    WHEN it is successful
+    THEN the function should return True
+    """
+    uploaded_file = io.BytesIO(b"report")
+    key = f"fund-monitoring-data-{datetime.now().strftime(DATETIME_ISO_8601)}"
+    upload_success = upload_file(
+        uploaded_file, Config.AWS_S3_BUCKET_FIND_DATA_FILES, datetime.now().strftime(DATETIME_ISO_8601)
+    )
+    assert upload_success
+    _S3_CLIENT.delete_object(Bucket=Config.AWS_S3_BUCKET_FIND_DATA_FILES, Key=key)
```
samuelhwilliams, line 58: I don't think we need this test - we already have something in `tests.integration_tests.test_aws:test_upload_file`. I don't think this adds anything extra.

tests/integration_tests/test_async_download.py
```diff
@@ -0,0 +1,93 @@
...
+    notifications_client.send_email_notification(
+        email_address="testing@levellingup.gov.uk",
+        template_id=str(uuid.uuid4()),
+        personalisation={
+            "download_url": f"fund-monitoring-data-{current_datetime}",
+            "find_service_url": "fake_url",
+            "file_format": "xlsx | JSON",
+            "file_size": "11kb",
+        },
+    )
```
samuelhwilliams, line 88: This seems to just test that the Notify client works correctly all by itself - I don't think there's a huge amount of value here, as the Notify library already has a suite of tests assuring its functionality.

tests/integration_tests/test_async_download.py
```diff
@@ -0,0 +1,93 @@
...
+
+
+def test_invalid_file_format(test_session):
+    response = test_session.get("/async_download?file_format=invalid")
+    assert response.status_code == 400
```
samuelhwilliams, line 93: 👍

core/controllers/async_download.py
```diff
@@ -0,0 +1,147 @@
...
+
+    _S3_CLIENT.put_object(Bucket=bucket, Key=object_name, Body=file)
+
+    url = _S3_CLIENT.generate_presigned_url(
+        "get_object",
+        Params={"Bucket": bucket, "Key": object_name},
+        ExpiresIn=expiry,
+    )
+
+    return url
```
samuelhwilliams, line 92: This function probably lives better in `core/aws`. Note that there is already an `upload_file` function there - maybe we should reuse that?

This function feels a bit overloaded - it uploads the file _and_ generates a presigned URL. These feel like different things.

I think we can either split them out (just make these two calls in `async_download`, or just rename the function to be a bit clearer? Especially because the current signature is called `upload_file` with an `expiry` argument. That sounds like the file will expire after that long - it doesn't say that's just how long the link will (should) last.

core/controllers/async_download.py
```diff
@@ -0,0 +1,147 @@
...
+        notifications_client.send_email_notification(
+            email_address=email_address,
+            template_id=template_id,
+            personalisation={
+                "download_url": download_url,
+                "find_service_url": find_service_url,
+                "file_format": file_format,
+                "file_size": file_size_str,
+            },
+        )
```
samuelhwilliams, line 121: Can we call this something more specific than `send_email`? I think we can either inline this function, or make it more specific. Maybe something like `send_email_for_find_download`?

Also, this function takes a `template_id` but it's not generic for all templates because we're passing specific personalisation. If we're keeping this function, then maybe just put the template ID directly in the function?

core/controllers/async_download.py
```diff
@@ -0,0 +1,147 @@
...
+    """Return a human-readable file size string.
+    :param file_size_bytes: file size in bytes,
+    :return: human-readable file size,
+    """
+
+    file_size_kb = round(file_size_bytes / 1024, 1)
+    if file_size_kb < 1024:
+        return f"{round(file_size_kb, 1)} KB"
+    else:
+        return f"{round(file_size_kb / 1024, 1)} MB"
```
samuelhwilliams, line 147: Thanks for splitting these out into their own functions - nice 👍  Can we add a few simple tests for them? And for the `get_human_readable_file_size`. It's probably fine that this only does KB and MB but wouldn't be super against adding support for bytes and GB too :) Not compulsory though.

samuelhwilliams:
I don't think we need a test for the end-to-end `async_download` view since we'll be splitting that up soon anyway, so we can just think about the tests we'll use instead in the future tickets.

tests/integration_tests/test_async_download.py
sc15zs: Test removed.

tests/integration_tests/test_async_download.py
sc15zs: Test removed.

core/controllers/async_download.py
sc15zs: Function split into separate functions for file upload and generating the url.

core/controllers/async_download.py
sc15zs: Function renamed, template_id removed from function parameters and template ID has been put directly in the function.

core/controllers/async_download.py
sc15zs: Return option added for 'GB', unit tests have been added for both function.

core/controllers/async_download.py
```diff
@@ -0,0 +1,152 @@
...
+    # Upload the file to S3 and get presigned URL
+    file_obj = io.BytesIO(response.data)
+    bucket = Config.AWS_S3_BUCKET_FIND_DATA_FILES
+    find_service_base_url = Config.FIND_SERVICE_BASE_URL
+
+    if not bucket:
+        current_app.logger.error("AWS_S3_BUCKET_FIND_DATA_FILES is not set.")
+    elif not find_service_base_url:
+        current_app.logger.error("FIND_SERVICE_BASE_URL is not set.")
+    else:
```
samuelhwilliams, line 55: I'd rather the app refused to start up if these vars are missing. Better for the app to not start than for it to start running and be misconfigured (unable to serve this endpoint), I think.

But this might be one to address separately as at the moment we're very forgiving on all of our env vars/config 🤔  I will make a ticket.

tests/integration_tests/test_async_download.py
```diff
@@ -0,0 +1,92 @@
...
+        yield
+
+
+@pytest.fixture()
+def mocked_notify_success():
+    notifications_client = Mock(spec=NotificationsAPIClient)
+    notifications_client.send_email_notification.return_value = (
+        "success",
+        200,
+    )
```
samuelhwilliams, line 58: I don't think any of this is used any more?

copilot/data-store/manifest.yml
```diff
@@ -36,20 +36,26 @@ variables:                    # Pass environment variables as key value pairs.
   FLASK_ENV: ${COPILOT_ENVIRONMENT_NAME}
   # Sentry DSN is OK to be public see: https://docs.sentry.io/product/sentry-basics/dsn-explainer/#dsn-utilization
   SENTRY_DSN: https://6a2623e302e641ba88eabe6675a70ddf@o1432034.ingest.sentry.io/4505390859747328
-#secrets:                      # Pass secrets from AWS Systems Manager (SSM) Parameter Store.
+secrets:                      # Pass secrets from AWS Systems Manager (SSM) Parameter Store.
+    NOTIFY_FIND_API_KEY: /copilot/${COPILOT_APPLICATION_NAME}/${COPILOT_ENVIRONMENT_NAME}/secrets/NOTIFY_FIND_API_KEY
+    NOTIFY_FIND_REPORT_DOWNLOAD_TEMPLATE_ID: /copilot/${COPILOT_APPLICATION_NAME}/${COPILOT_ENVIRONMENT_NAME}/secrets/NOTIFY_FIND_REPORT_DOWNLOAD_TEMPLATE_ID
```
samuelhwilliams, line 41: I've just noticed this is a secret value - it doesn't need to be, we can have it in the `variables` block above

core/controllers/async_download.py
sc15zs: Removed these logging lines.

core/controllers/async_download.py
sc15zs: Removed these logging lines.

tests/integration_tests/test_async_download.py
sc15zs: Removed these test bits.

copilot/data-store/manifest.yml
sc15zs: Removed NOTIFY_FIND_REPORT_DOWNLOAD_TEMPLATE_ID from here and added the to DefaultConfig with real value.

samuelhwilliams (dismissed):
I think this is good to merge now, once the commits are squashed and we have the PR on [docker-runner](https://github.com/communitiesuk/funding-service-design-post-award-docker-runner/) adding the new env var ready to go :)

core/controllers/async_download.py
```diff
@@ -0,0 +1,150 @@
...
+        presigned_url = generate_find_presigned_url(bucket=bucket, object_name=file_name, expiry=86400)
+
+        # email the URL to the user
+        send_email_for_find_download(
+            email_address=email_address,
+            download_url=presigned_url,
+            find_service_url=find_service_download_url,
+            file_format=file_format_str,
+            file_size_str=file_size_str,
+        )
```
samuelhwilliams, line 66:
```suggestion
    find_service_download_url = Config.FIND_SERVICE_BASE_URL + "/download"

    upload_find_file(file=file_obj, bucket=bucket, object_name=file_name)
    presigned_url = generate_find_presigned_url(bucket=bucket, object_name=file_name, expiry=86400)

    # email the URL to the user
    send_email_for_find_download(
        email_address=email_address,
        download_url=presigned_url,
        find_service_url=find_service_download_url,
        file_format=file_format_str,
        file_size_str=file_size_str,
    )
```

I don't think we need the conditional here. If `FIND_SERVICE_BASE_URL` is None, it's probably better for us to just error rather than silently do nothing. That way it will get reported to sentry and we can fix it. If we do nothing, we won't find out until a user gets in contact to ask why they haven't received anything.

sc15zs: Changed back to as it was previously.

samuelhwilliams (approved):

PR #633, opened 2024-07-09
Fpasf 153 download report

israr-ulhaq:
https://dluhcdigital.atlassian.net/browse/FPASF-153

Created two end points, one will fetch the file metdata from the S3.
Second end point will create a presigned url for a short time if file exists.


core/aws.py
```diff
@@ -25,7 +25,7 @@
     )
 
 
-def upload_file(file: IO | FileStorage, bucket: str, object_name: str, metadata: dict | None = None) -> bool:
+def upload_file(file: Union[IO, FileStorage], bucket: str, object_name: str, metadata: dict | None = None) -> bool:
```
samuelhwilliams, line 28: question: `IO | FileStorage` should be the same as `Union[IO, FileStorage]` - did this need to change?

tests/controller_tests/test_download.py
```diff
@@ -94,3 +98,46 @@ def test_sort_columns_exception(test_session):
...
+    assert response.status_code == 200
+    assert "content_length" in response.json
+    assert "content_type" in response.json
+    assert "last_modified" in response.json
+
+
+def test_get_find_download_file_not_exist(test_session):
+    filename = "example.xlsx"
+    response = test_session.get(f"/get-find-download-metadata/{filename}")
+    print("response status", response)
```
samuelhwilliams, line 120:
```suggestion
```

tests/controller_tests/test_download.py
```diff
@@ -94,3 +98,46 @@ def test_sort_columns_exception(test_session):
...
+    filebytes = b"example file contents"
+    file = FileStorage(io.BytesIO(filebytes), filename=filename, content_type=EXCEL_MIMETYPE)
+
+    upload_file(file=file, bucket=Config.AWS_S3_BUCKET_FIND_DATA_FILES, object_name=filename)
+
+    response = test_session.get(f"/get-find-download-metadata/{filename}")
+    assert response.status_code == 200
+    assert "content_length" in response.json
+    assert "content_type" in response.json
+    assert "last_modified" in response.json
```
samuelhwilliams, line 114: Can we please move these to `integration_tests`, since we're testing actual communication to/from S3 (localstack)?

core/controllers/async_download.py
```diff
@@ -91,69 +91,74 @@ def async_download(
...
-    :return: presigned URL,
-    """
+    try:
+        # Check if the object (file) exists in S3
+        _S3_CLIENT.head_object(Bucket=Config.AWS_S3_BUCKET_FIND_DATA_FILES, Key=filename)
+    except ClientError as error:
+        if error.response["Error"]["Code"] == "404":
+            return abort(404, f"Could not find file: {filename} on S3.")
+        else:
+            return abort(500, f"Error checking object: {filename} existence in S3")
```
samuelhwilliams, line 147: We should ideally have a consistent data type as the return value. We generally don't just return raw strings in JSON responses - can we at least wrap these in a dictionary, like the "happy path" response?

Something like: `return jsonify({"error": {"code": "file-not-found", "message": f"Could not find file {filename} in S3."}}), 404`

core/controllers/async_download.py
```diff
@@ -167,36 +172,5 @@ def send_email_for_find_download(
...
-    :return: human-readable file size,
-    """
-
-    file_size_kb = round(file_size_bytes / 1024, 1)
-    if file_size_kb < 1024:
-        return f"{round(file_size_kb, 1)} KB"
-    elif file_size_kb < 1024 * 1024:
-        return f"{round(file_size_kb / 1024, 1)} MB"
-    else:
-        return f"{round(file_size_kb / (1024 * 1024), 1)} GB"
```
samuelhwilliams, line 202: Have you moved these to the data frontend, or are we just removing them altogether?

samuelhwilliams: I'd be tempted to leave them here because some of your other tests are now broken. We can call them in the metadata endpoint to return human-readable values instead of raw content_length and content_type values?

samuelhwilliams (approved):

PR #641, opened 2024-07-25
The making of a monolith

samuelhwilliams:
Pulls in the latest changes from https://github.com/communitiesuk/funding-service-design-post-award-data-frontend/ and https://github.com/communitiesuk/funding-service-design-post-award-submit/, and then integrates both of those services into the existing `data-store` service, creating a single "monolith"-style service.

FPASF-249


gidsg (approved):
LGTM 😅 👍

PR #643, opened 2024-07-26
Point to docker-runner at the top; fix some markdown linting issues

samuelhwilliams:
Update the README to point at docker-runner immediately, as that's now the recommended/"officially supported" way to run the service(s).


israr-ulhaq (approved):
Looks great to me!

PR #646, opened 2024-07-30
Make Find work without data

albertkol:
Two changes:
1. Make filter functions return empty lists or None if no data is found. Now it's throwing exceptions.
2. Make seeder also insert example data

### How to test
```bash
docker compose down
docker compose up

# Browse to Find /download
Does it work?

### DIFF
--- a/data_store/cli.py
+++ b/data_store/cli.py
@@ -45,7 +45,7 @@ def create_cli(app):
         print("Reference data seeded successfully.")
 
     @database_cli.command("seed-sample-data")
-    def seed_data():
+    def seed_sample_data():
         """Seed the database with sample data.
 
         Example usage:
--- a/data_store/cli.py
+++ b/data_store/cli.py
@@ -41,9 +41,20 @@ def create_cli(app):
         with current_app.app_context():
             seed_geospatial_dim_table()
             seed_fund_table()
-            load_example_data()
 
         print("Reference data seeded successfully.")
+
+    @database_cli.command("seed-sample-data")
+    def seed_data():
+        """Seed the database with sample data.
+
+        Example usage:
+            flask db-data seed-sample-data
+        """
+        with current_app.app_context():
+            load_example_data()
+
+        print("Sample data seeded successfully.")
 
     @database_cli.command("reset")
     def reset():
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: data_store/cli.py [hunk 1]
```diff
@@ -40,6 +41,7 @@ def seed_reference():
         with current_app.app_context():
             seed_geospatial_dim_table()
             seed_fund_table()
+            load_example_data()
```
MarcUsher, line 44: Nice fix! I would say there's probably value in having the `load_example_data()` get called in a separate CLI command so that this one stays distinct to just seeding reference data (and then updating the docker-runner to call both on start-up). There might be scenarios where we want to seed or update just the reference dat.

samuelhwilliams, line 44: I don't think this should be part of the `seed-ref` command - it's not reference data.

We could add/restore a separate command for loading this example data, and then add that to the docker runner startup.

albertkol: Started making the change, but now I notice that `flask db-data reset` populates back only reference data.

Hmm.. there are a number of ways to go about it:
1. Change `seed_reference` to `seed_data` and have it do both ref and data seeding. Make reset add both back.
2. Make command accept a parameter if it should populate/reset sample data with the ref
3. Duplicate commands. Separate seeding commands for ref/data and reset ref/data.

Thoughts?

MarcUsher: My preference would be for option 3. Reasoning being:

- Keeps things separate - if we want to seed, we have commands to do it, and if we want to reset, we have commands to do it.
- Leaves `flask db-data reset` as is, as I think the intention with that command was to reset the DB into a useable state (ie. empty but with the necessary reference data seeded so you could start to ingest spreadsheets). I don't know if reset should also seed non-ref data, but happy to be convinced otherwise - logically it makes sense to me that if you wanted to do that you'd reset and then run whatever new seed data command you add.

data_store/cli.py
```diff
@@ -43,6 +44,18 @@ def seed_reference():
 
         print("Reference data seeded successfully.")
 
+    @database_cli.command("seed-sample-data")
```
MarcUsher, line 47: Nice! Do you mind just updating the README to include this as well as it has a list of all the CLI commands that you can run?

albertkol: Added! :+1:

MarcUsher (approved):
Looks good, think you just need a rebase 🙌

PR #650, opened 2024-07-31
BAU: Add additional auth check for Find

MarcUsher:
### Change description
Adds an additional check on all Find routes to ensure only users with `@communities.gov.uk` or `@test-communities.gov.uk` can access the service. Adds unit test to cover this behaviour enforcenment.

Also consolidates all error templates into the `common` templates folder, removing them from Find & Submit modules.

### How to test


find/const.py
```diff
@@ -4,3 +4,8 @@
 class MIMETYPE(StrEnum):
     XLSX = "application/vnd.openxmlformats-officedocument.spreadsheetml.sheet"
     JSON = "application/json"
+
+
+class INTERNAL_DOMAIN(StrEnum):
+    COMMUNITIES = "@communities.gov.uk"
+    TEST_COMMUNITIES = "@test-communities.gov.uk"
```
samuelhwilliams, line 11:
```suggestion
    TEST_COMMUNITIES = "@test.communities.gov.uk"
```

find/main/decorators.py
```diff
@@ -0,0 +1,19 @@
...
+
+from flask import abort, g
+
+from find.const import INTERNAL_DOMAIN
+
+
+def check_internal_user(func):
+    @wraps(func)
+    def decorated_function(*args, **kwargs):
+        internal_domains = tuple(INTERNAL_DOMAIN._value2member_map_.keys())
```
samuelhwilliams, line 11: I think we can avoid using internal attributes

```suggestion
        internal_domains = tuple(k.value for k in InternalDomain)
```

tests/conftest.py
```diff
@@ -671,6 +684,18 @@ def unauthenticated_find_test_client() -> Generator[FlaskClient, None, None]:
...
+@pytest.fixture(scope="function")
+def non_internal_user_find_test_client(mocked_pf_auth) -> Generator[FlaskClient, None, None]:
+    """
+    :return: A flask test client authenticated for an external user
+    ie. without @communities.gov.uk or @test-communities.gov.uk email.
+    """
+    app = create_app(config.Config)
+    app.test_client_class = _FindFlaskClient
+    with app.test_client() as test_client:
+        yield test_client
```
gidsg, line 696: Where are we specifying the email for this user? Would be goods if they can have a user explicitly auth'd with a non-communities email?

MarcUsher: Yes it's using the `mocked_pf_auth` fixture which mocks an authorised user with a non-communities email

find/main/decorators.py
MarcUsher: Thanks @samuelhwilliams, just amended and pushed the last commit to include these suggested changes 👍

common/templates/common/403.html
```diff
@@ -0,0 +1,19 @@
...
+
+{% set mainClasses = "govuk-main-wrapper--l" %}
+
+{% block content %}
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <h1 class="govuk-heading-l">Sorry, you don't currently have permission to access this service</h1>
+    <p class="govuk-body">
+      If you believe this is incorrect, please email us on
+      <a href="mailto:{{ config['CONTACT_EMAIL'] }}">{{ config['CONTACT_EMAIL'] }}</a>
```
samuelhwilliams, line 15: tiny nit just to fix some styling (same in a few other places) 🙏 

```suggestion
      <a class="govuk-link" href="mailto:{{ config['CONTACT_EMAIL'] }}">{{ config['CONTACT_EMAIL'] }}</a>
```

app.py
```diff
@@ -177,17 +177,13 @@ def http_exception(error):
...
-            folder_name = "submit"
+        error_templates = [401, 403, 404, 429, 500, 503]
 
         if error.code in error_templates:
-            return render_template(f"{folder_name}/main/{error.code}.html"), error.code
+            return render_template(f"common/{error.code}.html"), error.code
         else:
             flask_app.logger.info("Unhandled HTTP error {error_code} found.", extra=dict(error_code=error.code))
-            return render_template("find/main/500.html"), error.code
+            return render_template("common/500.html"), error.code
```
gidsg, line 186: Should we extract this error related stuff into a separate PR as its unrelated to auth?

MarcUsher: Yes sure we can do, though there is some relation - Find (and Submit) didn't have a template for 403s previously so needed a new template and it wasn't one of the error codes we'd catch and serve a specific template for, it would just show the 500 previously.

find/main/routes.py
```diff
@@ -45,6 +48,7 @@ def index():
 
 @bp.route("/start", methods=["GET", "POST"])
 @login_required(return_app=SupportedApp.POST_AWARD_FRONTEND)
```
gidsg, line 50: Can we make a backlog task to move this into utils and have an optional argument to specify an email domain restriction?

MarcUsher: Have added the [backlog task here](https://dluhcdigital.atlassian.net/browse/FPASF-474)

common/templates/common/403.html
MarcUsher: Cheers Sam, nice catch, have updated on the common templates 👍

find/main/decorators.py
```diff
@@ -0,0 +1,19 @@
...
+
+from find.const import InternalDomain
+
+
+def check_internal_user(func):
+    @wraps(func)
+    def decorated_function(*args, **kwargs):
+        internal_domains = tuple(k.value for k in InternalDomain)
+        authenticated = g.is_authenticated
+        is_communities = g.is_authenticated and g.user.email.endswith(internal_domains)
```
samuelhwilliams, line 13: I think this `.endswith` check is broken - checking if a string ends with a tuple of domains?

Should it be something like this instead?

```suggestion
        is_communities = g.is_authenticated and any(g.user.email.endswith(domain) for domain in internal_domains)
```

samuelhwilliams: Huh, TIL that `.endswith` [accepts a tuple of suffixes to look for](https://docs.python.org/3/library/stdtypes.html#str.endswith). That explains why the tests pass.

Carry on!

samuelhwilliams (approved):

PR #651, opened 2024-07-31
Add exception for missing postcode data

samuelhwilliams:
### Change description
`add_project_geospatial_relationship` is in our `data_store` module, which should need no concept of HTTP requests. This function, however, uses flask's `abort` call in situations where an error has happened. Instead, we should just raise an exception, which will bubble up to the relevant view and then be handled by flask as a 500. (Or it could be caught and handled appropriately)

https://dluhcdigital.atlassian.net/browse/FPASF-288

### DIFF
--- a/data_store/controllers/load_functions.py
+++ b/data_store/controllers/load_functions.py
@@ -5,7 +5,6 @@ as well as helper functions for loading.
 """
 
 import pandas as pd
-from flask import abort
 
 from data_store.const import SUBMISSION_ID_FORMAT
 from data_store.controllers.mappings import DataMapping
@@ -15,6 +14,7 @@ from data_store.db.queries import (
     get_latest_submission_by_round_and_fund,
     get_organisation_exists,
 )
+from data_store.exceptions import MissingGeospatialException
 from data_store.util import get_postcode_prefix_set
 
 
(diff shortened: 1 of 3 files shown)


data_store/controllers/load_functions.py
```diff
@@ -325,8 +324,12 @@ def add_project_geospatial_relationship(project_models: list[db.Model]) -> list[
...
         sorted_failing_postcode_prefixes = sorted(list(failing_postcode_prefixes))
-        return abort(
-            500, f"Postcode prefixes not found in geospatial table: {', '.join(sorted_failing_postcode_prefixes)}"
-        )
+        raise MissingGeospatialException(sorted_failing_postcode_prefixes)
 
     return project_models
+
+
+class MissingGeospatialException(Exception):
```
wjrm500, line 332: Might be better off in an `exceptions.py` module?

wjrm500 (commented):
Makes sense, raising `ValueError` could be a simpler alternative but custom class offers benefits also so happy with that. Only thing less convinced on is having the new custom exception at the bottom of the module, rather than in a separate module for exceptions.

data_store/controllers/load_functions.py
samuelhwilliams: Moved 👍

wjrm500 (approved):

PR #658, opened 2024-08-05
Dynamically set viewport size for playwright tests

samuelhwilliams:


MarcUsher (approved):
LGTM! Nice thing to add. If we ever start adding screenshot comparisons we'll have to make sure they save and check against the same sized viewport as is passed in the argument. 

Just added one small comment that you can feel free to ignore!

PR #661, opened 2024-08-06
Create set-users-with-roles-command

albertkol:
### Change description
Assign roles to the users from the csv.

File example:
```
# users.csv

email,full_name,azure_ad_subject_id
aaa@g.c,Aaa Bbb,12345678-1234-1234-1234-1234567890100
ccc@y.c,Ccc Ddd,12345678-1234-1234-1234-1234567890111
```

```bash
flask admin add-roles-to-users --filepath <path> --roles <role1>,<role2>
```

# Screenshots

Screenshot from 2024-08-12 22-27-23

Screenshot from 2024-08-12 22-22-48


data_store/cli.py
```diff
@@ -161,5 +164,93 @@ def reingest_files_from_s3(filepath):
                 pd.set_option("display.max_colwidth", None)
                 print(reingest_outputs)
 
+    @admin_cli.command("set-user-with-roles")
```
samuelhwilliams, line 167:
```suggestion
    @admin_cli.command("add-roles-to-users")
```

maybe?

data_store/cli.py
```diff
@@ -161,5 +164,93 @@ def reingest_files_from_s3(filepath):
...
+                        Config.ACCOUNT_STORE_API_HOST + Config.ACCOUNT_ENDPOINT.format(account_id=account_id),
+                        headers={"Content-Type": "application/json"},
+                        json={k: v for k, v in data.items() if v is not None},
+                    )
+
+                    response.raise_for_status()
+                except requests.exceptions.RequestException as e:
+                    click.echo(f"Update account for {email} error: {e.response.text}")
+
+                click.echo(f"{email} has been assigned the roles: {', '.join(roles)}")
```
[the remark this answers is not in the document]
albertkol: Got it to do something like this:
```
Result:
7 accounts created with the roles: ROLE_1
7 accounts updated with the roles: ROLE_1
No errors occurred.
```
or
```
Result:
1 accounts created with the roles:
0 accounts updated with the roles:
3 accounts failed:
-  albertartk@gmail.com: {"type": "about:blank", "title": "Bad Request", "detail": "'roles' is a required property", "status": 400}
-  albertartk1000@gmail.com: {"type": "about:blank", "title": "Bad Request", "detail": "'roles' is a required property", "status": 400}
-  albertartk100@gmail.com: {"type": "about:blank", "title": "Bad Request", "detail": "'roles' is a required property", "status": 400}
```

data_store/cli.py
albertkol: Changed.

data_store/cli.py
```diff
@@ -21,169 +23,188 @@
...
+            print(reingest_outputs)
+
+
+@admin_cli.command("add-roles-to-users")
+@click.option("--filepath", required=True, type=click.Path(exists=True, dir_okay=False, file_okay=True))
+@click.option("--roles", multiple=True, help="List of roles comma separated to assign to each account")
+def add_roles_to_users(filepath, roles):
+    """Assign roles to the users from the csv.
+
+    :param filepath (list):  Path to csv file containing email and full name
```
samuelhwilliams, line 182:
```suggestion
    :param filepath:  Path to csv file containing email and full name
```

README.md
```diff
@@ -203,10 +203,23 @@ flask admin reingest-s3 <filepath>
...
+```
+# users.csv
+
+-----------------------------
+|    email   |   fullname   |
+-----------------------------
+|   aaa@g.c  |   Aaa Bbb    |
+-----------------------------
+|   ccc@y.c  |   Ccc Ddd    |
+-----------------------------
```
samuelhwilliams, line 218:
```suggestion
email,full_name
aaa@g.c,Aaa Bbb
ccc@y.c,Ccc Ddd
```

samuelhwilliams:
Can we add a default value for `ACCOUNT_STORE_API_HOST` - either in `default.py` or in `development.py` config?

samuelhwilliams:
That looks correct to me - worth deploying to dev and trying there if you haven't already 👍

data_store/cli.py
```diff
@@ -18,148 +23,276 @@
...
+
+                        # Go to next email
+                        continue
+
+                # Update account with roles
+                data = {
+                    "email_address": email,
+                    "azure_ad_subject_id": account.get("azure_ad_subject_id"),
+                    "full_name": full_name,
+                    "roles": roles,
```
samuelhwilliams, line 262: This doesn't add to the roles - it overwrites them. Is that what you're expecting to do? I think we discussed making it the combination of roles rather than overwriting.

If it removes then ... maybe `set-roles-for-users` **is** better than `add-roles-to-users`?

data_store/cli.py
```diff
@@ -18,148 +23,276 @@
...
+                }
+
+                account_id = account.get("account_id")
```
samuelhwilliams, line 265: I think this should be a hard lookup - if there's no account ID, we can't make the API call?

```suggestion
                account_id = account["account_id"]
```

data_store/cli.py
```diff
@@ -18,148 +23,276 @@
...
+                        headers={"Content-Type": "application/json"},
+                        json={k: v for k, v in data.items() if v is not None},
+                    )
+
+                    response.raise_for_status()
+
+                    result["updated"] += 1
+                except requests.exceptions.RequestException as e:
+                    result["errors"][email] = e.response.text
+                    click.echo(f"Update account for {email} error: {e.response.text}")
```
samuelhwilliams, line 279: We're missing a `continue` here that the other `except` blocks have

data_store/cli.py
albertkol: So.. I think we can do it either way, especially now that this script will be used mainly for testing purposes.

I think setting the roles would be useful in more circumstances, as it deletes and creates the roles. Adding does not enable us to remove existing roles.

What do you think?

samuelhwilliams: I think setting is fine - but maybe it should echo the roles that are removed just so there's some chance of catching a mistake?

config/envs/default.py
```diff
@@ -30,6 +30,11 @@ class DefaultConfig(object):
     AWS_S3_BUCKET_FIND_DOWNLOAD_FILES = os.getenv("AWS_S3_BUCKET_FIND_DOWNLOAD_FILES")
     FIND_SERVICE_BASE_URL = os.getenv("FIND_SERVICE_BASE_URL")
 
+    # Config variables for ACCOUNT STORE
+    ACCOUNT_STORE_API_HOST = os.getenv("ACCOUNT_STORE_API_HOST")
+    ACCOUNTS_ENDPOINT = "/accounts"
+    ACCOUNT_ENDPOINT = "/accounts/{account_id}"
```
gidsg, line 36: Can we just hard-code these as they're not going to vary by environment?

config/envs/default.py
```diff
@@ -30,6 +30,11 @@ class DefaultConfig(object):
     AWS_S3_BUCKET_FIND_DOWNLOAD_FILES = os.getenv("AWS_S3_BUCKET_FIND_DOWNLOAD_FILES")
     FIND_SERVICE_BASE_URL = os.getenv("FIND_SERVICE_BASE_URL")
 
+    # Config variables for ACCOUNT STORE
+    ACCOUNT_STORE_API_HOST = os.getenv("ACCOUNT_STORE_API_HOST")
```
gidsg, line 34: Can we default this to whatever the hostname is when running it using the docker runner?

copilot/post-award-celery/manifest.yml
```diff
@@ -40,6 +40,7 @@ variables:
     from_cfn: ${COPILOT_APPLICATION_NAME}-${COPILOT_ENVIRONMENT_NAME}-postawardsuccessfulfilesBucketName
   AWS_S3_BUCKET_FAILED_FILES:
     from_cfn: ${COPILOT_APPLICATION_NAME}-${COPILOT_ENVIRONMENT_NAME}-postawardfailedfilesBucketName
+  ACCOUNT_STORE_API_HOST: "http://fsd-account-store:8080"
```
gidsg, line 43: Don't think this is needed here as Celery instance shouldn't be running this script?

data_store/cli.py
albertkol: Changed name back to `set-roles-for-users`. Roles are overwritten. Now it echoes roles that will be removed.

data_store/cli.py
albertkol: Fixed! :+1:

config/envs/default.py
albertkol: Hard-coded them. It's what `authtenticator` does, so stuck with it initially.

config/envs/default.py
albertkol: Defaults to "http://localhost:3003" in `default.py`. Removed `ACCOUNT_STORE_API_HOST` from `development.py`.

copilot/post-award-celery/manifest.yml
albertkol: I removed it from Celery manifest.
I thought we want to keep these consistent. But it's likely not going to be needed on Celery.

albertkol:
I hope you guys like colours! :sparkles: :sparkles: :sparkles: 

Screenshot from 2024-08-12 22-27-23
Screenshot from 2024-08-12 22-22-48

samuelhwilliams:
beautiful ;)

data_store/cli.py
```diff
@@ -18,148 +23,292 @@
...
+                        result["errors"].append(error_message)
+                        click.secho(error_message, fg="red")
+
+                        # Go to next email
+                        continue
+
+                # Update account with roles
+                data = {
+                    "email_address": email,
+                    "azure_ad_subject_id": account.get("azure_ad_subject_id"),
```
samuelhwilliams, line 269: If the `account` retrieved doesn't have `azure_ad_subject_id` - then this call fails? If `azure_ad_subject_id` is required in the CSV, then that should probably be used as a fallback value here for if the account already exists?

It is possible for accounts to exist without `azure_ad_subject_id` being set (whether that's a valid state in our system, I'm less sure - but I believe so).

copilot/post-award-celery/manifest.yml
samuelhwilliams: My preference would be for all of the config to be consistently applied between both apps 🤷‍♂️  They are the same codebase - it feels likely to just trip people up in the future if config is different between them (or some values are just unset).

data_store/cli.py
```diff
@@ -18,148 +23,292 @@
...
+    click.secho(f"Assigning roles: {roles}", fg="green")
+    click.secho("\nDetails:", fg="blue")
+
+    with current_app.app_context():
+        with click.open_file(filepath) as csvfile:
+            reader = csv.DictReader(csvfile)
+            for row in reader:
+                email = row["email"]
+                full_name = row["full_name"]
+                azure_ad_subject_id = row["azure_ad_subject_id"]
```
gidsg, line 207: Should we default this if we don't care about its value most of the time?

data_store/cli.py
samuelhwilliams: I think I just find this whole thing around `azure_ad_subject_id` a bit weird.

If it's not already on the account in account-store, how are we going to have it to put in our CSV? And if we don't have it in either place, then the PUT just won't go through? I feel like we either need to be able to create accounts with roles in one go, or the update account PUT endpoint should not require `azure_ad_subject_id`.

I know you and @gidsg spoke about this already - what am I missing? It feels like I must be missing something.

users.csv
```diff
@@ -0,0 +1,6 @@
+email,full_name,azure_ad_subject_id
+albert@communities.gov.uk,Albert,12345678-1234-1234-1234-1234567890199
+albert2@communities.gov.uk,Albert T,12345678-1234-1234-1234-123456789013
+albert5@communities.gov.uk,Albert T,12345678-1234-1234-1234-123456789016
+albert4@communities.gov.uk,Albert G,12345678-1234-1234-1234-123456789014
+albert3@communities.gov.uk,Albert G,12345678-1234-1234-1234-123456789015
```
samuelhwilliams, line 6: oops?

albertkol: :hankey:

data_store/cli.py
albertkol: Azure_ad_subsject_id defaults to `uuid4`, if not provided.
full_name defaults to empty string.

data_store/cli.py
albertkol: I changed this back. Now that `azure_ad_subject_id` defaults to random uuid, falling back on that value will result in a broken request. 

I think there are work arounds, but until we use this for non-test purposes, it won't affect us.

users.csv
albertkol: I pushed that file twice by accident no matter how careful I was. :cry:

samuelhwilliams (approved):
lgtm 👀

PR #662, opened 2024-08-07
FPASF-461: Update deployment workflows

MarcUsher:
### Change description
Implements a single pipeline that builds, tags (using the commit SHA rather than the branch name or `latest`), pushes and deploys to the `test` environment when PRs are merged into `main` and then just deploys the same image to production (`prod`) after approval. 

Also updates the old manual branch deployment workflow to make use of the new reusable workflow for consistency and also (for now) removes the ability to deploy to prod from this workflow to avoid clashes.

Testing still needed on how this all works in practice, which we can't do until this gets merged in and we can observe the behaviour 😬, but testing of the reusable workflow has been done with the manual deploy to `dev`.

### How to test

### DIFF
--- a/.github/workflows/deployment_cd_test_prod.yml
+++ b/.github/workflows/deployment_cd_test_prod.yml
@@ -1,0 +1,72 @@
+name: Deploy post-award services to AWS
+run-name: Deploy branch ${{ github.ref_name }} to test and prod environments (${{ format('SHA:{0}', github.sha) }})
+on:
+  push:
+    branches:
+      - main
+
+env:
+  ENVIRONMENT: test
+
+concurrency:
+      group: deploy-test-prod
+      cancel-in-progress: false
+
+jobs:
+  paketo_build:
+    permissions:
+      packages: write
+    concurrency: build-${{ github.sha }}
+    uses: communitiesuk/funding-service-design-workflows/.github/workflows/package.yml@main
+    with:
+      assets_required: true
+      version_to_build: ${{ github.sha }}
+      owner: ${{ github.repository_owner }}
+      application: ${{ github.event.repository.name }}
+
+  copilot_test_deploy:
+    needs: [ paketo_build ]
+    # concurrency:
+    #   group: deploy-$ENVIRONMENT
+    #   cancel-in-progress: false
+    strategy:
+      matrix:
+        include:
+          - deployment: post-award
+            command: svc
+          - deployment: post-award-celery
+            command: svc
+          - deployment: download-report
+            command: job
+    uses: ./.github/workflows/aws_copilot_deployment.yml
+    with:
+      copilot_environment: $ENVIRONMENT
+      github_sha: ${{ github.sha }}
+      deployment: ${{ matrix.deployment }}
+      command: ${{ matrix.command }}
+      IMAGE_LOCATION: ghcr.io/${{ github.repository_owner }}/${{ github.event.repository.name }}:${{ github.sha }}
+    secrets: inherit
+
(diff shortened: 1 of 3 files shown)


.github/workflows/deployment_cd_test_prod.yml
```diff
@@ -0,0 +1,63 @@
...
+    needs: [ paketo_build ]
+    strategy:
+      matrix:
+        include:
+          - deployment: post-award
+            command: svc
+          - deployment: post-award-celery
+            command: svc
+          - deployment: download-report
+            command: job
```
samuelhwilliams, line 34: Can we move this matrix into `aws_copilot_deployment.yml` - then we don't need to duplicate it below?

.github/workflows/deployment_cd_test_prod.yml
```diff
@@ -0,0 +1,63 @@
+name: Deployment pipeline for post-award services
+run-name: Deploy branch ${{ github.ref_name }} to test and prod (${{ format('SHA:{0}', github.sha) }})
+on:
+  push:
+    branches:
+      - main
+
+concurrency:
+  group: deploy-test
+  cancel-in-progress: false
```
samuelhwilliams, line 10: Do we want this here? I think we have the concurrency defined inside `copilot_test_deploy` and `copilot_prod_deploy` (plus the change below) then we get the concurrency behaviour we want (all matrix jobs running at once, but only one set of matrix jobs running at a time in an environment)?

.github/workflows/deployment_cd_test_prod.yml
```diff
@@ -0,0 +1,63 @@
...
+            command: svc
+          - deployment: download-report
+            command: job
+    uses: ./.github/workflows/aws_copilot_deployment.yml
+    with:
+      copilot_environment: test
+      github_sha: ${{ github.sha }}
+      deployment: ${{ matrix.deployment }}
+      command: ${{ matrix.command }}
+      IMAGE_LOCATION: ghcr.io/${{ github.repository_owner }}/${{ github.event.repository.name }}:${{ github.sha }}
```
samuelhwilliams, line 41: If this is the same in both environments, should it not be an input and instead just be built in `aws_copilot_deployment.yml` job?

MarcUsher: Yes you're right, the called workflow should have access to the github context so this should be doable with the change above! I've pushed a commit that makes these changes.

.github/workflows/deployment_cd_test_prod.yml
MarcUsher: This is the bit I'm not sure about so yes good to test - I've added a commit that makes this change but I'm not entirely sure how adding concurrency in a job that uses a reusable workflow that has the matrix in will work, if it'll block the jobs as if the matrix was set up in the calling workflow job. Have added a commit that makes this change and can test with the original branch deploy workflow to see what happens!

MarcUsher: Check you out genius Sam, that does work 😄

samuelhwilliams: Nice. I think we can remove these lines 8-10 now?

```suggestion
```

MarcUsher: Thank you good catch - switching this to a real PR now for any last fixes and then merging once we're happy to start testing more

.github/workflows/deployment_cd_test_prod.yml
samuelhwilliams, line 1: this looks so good now 😍

PR #664, opened 2024-08-08
Fix migrations if zero schema changes

samuelhwilliams:
File
"/Users/sam/work/dluhc/funding/funding-service-design-post-award-data-store/db/migrations/env.py", line 94, in process_revision_directives
    migration_script = directives[0]
                       ~~~~~~~~~~^^^
IndexError: list index out of range


MarcUsher (approved):
LGTM 👍  interesting that this hasn't cropped up when the migrations have tried to run with no schema changes (maybe it has happened and we've just not noticed). But good catch!

PR #674, opened 2024-08-19
Add checks if migrations are needed

albertkol:
### Change description
The idea is:
- run `flask db current` to get current revision id
- check `.current-alembic-head`
- if they match don't run migration

### How to test
Deploy `test` it should reduce the migration step from 2.5mins to about 10 seconds
Job without new migration: https://github.com/communitiesuk/funding-service-design-post-award-data-store/actions/runs/10491220510/job/29059940501
Job with new migration: https://github.com/communitiesuk/funding-service-design-post-award-data-store/actions/runs/10491335472/job/29060378690

### Screenshots of UI changes (if applicable)
Screenshot from 2024-08-19 13-04-05

### DIFF
--- a/scripts/migration-task-script.py
+++ b/scripts/migration-task-script.py
@@ -5,6 +5,22 @@ import sys
 
 
 def get_db_revision_id() -> str | None:
+    """Get the current revision ID of the database
+
+    We are doing `flask db current 2> /dev/null` and then grepping it,
+    because we need the output of the copilot svc exec command to be short.
+
+    Copilot svc exec uses AWS ssm session-manager, which isn't designed to run
+    in github actions, where it doesn't have a TTY.
+
+    Too much output leads to output buffering, which leads to github actions not
+    getting all of the output. Revision ID is on the last line, so it would
+    not show up.
+
+    By getting the command to generate only a few lines, we ensure that the
+    github action reliably gets the revision ID from copilot.
+    """
+
     try:
         grep_result = subprocess.check_output(
             (
--- a/scripts/migration-task-script.py
+++ b/scripts/migration-task-script.py
@@ -1,36 +1,100 @@
 #!/usr/bin/env python3
+import re
 import subprocess
 import sys
 
-command_to_run = "flask db upgrade"
+
+def get_db_revision_id() -> str | None:
+    try:
+        grep_result = subprocess.check_output(
+            (
+                "copilot svc exec --name post-award "
+                "--command \"bash -c 'launcher flask db current 2> /dev/null | grep head'\""
+            ),
+            shell=True,
+            text=True,
+        )
+    except subprocess.CalledProcessError as e:
+        sys.exit(e.returncode)
+
+    match = re.search(r"[a-zA-Z0-9_]+(?=\s\(head\))", grep_result)
+
+    return match.group() if match else None
+
+
+def init_task() -> str:
+    try:
+        task_command = subprocess.run(
+            args=["copilot", "task", "run", "--generate-cmd", f"pre-award/{environment}/post-award"],
+            capture_output=True,
+            check=True,
+            text=True,
+        )
+    except subprocess.CalledProcessError as e:
+        sys.exit(e.returncode)
+
+    return task_command.stderr
+
+
+def run_migration_in_task(task_command: str, extra_args: list[str]) -> None:
+    # Remove final line break and append arguments
+    try:
+        subprocess.run(
+            args=task_command[:-1] + " \\\n" + " \\\n".join(extra_args),
+            shell=True,
+            check=True,
+        )
+    except subprocess.CalledProcessError as e:
+        # Don't want to leak the command output here so just exit with the command's return code
+        sys.exit(e.returncode)
+
+
+def get_current_alembic_head_file_content() -> str:
+    """Get the content of .current-alembic-head"""
+
+    with open("db/migrations/.current-alembic-head", "r") as file:
+        file_content = file.read().strip()
+
+    return file_content
+
+
+def has_migrations_to_apply() -> bool:
+    """Check if there are migrations to apply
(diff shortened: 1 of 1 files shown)


scripts/migration-task-script.py
```diff
@@ -7,6 +7,32 @@
...
+
+    result = subprocess.getoutput("copilot svc exec --name post-award --command 'grep head current.txt'")
+except subprocess.CalledProcessError as e:
+    sys.exit(e.returncode)
+
+# command returns extra lines besides the grep'd text
+current_revision_id = ""
+for line in result.splitlines():
+    if "(head)" in line:
+        current_revision_id = line.split(" ")[0]
```
gidsg, line 22: Can we do this in a regex and avoid mutating the variable if possible?

scripts/migration-task-script.py
```diff
@@ -7,6 +8,30 @@
 environment = sys.argv[1]
 docker_image = sys.argv[2]
 
+# Check if there are migrations to apply
```
gidsg, line 11: Can we maybe move this to a function instead? (To make the script easier to understand)

scripts/migration-task-script.py
```diff
@@ -7,6 +8,30 @@
...
+try:
+    subprocess.getoutput("copilot svc exec --name post-award --command launcher flask db upgrade > current.txt")
+
+    result = subprocess.getoutput("copilot svc exec --name post-award --command 'grep head current.txt'")
```
samuelhwilliams, line 15: You are running `flask db upgrade` - that is going to upgrade the database, which we don't want here - we just want to check if migrations are _required_?

Also, it's unclear to me if the current approach of doing two `copilot svc exec` calls is reliable. Could it choose two different instances for each command, which would mean the file doesn't exist for the second command?

Also `subprocess.getoutput` appears to be a legacy function. Can we use `check_output` into a variable and scan that?

```
output = subprocess.check_output(["copilot", "svc", "exec", "--name", "post-award", "--command", "launcher 'flask db current'"])
```

scripts/migration-task-script.py
```diff
@@ -7,6 +8,30 @@
...
+except subprocess.CalledProcessError as e:
+    sys.exit(e.returncode)
+
+# command returns extra lines besides the grep'd text
+match = re.search(r"\d+(?=\s\(head\))", result)
```
samuelhwilliams, line 20: This regex doesn't match our standardised alembic version name, which is `###_<some string>`. So the command outputs eg `038_current_fund_data (head)`

scripts/migration-task-script.py
albertkol: Silly mistake.. was meant to be `flask db current`. 

I will look into the instances situation, that one might be trickier. `svc exec` accepts a `task-id`. Will see if providing the same task id to the second command assures that same instance will be hit.

albertkol: I still need to test it a bit more. I am getting a task-id and providing it to both commands.

scripts/migration-task-script.py
```diff
@@ -7,6 +8,30 @@
...
+current_revision_id = match.group() if match else ""
+
+sys.stdout.write(f"Current revision ID: {current_revision_id}\n")
```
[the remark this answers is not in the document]
albertkol: Switched back to print.. early on I would not see print messages show up in the logs. In hindsight, that was for other reasons. It works just fine. `sys.stdout.write()` does look cool though.

scripts/migration-task-script.py
samuelhwilliams: Can you just do it all in one `copilot svc exec` using shell mode?

`copilot svc exec ... | grep ' (head)'`

?

albertkol: This is what I get if I do them together `flask db current | grep head`:
image

Same with `flask db current > current.txt | cat current.txt`:
image
(likely same print with `| grep head current.txt`)

It would indeed be nice if we only had one command. And it's really sad we got to three commands instead of two.

gidsg:
pre-commit.ci autofix

scripts/migration-task-script.py
```diff
@@ -1,36 +1,100 @@
...
+def get_db_revision_id() -> str | None:
+    try:
+        grep_result = subprocess.check_output(
+            (
+                "copilot svc exec --name post-award "
+                "--command \"bash -c 'launcher flask db current 2> /dev/null | grep head'\""
+            ),
+            shell=True,
+            text=True,
+        )
```
samuelhwilliams, line 16: I think a short comment here explaining why we're doing all this malarky wouldn't hurt - i.e. we need the output of the copilot svc exec command to be short because copilot svc exec uses AWS ssm session-manager, which isn't designed to run in github actions (where it doesn't have a TTY). Too much output leads to output buffering, which leads to github actions not getting all of the output (specifically the line of output we need), leading to the script failing.

So we're doing a bit of a song-and-dance to make the command ran inside copilot svc exec only generate a few lines of output, which is short enough for github actions to reliably read, which means we can get the DB revision ID

samuelhwilliams (approved):
🎉 

Good work on this - it's a great time saver when we don't need to run migrations :) Thanks for sticking with it

gidsg (approved):
LGTM 👍

scripts/migration-task-script.py
```diff
@@ -1,36 +1,116 @@
...
-    )
-except subprocess.CalledProcessError as e:
-    # Don't want to leak the command output here so just exit with the command's return code
-    sys.exit(e.returncode)
+
+task_command = init_task()
+
+run_migration_in_task(task_command, extra_args)
+
+print("Migrations to applied")
```
gidsg, line 115: Typo here
should be `Migrations applied` ?
```suggestion
print("Migrations applied")
```

PR #675, opened 2024-08-19
Upgrade flask and flask-wtf

albertkol:
### Change description
Upgrade app to flask 3.0.* and flask-wtf to 1.2.1.

### How to test
Go to `submit/routes.py:login()` add `print(nonexistent_variable)`.

With current Flask version, you get the trace error page.
With new version, you get the HTML error page, unless you have `FLASK_DEBUG=1` in your ENV.


samuelhwilliams (commented):
Can you also look into FLASK_ENV. We have a lot of references to this and it's supposedly removed in Flask 3.0. This may need changes in fsd-utils so that it's compatible with flask 2.2/2.3/3.0. Flask team have done some really weird stuff around this env var so if you could make it clear what we need to do to maintain existing behaviour that would be super great 🙏

albertkol:
A bit of a brain dump looking into `FLASK_ENV`. 

We will still need some kind of variable name to store the environment string: `test`, `dev` or `prod`.. for situations like this:
```
match FLASK_ENV:
    case "unit_test":
        from config.envs.unit_test import UnitTestConfig as Config
    case "development":
        from config.envs.development import DevelopmentConfig as Config  # type: ignore # TODO: fixme
    case "dev" | "test" | "prod":
        from config.envs.aws import AwsConfig as Config  # type: ignore # TODO: fixme
    case _:
        from config.envs.default import DefaultConfig as Config  # type: ignore # TODO: fixme
```

We can rename it to `APP_ENV`. But not sure how productive that is.

What newer version of `flask` only broke the connection between `FLASK_ENV=='development'` enabling debug mode. Now we would have to set `FLASK_DEBUG1 to 1 or 0. 

I still have to look a bit more into `fsd-utils` and `account-store`. 
`account-store` is already on latest version of `flask` with an `fsd-utils` dependency. 
At first glance I don't see `FLASK_DEBUG` set at all on `account-store` and not mentioned enough on `fsd-utils`. 
So I wonder if debug mode even works on `account-store` currently. :thinking: 
TBD.

albertkol:
Conclusion on `FLASK_ENV`:
- `FLASK_ENV=development` used to enable debugging, newer Flask versions no longer user `FLASK_ENV`, so we have to manually set `FLASK_DEBUG=1` 
- We could rename `FLASK_ENV`, but I think it would cause more harm than good, as for example: [funding-service-design-utils](https://github.com/search?q=repo%3Acommunitiesuk%2Ffunding-service-design-utils%20FLASK_ENV&type=code) still uses it. We use it too for conditioning which config we're using.
- We don't use debug mode on any other environment than `development`, which is our `local`. So we can get away with a line in our `.flaskenv`

Happy to discuss these..

samuelhwilliams:
Should we do an update/fix on fsd-utils `CommonConfig` that sets `FLASK_DEBUG` when `FLASK_ENV` is `development`? Is that a thing we should have done when adding Flask 3 support to fsd-utils?

Then we could probably pull that in and it would "just work" for all FS apps?

albertkol:
I tried it, kind of... I defined it in our `DefaultConfig` :
```python
class DefaultConfig(object):
    FLASK_ROOT = Path(__file__).parent.parent.parent
    FLASK_ENV = CommonConfig.FLASK_ENV
    DEBUG = True
```

It would not enable debug more, reading the [docs](https://flask.palletsprojects.com/en/3.0.x/config/#debug-mode):
_While it is possible to set [DEBUG](https://flask.palletsprojects.com/en/3.0.x/config/#DEBUG) in your config or code, this is strongly discouraged. It can’t be read early by the flask run command, and some systems or extensions may have already configured themselves based on a previous value._

Is what I suspect is happening..

samuelhwilliams:
Goodness 🙃 Ok. So we shouldn't have a FLASK_DEBUG/DEBUG config variable and should instead rely on passing in the `--debug` flag to `flask run` commands? So we would need to update our docker-compose file basically?

albertkol:
`FLASK_DEBUG` env does the job. I did not add it in the `docker-compose` as `.flaskenv` is enough. (we can add it, if needed)

I see that we have `FLASK_ENV` on all fronts: docker-runner, .flaskenv and CommonConfig. I have not tested, but it's likely that `CommonConfig.FLASK_ENV` is not the one enabling debug for us atm, but rather the ENV variable.

albertkol:
Added `How to test` info.

samuelhwilliams (approved):

PR #678, opened 2024-08-22
Migrate DB before code deployments

samuelhwilliams:
Our current deployment pipeline runs three jobs in parallel:

1. Migrate DB and deploy post-award service
2. Deploy post-award-celery service
3. Deploy download-report job

If the DB migration fails, we still deploy post-award-celery and the
download-report job. These might be relying on model changes from a DB
migration, so it is logically incorrect to be deploying these if the
migration fails. It also leaves us in an inconsistent place - post-award
service is on the old code, post-award-celery/download-report are on the
new code.

We should run the DB migrations before deploying any of the code
changes. If these succeed, we can run all three deployments. If they
fail, we shouldn't run any.

### DIFF
--- a/.github/workflows/aws_copilot_deployment.yml
+++ b/.github/workflows/aws_copilot_deployment.yml
@@ -13,8 +13,34 @@ on:
             required: true
 
 jobs:
+  migrate:
+    name: Migrate DB in ${{ inputs.copilot_environment || 'test' }}
+    permissions:
+      id-token: write # This is required for requesting the JWT
+      contents: read  # This is required for actions/checkout
+    runs-on: ubuntu-latest
+    environment: ${{ inputs.copilot_environment || 'test' }}
+    steps:
+      - name: Git clone the repository
+        uses: actions/checkout@v4
+
+      - name: Setup Copilot
+        uses: communitiesuk/funding-service-design-workflows/.github/actions/copilot_setup@main
+        with:
+          copilot_environment: ${{ inputs.copilot_environment || 'test' }}
+          AWS_ACCOUNT: ${{ secrets.AWS_ACCOUNT }}
+
+      - name: Inject env specific values into manifest
+        run: |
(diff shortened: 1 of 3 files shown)


albertkol:
It's missing the change for the `test`/`prod` deploy pipeline, no?

samuelhwilliams:
Er yes - it totally is. Good catch.

.github/workflows/deploy_combined_service.yml
```diff
@@ -12,6 +12,7 @@ on:
 
 jobs:
   paketo_build:
+    name: Build post-award image
```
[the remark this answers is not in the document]
albertkol: Added it quickly.. wanted to try and make a commit from the web vscode.

.github/workflows/aws_copilot_deployment.yml
```diff
@@ -13,8 +13,34 @@ on:
...
+      - name: Setup Copilot
+        uses: communitiesuk/funding-service-design-workflows/.github/actions/copilot_setup@main
+        with:
+          copilot_environment: ${{ inputs.copilot_environment || 'test' }}
+          AWS_ACCOUNT: ${{ secrets.AWS_ACCOUNT }}
+
+      - name: Inject env specific values into manifest
+        run: |
+          yq -i '.variables.GITHUB_SHA = "${{ github.sha }}"'  copilot/post-award/manifest.yml
+          yq -i '.image.location = "${{ inputs.image_location }}"' copilot/post-award/manifest.yml
```
albertkol, line 36: I don't think this is necessary for the `migrate` job.

albertkol (changes requested):

.github/workflows/aws_copilot_deployment.yml
albertkol: Hmm.. might be wrong.. :thinking:

albertkol: You need the manifest when running migrations. 

I am monologuing in your PR comment section. XD

albertkol (approved):

PR #680, opened 2024-08-28
[Draft] Fix file too big bug

albertkol:

### Change description

### How to test


find/main/routes.py
```diff
@@ -175,6 +178,49 @@ def retrieve_download(filename: str):
...
+                url_for(
+                    ".retrieve_spreadsheet",
+                    fund_code=fund_code,
+                    submission_id=submission_id,
+                )
+            )
+
+        return render_template("find/main/spreadsheet-not-found.html")
+
+    filename = f"{fund_code}---{submission_id}.xlsx"
```
samuelhwilliams, line 201: suggestion: Should/could we also include the LA name to potentially be more user-friendly? Wonder if the date is useful as well? Maybe `{date} - {fund_code} - {submission_id} - {org_name}` - ie `2024-08-29 - TF - S-R05-1 - Bolton Council`?

nit: not sure we need three dashes in the filename, that feels a bit unusual to me.

find/main/routes.py
```diff
@@ -175,6 +178,49 @@ def retrieve_download(filename: str):
...
+    file_size = round(file_metadata["ContentLength"] / (1024 * 1024), 1)
```
samuelhwilliams, line 202: We have a function called `get_human_readable_file_size` in `async_download` - we can probably move this to some shared/central place and use that instead?

samuelhwilliams: There's also `get_find_download_file_metadata` which does some of what you've added with the `return_full_head_data` - can we come up with something consistent and standardise on that?

data_store/aws.py
```diff
@@ -86,11 +86,17 @@ def get_failed_file_key(failure_uuid: UUID) -> str:
     return file_key
 
 
-def get_file_metadata(bucket_name: str, file_key: str) -> Dict[str, str]:
+def get_file_metadata(
+    bucket_name: str,
+    file_key: str,
+    return_full_head_data: bool = False,
```
samuelhwilliams, line 92: I don't really love parameters that just toggle the return type - that feels more like you have two separate functions and makes it a bit harder to reason about.

This function is specifically about getting the S3 object Metadata tags - not arbitrary metadata. I think this crosses over a bit with what `get_find_download_file_metadata` does - can we refactor these to get one consistent sensible thing? Something like `get_s3_object_attributes` (finger-in-the-air name)?

This change also breaks the typing on this function's return signature (which I've now gone down a bit of a rabbit-hole on - looks like we need to be installing `boto3-stubs[essential]`, not `types-boto3` 🤦)

find/main/routes.py
```diff
@@ -175,6 +178,49 @@ def retrieve_download(filename: str):
...
+        if request.method == "POST":
+            return redirect(
+                url_for(
+                    ".retrieve_spreadsheet",
+                    fund_code=fund_code,
+                    submission_id=submission_id,
+                )
+            )
+
+        return render_template("find/main/spreadsheet-not-found.html")
```
samuelhwilliams, line 199: I think we can probably just render a normal 404 page - do we need a custom page for this? The spreadsheets don't have a short TTL on them, unlike the download reports, so this feels less likely to happen.

find/main/routes.py
```diff
@@ -175,6 +178,49 @@ def retrieve_download(filename: str):
...
 
 
+@bp.route("/retrieve-spreadsheet/<fund_code>/<submission_id>", methods=["GET", "POST"])
+@login_required(return_app=SupportedApp.POST_AWARD_FRONTEND)
+@check_internal_user
+def retrieve_spreadsheet(fund_code: str, submission_id: str):
+    object_name = f"{fund_code}/{submission_id}"
+
+    try:
+        file_metadata = get_file_metadata(Config.AWS_S3_BUCKET_SUCCESSFUL_FILES, object_name, True)
```
samuelhwilliams, line 188: Sort of related to the comment about this function signature - but this floating `True` is hard to read because at a glance I have no idea what it means. If we keep this (which maybe I hope we don't), it'd probably be good to force it to be a [keyword-only argument](https://peps.python.org/pep-3102/#specification).

submit/main/notify.py
```diff
@@ -59,33 +63,57 @@ def prepare_upload(file: FileStorage):
...
+
+    programme = get_programme_by_id_and_round(programme_id, round_number)
+    _, submission_id = get_or_generate_submission_id(programme, round_number, fund.fund_code)
+
+    object_name = f"{fund.fund_code}/{submission_id}"
+    personalisation.update(
+        {
+            "download_url": f"{Config.FIND_SERVICE_BASE_URL}/retrieve-spreadsheet/{object_name}",
+        }
+    )
```
samuelhwilliams, line 104: If all of this is just to generate the bit of personalisation, should we move it to `get_personalisation`? Also not sure I love this needing to make DB calls from here - can we just pass in the submission_id to `send_confirmation_emails` if that's what we need here?

I particularly think it's awkward to be calling `get_or_generate_submission_id` - that feels very much like a thing that should remain internal to `data_store`. Especially because it can just 'imagine' submission IDs (ie generate one if it can't find a match - but it wouldn't persist it anywhere)

submit/main/notify.py
```diff
@@ -59,33 +63,57 @@ def prepare_upload(file: FileStorage):
...
+
+    # email to Fund
+    send_email(
+        email_address=fund.email,
         template_id=Config.FUND_CONFIRMATION_EMAIL_TEMPLATE_ID,
-        file=excel_file,
+        notify_key=Config.NOTIFY_FIND_API_KEY,
```
samuelhwilliams, line 110: This doesn't feel right for a couple of reasons:

* Code in the `submit` module is now using the Find API key.
* These emails have always come via the Submit service before - why do we want to send them from the Find service now? It looks like you've made the new template in the Find service too - how come?

find/main/routes.py
samuelhwilliams: P.S. I'm happy for refactors to be a separate PR if it helps keep things small/contained/reviewable 🙏

data_store/aws.py
albertkol: I changed this to be.. `get_file_header`. 

You'll let me know what you think of that on the new version. It returns the whole file header info. If people want metadata they will do `file_header["Metadata"]`, accepts a flag `formatted:bool`, if they want the human readable formatting.

find/main/routes.py
albertkol: :+1:

find/main/routes.py
albertkol: :+1:

submit/main/notify.py
albertkol: I got rid of the `get_personalisation` function. You'll let me know what you thing of the new version.

submit/main/notify.py
albertkol: `notify.py` lives in `data_store.controllers`

data_store/aws.py
```diff
@@ -86,32 +87,36 @@ def get_failed_file_key(failure_uuid: UUID) -> str:
...
-    if return_full_head_data:
-        return s3_response
+    if formatted:
+        s3_response.update(
+            {
+                "ContentLength": get_human_readable_file_size(s3_response["ContentLength"]),
+                "ContentType": get_file_format_from_content_type(s3_response["ContentType"]),
+                "LastModified": s3_response["LastModified"].strftime("%d %B %Y"),
+            }
+        )
```
samuelhwilliams, line 117: I don't see anywhere that we use the unformatted response - do we just want to always format it and remove the `formatted` param from the signature?

data_store/controllers/notify.py
```diff
@@ -0,0 +1,108 @@
...
+
+    try:
+        notify_client = NotificationsAPIClient(notify_key)
+
+        notify_client.send_email_notification(
+            email_address=email_address,
+            template_id=template_id,
+            personalisation=personalisation,
+        )
+    except HTTPError as error:
```
albertkol, line 37: I am catching HTTPError.. despite not doing anything with it. Had second thoughts if it's worth it, but then I thought it might be nice to see what errors one should expect.

data_store/aws.py
samuelhwilliams: nit: I also find the `.update()` syntax a bit weird - curious why you prefer that over just doing assignments?

```python
s3_response['ContentLength'] = get_human_readable_file_size(s3_response["ContentLength"])
...
```

data_store/controllers/notify.py
```diff
@@ -0,0 +1,108 @@
...
+    excel_file: FileStorage,
+    fund: FundConfig,
+    user_email: str,
+    metadata: dict,
+):
+    place_name = metadata.get("Programme Name")
+    fund_type = metadata.get("FundType_ID")
+
+    if not (place_name or fund_type):
+        raise ValueError(
```
albertkol, line 51: Raising ValueError rather that logging directly inside the function.

data_store/controllers/notify.py
```diff
@@ -0,0 +1,108 @@
...
+from werkzeug.datastructures import FileStorage
+
+from config import Config
+from data_store.controllers.load_functions import get_or_generate_submission_id
+from data_store.db.queries import get_programme_by_id_and_round
+from submit.main.fund import FundConfig
+from submit.utils import get_friendly_fund_type
+
+
+def send_email(
```
albertkol, line 14: Deleted the part that accepted attachments. Maybe controversial. We don't use it anymore. Can be added again if needed.

data_store/aws.py
albertkol: I think I used `update` recently in another place, so now we are bffs. 
Happy to change it.

As for the `formatted` parameter. Happy to drop it too. My first instinct when I refactor is to try to maintain the former behavior.

data_store/controllers/notify.py
albertkol: I don't know if it's even possible for this `Exception` to happen and I was very tempted to delete it. But I reluctantly kept it.

find/main/routes.py
albertkol: Returning the common `404.html` page now.

find/main/routes.py
```diff
@@ -190,7 +196,7 @@ def retrieve_spreadsheet(fund_code: str, submission_id: str):
                 )
             )
 
-        return render_template("find/main/spreadsheet-not-found.html")
+        return render_template("common/templates/common/404.html")
```
samuelhwilliams, line 199: This will return a 200 status code - the Flask-y way to do this would be to just call `abort(404)` - which is tied up to render the correct error page.

data_store/controllers/notify.py
samuelhwilliams: I don't love this. HTTPError is one of many that could be thrown - don't think it makes much sense to call it out specifically. Error management in python is just one of those not-very-nice things.

data_store/controllers/notify.py
samuelhwilliams: Rather than just passing in `metadata` which is a random dict, why not have this take the place_name and fund_type explicitly as strings? Then we don't have to worry about this validation.

data_store/controllers/notify.py
samuelhwilliams: Makes sense to me.

The type hints of `notify_key: str | None` feels wrong - it's a required value, so should just be `str` now

albertkol: `mypy` shamed me for it.
 
Because the NOTIFY_KEY envs can be `None` as we currently fetch them from config. I can change that..

data_store/controllers/notify.py
albertkol: Should I remove the try-except then?

data_store/controllers/notify.py
```diff
@@ -0,0 +1,108 @@
...
+
+from notifications_python_client.errors import HTTPError
+from notifications_python_client.notifications import NotificationsAPIClient
+from werkzeug.datastructures import FileStorage
+
+from config import Config
+from data_store.controllers.load_functions import get_or_generate_submission_id
+from data_store.db.queries import get_programme_by_id_and_round
+from submit.main.fund import FundConfig
+from submit.utils import get_friendly_fund_type
```
albertkol, line 11: Thoughts on breaking this submit import? 
`get_friendly_fund_type` is not too hard to move.. perhaps.
`FundConfig` not too sure what it is, but I don't think it's a Submit only thing.

submit/main/notify.py
albertkol: Resolving this.

find/main/routes.py
albertkol: Changed.

submit/main/notify.py
albertkol: Good point on the template.. I created it on Find. But retrieve-download is a Fine related feature. 

Perhaps template should be on Submit? :thinking:

data_store/controllers/notify.py
samuelhwilliams: yes please :)

find/main/routes.py
albertkol: Changing it to:
`{date}-{fund-code}-{submission.id}` 
This is easy because I have this data already.

To get `submission.submission_id` and `org_name`. I'd have to go to the db which is a bit ugh..

samuelhwilliams: The org name - to me - feels a very useful thing for the fund team to have in the file name. I think we should include it?

data_store/controllers/notify.py
samuelhwilliams: Yes, data store shouldn't import from submit.

data_store/controllers/notify.py
albertkol: Changed the message.

submit/main/notify.py
samuelhwilliams: Doesn't overlay matter I guess... Current template is on submit so let's keep it there.

find/main/routes.py
Kateharries: 2024-07-08-pf-bolton-council-jan2024-mar2024.xlsx

This seems to have all the info they would need. Ideally, we'd want to find out from users what would be the most useful file naming convention for them, but since we're not doing any research on Submit MVP at the moment, I guess this isn't really possible.

data_store/controllers/notify.py
samuelhwilliams: Ah yeah... this sort of ties into https://dluhcdigital.atlassian.net/browse/FPASF-163 I guess. It being optional config is annoying...

submit/main/notify.py
albertkol: Moved it to submit.

samuelhwilliams: Is there any reason we need a new template at all...?

Also just looked at the template - it says the link will stop working after 7 days. That's not true? I think we could have just reused the existing email, not sure why we need a new template.

albertkol: Find email claims the link expires in 7 days. Are the two different?

samuelhwilliams: Yes. The find bucket has a 7 day lifecycle policy on it. The submitted spreadsheets bucket doesn't, I believe - maybe check.

samuelhwilliams: Ooo gross, you're right. I'll get on and fix that 7 days thing for find downloads 👀

samuelhwilliams: Fixed that 👍

albertkol: I managed to switch to the old template.. I thought that template would be missing some parts we need. But it does include a download link.

@Kateharries is it ok if we stick to it?
image

data_store/controllers/notify.py
```diff
@@ -0,0 +1,85 @@
...
+    filename: str,
+    user_email: str,
+    fund_name: str,
+    current_reporting_period: str,
+    programme_name: str,
+    fund_type: str,
+):
+    if not (programme_name or fund_type):
+        raise ValueError("Cannot personalise confirmation email without programme name or fund type")
+
```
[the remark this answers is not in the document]
albertkol: Removed.

data_store/controllers/notify.py
```diff
@@ -0,0 +1,85 @@
...
+def send_fund_confirmation_email(
+    fund_name: str,
+    fund_email: str,
+    round_number: int,
+    programme_name: str,
+    fund_type: str,
+    programme_id: str,
+):
+    if not (programme_name or fund_type):
+        raise ValueError("Cannot personalise confirmation email without programme name or fund type")
```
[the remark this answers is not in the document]
albertkol: Removed.

data_store/aws.py
```diff
@@ -86,23 +87,30 @@ def get_failed_file_key(failure_uuid: UUID) -> str:
...
+            raise FileNotFoundError(f"Could not find file {file_key} in S3.") from error
+
         raise error
 
-    return s3_response["Metadata"]
+    file_header = {
+        "ContentLength": get_human_readable_file_size(s3_response["ContentLength"]),
+        "ContentType": get_file_format_from_content_type(s3_response["ContentType"]),
+        "LastModified": s3_response["LastModified"],
+        "Metadata": s3_response["Metadata"],
```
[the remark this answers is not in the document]
albertkol: Changed.

data_store/aws.py
```diff
@@ -131,3 +139,32 @@ def create_presigned_url(bucket_name: str, file_key: str, filename: str, expirat
...
+    :return: human-readable file size,
+    """
+
+    file_size_kb = round(file_size_bytes / 1024, 1)
+    if file_size_kb < 1024:
+        return f"{round(file_size_kb, 1)} KB"
+    elif file_size_kb < 1024 * 1024:
+        return f"{round(file_size_kb / 1024, 1)} MB"
+    else:
+        return f"{round(file_size_kb / (1024 * 1024), 1)} GB"
```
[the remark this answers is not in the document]
albertkol: Moved.

data_store/controllers/notify.py
```diff
@@ -0,0 +1,85 @@
...
+
+    programme = get_programme_by_id_and_round(programme_id, round_number)
+    _, submission_id = get_or_generate_submission_id(programme, round_number, fund_type)
```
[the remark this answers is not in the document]
albertkol: Ok. I changed this quite a bit:
1. No longer using `get_or_generate_submission_id` instead I extracted the logic I needed in a separate function called `get_submission_by_programme_and_round` and used that instead.
2. Raising an error if `programme` is None

data_store/controllers/notify.py
```diff
@@ -0,0 +1,85 @@
...
+
+    if submission_id is None:
+        raise ValueError("Submission ID not found")
+
+    object_name = f"{fund_type}/{submission_id}"
+    link_to_file = f"{Config.FIND_SERVICE_BASE_URL}/retrieve-spreadsheet/{object_name}"
```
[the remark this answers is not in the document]
albertkol: This is the only I am not too sure about..

Using the `url_for` function would create an Flask `app.context` dependency inside `notify.py`. Which I think we should try to avoid.

Passing the URL as a parameter to the function is also not as simple because I don't have `submission.id` outside of my `notify.py` function.

Happy to discuss it tomorrow.

I see we use `app.logger` and `g` inside `data-store` directory (f.x. `ingest.py`). So maybe it's not a big deal. In my mind app context should not leave `find` and `submit`.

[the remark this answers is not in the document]
samuelhwilliams: Hmm you make a good point...let's chat indeed

data_store/controllers/load_functions.py
```diff
@@ -188,6 +188,19 @@ def delete_existing_submission(submission_to_del: str) -> None:
     db.session.flush()
 
 
+def get_submission_by_programme_and_round(
+    programme: Programme,
```
samuelhwilliams, line 192: Tiny thing - but I guess this feels more reusable if it takes the programme_id instead of the full instance?

data_store/controllers/notify.py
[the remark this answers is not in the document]
albertkol: Using `url_for` now.

samuelhwilliams (approved):

PR #681, opened 2024-08-29
Fix type hints for boto3

samuelhwilliams:
### Change description
We weren't installing many of the actual type hints for boto3, leading to some AWS types falling back to `Any` (ie no type checking). This pulls in the essential packages and then fixes some of the code that is incorrect.

### DIFF
--- a/data_store/aws.py
+++ b/data_store/aws.py
@@ -49,7 +49,7 @@ def upload_file(file: Union[IO, FileStorage], bucket: str, object_name: str, met
 
     file.seek(0)
     _S3_CLIENT.upload_fileobj(
-        file,
+        file.stream if isinstance(file, FileStorage) else file,
         bucket,
         object_name,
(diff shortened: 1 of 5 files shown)


data_store/aws.py
```diff
@@ -49,7 +49,7 @@ def upload_file(file: Union[IO, FileStorage], bucket: str, object_name: str, met
 
     file.seek(0)
     _S3_CLIENT.upload_fileobj(
-        file,
+        file.stream if isinstance(file, FileStorage) else file,
```
albertkol, line 52: Hmm.. not really related to the PR.
But I think `file: Union[IO, FileStorage]` could have easily been just `FileStorage`.

albertkol (approved):
:+1:

data_store/aws.py
samuelhwilliams: Definitely it should either just be IO or FileStorage, not both, agree 👍

PR #695, opened 2024-09-11
Logging for download

sc15zs:
Ticket:
https://dluhcdigital.atlassian.net/browse/FPASF-537

Description:
The users' email addresses should not be logged for download, to avoid logging of PIIs.
To have the email addresses included in the file generated by the report script, a call for each user_id is done to account-store to get that.

Unit test updated, to check that email_address is no longer logged.

### DIFF
--- a/find/scripts/extract_download_logs.py
+++ b/find/scripts/extract_download_logs.py
@@ -15,10 +15,13 @@ import time
 from io import StringIO
 from typing import List
 
+import requests
 from boto3 import client
 from dateutil.relativedelta import relativedelta
 from notifications_python_client import prepare_upload
 from notifications_python_client.notifications import NotificationsAPIClient
+
+from config import Config
 
 # Default FLASK_ENV here to allow import when running locally
 if not os.getenv("FLASK_ENV"):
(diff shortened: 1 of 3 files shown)


tests/find_tests/test_logging.py
```diff
@@ -4,4 +4,5 @@ def test_download_logging(find_test_client, caplog, mocked_routes_trigger_async_
     assert len(log_line) == 1
     assert log_line[0].request_type == "download"
     assert log_line[0].user_id == "test-user"
-    assert log_line[0].query_params == {"file_format": "xlsx", "email_address": "test-user@communities.gov.uk"}
+    assert log_line[0].query_params == {"file_format": "xlsx"}
+    assert "email_address" not in log_line[0].query_params
```
samuelhwilliams, line 8: question: These checks are redundant, no?

sc15zs: Yes, kind of redundant, I'm just hesitant to delete tests. 
All its use at the moment is to check that in find/main/routes/download the email_address has indeed been removed from query_params:
`query_params_without_email_address = {k: v for k, v in query_params.items() if k != "email_address"}
`
Not much of a test. Shall I delete it?

samuelhwilliams: I suppose I mean that - we only need one of these two lines. Each line is testing the same thing - that `email_address` is not present.

sc15zs: Sorry, I misunderstood you. I removed one of the lines.

find/scripts/extract_download_logs.py
```diff
@@ -53,11 +56,54 @@ def send_notify(
...
+    for user_id in user_ids:
+        try:
+            response = requests.get(
+                f"{Config.ACCOUNT_STORE_API_HOST}/accounts",
+                headers={"Content-Type": "application/json"},
+                params={"azure_ad_subject_id": user_id},
+            )
+            response.raise_for_status()
+            response_data = response.json()
+            responses[user_id] = response_data.get("email_address")
```
samuelhwilliams, line 91: is `email_address` an optional field here? If it's not, should we do a hard lookup? If it is, should we add a default value if it's not present, like `no-email-address`?

sc15zs: Yes, you're right, it's just that I most of the time use get instead of [], even if there's actually no chance of getting a key error. 
Email is not nullable, so I've changed it to a hard lookup.

find/scripts/extract_download_logs.py
```diff
@@ -53,11 +56,54 @@ def send_notify(
...
+
 def cloudwatch_logs_to_rows(data: List[List[dict]]) -> List[dict]:
+    account_ids = extract_ids_from_message(data)
+    user_id_mapping = fetch_user_data(account_ids)
+
     def parse_item(item: List[dict]) -> dict:
         message = json.loads([i for i in item if i["field"] == "@message"][0]["value"])
         user_id = message["user_id"]
-        email = message.get("email")
+        email = user_id_mapping.get(user_id, None)
```
samuelhwilliams, line 106: suggestion: this could also be a hard lookup

samuelhwilliams (approved):

find/scripts/extract_download_logs.py
sc15zs: Changed to hard lookup.

samuelhwilliams:
Please squash commits before merging :)

PR #699, opened 2024-09-13
Update e2e tests to work on dev and test envs

samuelhwilliams:
https://dluhcdigital.atlassian.net/browse/FPASF-471

### Change description
This PR fixes the Find test to work in an arbitrary environment (rather than expecting specific data to be seeded into the database for Find filters).

And then refactors the end-to-end test code so that it can point at either the `local` developer environment, or the deployed `dev`/`test` environments. It also can now run _from_ a local developer environment or in GitHub actions, so that we can integrate it into CI later.

Because we need to do some branching logic depending on the source and target environment, I've extracted the end-to-end config from the `UnitTest` config object into a separate file for end-to-end test secrets. This lets us use either aws-vault (locally) or expect implicit AWS credentials (github OIDC).

### How to test


tests/e2e_tests/conftest.py
```diff
@@ -65,50 +70,109 @@ def authenticator_fund_config(request: FixtureRequest) -> TestFundConfig:
...
+                "secure": True,
+                "sameSize": "Strict",
+            }
+        ]
     )
 
-    response = requests.get(f"{domains.authenticator}/magic-links")
-    magic_links_before = set(response.json())
-
-    new_magic_link_page = NewMagicLinkPage(page, domain=domains.authenticator, fund_config=authenticator_fund_config)
```
albertkol, line 90: You can probably remove the authentication.py MagicLink classes.

tests/e2e_tests/conftest.py
```diff
@@ -65,50 +70,109 @@ def authenticator_fund_config(request: FixtureRequest) -> TestFundConfig:
         raise ValueError(f"not configured for {e2e_env}")
 
 
+def _read_jwt_signing_key(e2e_env: str, e2e_aws_vault_profile: str | None):
```
albertkol, line 73: :exploding_head:

tests/e2e_tests/conftest.py
samuelhwilliams: good shout

.github/workflows/test_e2e.yml
```diff
@@ -0,0 +1,46 @@
...
+        default: dev
+  workflow_dispatch:
+    inputs:
+      copilot_environment:
+        description: "Copilot environment to deploy to"
+        type: choice
+        options:
+          - dev
+          - test
+        default: dev
```
samuelhwilliams, line 17: temp for testing

.github/workflows/test_e2e.yml
```diff
@@ -0,0 +1,46 @@
...
+    - name: Setup AWS credentials
+      uses: communitiesuk/funding-service-design-workflows/.github/actions/copilot_setup@main
+      with:
+        copilot_environment: ${{ inputs.copilot_environment }}
+        AWS_ACCOUNT: ${{ secrets.AWS_ACCOUNT }}
+    - name: Run tests
+      run: pytest tests tests/e2e_tests --e2e --e2e-env ${{ inputs.copilot_environment }}
+      env:
+        E2E_DEVTEST_BASIC_AUTH_USERNAME: test
+        E2E_DEVTEST_BASIC_AUTH_PASSWORD: test
```
samuelhwilliams, line 46: temp for testing

tests/e2e_tests/config.py
```diff
@@ -0,0 +1,89 @@
+import base64
+import json
+import os
+import subprocess
+from pathlib import Path
+
+import boto3
+
+
+class EndToEndTestSecrets:
```
albertkol, line 10: nit: Messy class. It's probably easiest to write this way. But all these `if`'s in case it's local, could have been a class on its own.

tests/e2e_tests/conftest.py
```diff
@@ -65,50 +66,63 @@ def authenticator_fund_config(request: FixtureRequest) -> TestFundConfig:
         raise ValueError(f"not configured for {e2e_env}")
 
 
+@pytest.fixture
+def context(request: FixtureRequest, browser: Browser, e2e_test_secrets: EndToEndTestSecrets):
+    e2e_env = request.config.getoption("e2e_env")
+    return browser.new_context(
+        **(dict(http_credentials=e2e_test_secrets.AWS_BASIC_AUTH) if e2e_env in {"dev", "test"} else {})
```
albertkol, line 73: nit: I would have split this in multiple lines for readability

tests/e2e_tests/conftest.py
```diff
@@ -65,50 +66,63 @@ def authenticator_fund_config(request: FixtureRequest) -> TestFundConfig:
...
+        "email": email_address,
+        "fullName": f"E2E Test User - {request.node.originalname}",
+        "roles": user_roles,
+        "iat": now,
+        "exp": now + (15 * 60),  # 15 minutes from now
+    }
+
+    def create_token(payload):
+        algorithm = "RS256"  # Must match the algorithm used by fsd-authenticator
+        return jwt.encode(payload, e2e_test_secrets.JWT_SIGNING_KEY, algorithm=algorithm)
```
albertkol, line 113: nit: Do we need this function inside the function?

albertkol (approved):
3 nit comments. The thing works. :+1: 

I tested it locally with `dev` and `test`. And dispatched a `dev` and `test` job, which passed.

Awesome stuff!

tests/e2e_tests/config.py
samuelhwilliams: Fair call out.

See 2ade59187ef4236d13eeb2c74e50c7f2efef94a3 for updates - turned this into a protocol now 👀 Mild over-engineering, but a proof of concept for how we can do things without inheritance utilising mypy/type hinting.

tests/e2e_tests/config.py
```diff
@@ -0,0 +1,111 @@
...
+                    ],
+                ).decode()
+            )["Parameter"]["Value"]
+
+        # This flow is used when running tests *in* CI/CD, where AWS credentials are available from OIDC auth
+        else:
+            ssm_client = boto3.client("ssm")
+            value = ssm_client.get_parameter(Name=parameter, WithDecryption=True)["Parameter"]["Value"]
+
+        return value
```
albertkol, line 81: I'd avoid the `else` and quick return `value`.

tests/e2e_tests/config.py
```diff
@@ -0,0 +1,111 @@
...
+        return os.environ["E2E_NOTIFY_SUBMIT_API_KEY"]
+
+
+class AWSEndToEndSecrets:
+    def __init__(self, e2e_env: Literal["dev", "test"], e2e_aws_vault_profile: str | None):
+        self.e2e_env = e2e_env
+        self.e2e_aws_vault_profile = e2e_aws_vault_profile
+
+        if self.e2e_env == "prod":  # type: ignore[comparison-overlap]
+            raise ValueError("shouldn't be possible, but also must never happen")
```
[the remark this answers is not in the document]
samuelhwilliams: I'm doing this intentionally to help prevent bad things happening.

I specifically do not want someone to look at the `Literal['dev', 'test']` and think "oh hey I can just add prod and away we go". I do not want that to happen - because that would mean reading prod secrets, and no - I want to add layers of "stop and think what you're doing" before anyone does that.

tests/e2e_tests/config.py
```diff
@@ -0,0 +1,111 @@
...
+import os
+import subprocess
+from pathlib import Path
+from typing import Literal, Protocol
+
+import boto3
+from playwright.sync_api import HttpCredentials
+
+
+class EndToEndTestSecrets(Protocol):
```
albertkol, line 12: cool

PR #701, opened 2024-09-16
Move from script hashes to CSP nonces

samuelhwilliams:
https://dluhcdigital.atlassian.net/browse/FPASF-433

### Change description
Previously, we allowed custom JS to execute in the browser by whitelisting the hashes of the script contents. This had a few issues

* We reference some JS by domain names, so needed to have the hash for each of our environments whitelisted.
* We didn't comment these well, so it's unclear which hashes are whitelisting which scripts.

We now stop using the hash whitelist and move over to CSP nonces. These are random values that we associate with each custom inline script to tell the browser they're trusted and can be executed.

The CSP nonce present on each script tag must match the CSP nonce passed back in the HTTP response header, otherwise the browser will refuse to run the script.

## How to test

- Run the tests. Maybe even the nice new end-to-end tests ;)
- Check browser console log - shouldn't see any errors about scripts/styles not being executed/loaded.

### DIFF
--- a/app.py
+++ b/app.py
@@ -6,7 +6,7 @@ from celery import Celery
 from celery.signals import setup_logging
 from flask import Flask, current_app, flash, redirect, render_template, request
 from flask_assets import Environment
-from flask_talisman import Talisman
+from flask_talisman import DEFAULT_CSP_POLICY, Talisman
 from flask_wtf.csrf import CSRFError, CSRFProtect
 from fsd_utils import init_sentry
 from fsd_utils.healthchecks.checkers import FlaskRunningChecker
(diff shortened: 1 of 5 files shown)


app.py
```diff
@@ -123,19 +123,15 @@ def pop_host_from_current_request(endpoint, values):
...
-    # TODO: Work out which scripts these SHAs relate to and comment this better
-    csp = {
-        "default-src": "'self'",
-        "script-src": [
-            "'self'",
-            "'sha256-+6WnXIl4mbFTCARd8N3COQmT3bJJmo32N8q8ZSQAIcU='",
-            "'sha256-l1eTVSK8DTnK8+yloud7wZUqFrI0atVo6VlC6PJvYaQ='",
-            "'sha256-GUQ5ad8JK5KmEWmROf3LZd9ge94daqNvd8xy9YS1iDw='",
-            "'sha256-ndKdvEBfUn27+cpVrq2H697Dg88x3tsepe3veTUtsaA='",
-            "'sha256-oR+vHsLl1DMudaF9Ay6TIDK2lXwFka0z5sbdN5RZQxE='",
```
albertkol, line 135: Any idea where these scripts are? Are they in templates we import? Did we just forget to delete them?

Only script I see we are directly importing is `govuk-frontend/govuk-frontend-5.4.0.min.js` which should be covered by `self`.

albertkol (approved):
I interviewed ChatGPT quite a bit to learn about this matter. So after a few minutes I can say I have black belt in web security. :ninja: 

Tested it a bit:
- I checked the HTTP headers for the nonce
- I saw the nonce echoed on the script HTML elements
- I checked the code for all the js script tags we have to ensure nonce is echo'ed everywhere
- I asked Chat on it's opinion about not having `nonce` on `script` elements with `src`; Apparently it's fine because we only allow scripts from our domain, with `script-src: self`.

app.py
samuelhwilliams: The most common example we have is those being the sha256's for these blocks:

```
  <script type="module">
      import { initAll } from '{{ govukFrontendJsURI }}'
      initAll()
  </script>
```

Which are now being nonced

PR #702, opened 2024-09-16
Changes to download-report job

sc15zs:
Description:
Changes command to run the download-report job in order to be able to run the job manually on test / dev. 
Uses user_id for bulk-lookup of user details from account-store, rather than azure_id.

How to test:
1. aws-vault exec <dev / test profile_name>
2. copilot job run --name download-report
3. Check email sent on Notify (_Submit monitoring and evaluation data_ service), download file.


samuelhwilliams:
I'm not immediately sure how this helps you run the script on test - what problem are you having?

sc15zs:
@samuelhwilliams 
Gideon told that if I change it to run as a module, I would not have issues with env variables:
copilot/download-report/1   File "/workspace/./find/scripts/extract_download_logs.py", line 24, in <module>
copilot/download-report/1     from config import Config
copilot/download-report/1 ModuleNotFoundError: No module named 'config'

Sam, if there is anything else to do and it's not needed, please let me know.

find/scripts/extract_download_logs.py
```diff
@@ -21,8 +21,6 @@
 from notifications_python_client import prepare_upload
 from notifications_python_client.notifications import NotificationsAPIClient
 
-from config import Config
-
```
gidsg, line 25: Can we revert changes to this file? It should work now by changing the run command only.

samuelhwilliams: Zoltan and I were chatting about this.

It doesn't work because if you try to load the `Config` object you need to inject a bunch of other environment variables, which the download-report job doesn't have access to.

It feels redundant to add them all when they're unused. The only thing that is needed is the account-store API host, and this is hardcoded directly in the service manifest.

So it feels easier to just hardcode it here too. I appreciate it's a little bit clunky - but it's that, or update the download-report job manifest to include a bunch of env vars/secrets that it has no use for (incl. the DB secret, which feels less than ideal), or rework the config object further to consider more things optional (which I don't think is good).

This feels the least-worst thing to get back on track easily.

find/scripts/extract_download_logs.py
```diff
@@ -74,23 +72,36 @@ def extract_ids_from_message(message: List[List[dict]]) -> List[str]:
...
-                f"{Config.ACCOUNT_STORE_API_HOST}/accounts",
-                headers={"Content-Type": "application/json"},
-                params={"azure_ad_subject_id": user_id},
-            )
-            response.raise_for_status()
-            response_data = response.json()
-            responses[user_id] = response_data["email_address"]
-        except requests.exceptions.RequestException as e:
-            print(f"An error occurred for user_id {user_id}: {e}")
+    aws_account_store_api_host = "http://fsd-account-store.dev.pre-award.local:8080"
```
gidsg, line 81: Please define this in the download-report manifest: https://github.com/communitiesuk/funding-service-design-post-award-data-store/blob/main/copilot/download-report/manifest.yml as `ACCOUNT_STORE_API_HOST`.

find/scripts/extract_download_logs.py
```diff
@@ -74,23 +72,36 @@ def extract_ids_from_message(message: List[List[dict]]) -> List[str]:
...
+
+    try:
+        url_params = ""
+        for id in user_ids:
+            url_params += f"account_id={id}&"
```
gidsg, line 86: Do we need to define `account_id` each time, or does it accept a comma separated list?

find/scripts/extract_download_logs.py
```diff
@@ -74,23 +72,36 @@ def extract_ids_from_message(message: List[List[dict]]) -> List[str]:
 
 def fetch_user_data(user_ids: List[str]) -> dict:
```
gidsg, line 73: Can we call this `get_email_addresses_for_user_ids` or something more descriptive?

find/scripts/extract_download_logs.py
```diff
@@ -72,33 +70,42 @@ def extract_ids_from_message(message: List[List[dict]]) -> List[str]:
...
+            headers={"Content-Type": "application/json"},
+        )
+        response.raise_for_status()
+        response_data = response.json()
+
+        for user_id in user_ids:
+            if user_id in response_data:
+                responses[user_id] = response_data[user_id]["email_address"]
+            else:
+                responses[user_id] = None
```
gidsg, line 96: Can we represent this more concisely with a list comprehension e.g.
```suggestion
responses = {user_id: response_data.get(user_id, {}).get("email_address") for user_id in user_ids}
```

find/scripts/extract_download_logs.py
```diff
@@ -72,33 +70,42 @@ def extract_ids_from_message(message: List[List[dict]]) -> List[str]:
...
+def get_email_addresses_for_user_ids(user_ids: List[str]) -> dict:
     """
-    Fetch user data from account-store API, using users' azure_ad_subject_id-s
+    Fetch user data from account-store API, using users' account_id-s
     args: user_ids: list of user_ids
     returns: dict of user_id: email_address
     """
+
+    account_store_api_host = os.getenv("ACCOUNT_STORE_API_HOST", "http://localhost:3003")
     responses = {}
```
gidsg, line 81: Can we call this `email_addresses` or something more descriptive?

find/scripts/extract_download_logs.py
sc15zs: Changed it, so it now accepts a comma separated list.

find/scripts/extract_download_logs.py
sc15zs: Changed the function name.

find/scripts/extract_download_logs.py
sc15zs: As discussed with @gidsg , Config is no longer needed, ACCOUNT_STORE_API_HOST has been added to manifest.yml, and gets retrieved directly in the script.

find/scripts/extract_download_logs.py
sc15zs: Changed it as requested.

find/scripts/extract_download_logs.py
sc15zs: Renamed the variable, as requested.

find/scripts/extract_download_logs.py
```diff
@@ -72,33 +70,42 @@ def extract_ids_from_message(message: List[List[dict]]) -> List[str]:
...
-                headers={"Content-Type": "application/json"},
-                params={"azure_ad_subject_id": user_id},
-            )
-            response.raise_for_status()
-            response_data = response.json()
-            responses[user_id] = response_data["email_address"]
-        except requests.exceptions.RequestException as e:
-            print(f"An error occurred for user_id {user_id}: {e}")
+    try:
+        url_params = ",".join(user_ids)
```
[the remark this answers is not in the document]
sc15zs: Renamed the variable.

copilot/download-report/manifest.yml
```diff
@@ -28,6 +28,7 @@ variables:                    # Pass environment variables as key value pairs.
   # Sentry DSN is OK to be public see: https://docs.sentry.io/product/sentry-basics/dsn-explainer/#dsn-utilization
   SENTRY_DSN: https://6a2623e302e641ba88eabe6675a70ddf@o1432034.ingest.sentry.io/4505390859747328
   FLASK_ENV:  ${COPILOT_ENVIRONMENT_NAME}
+  ACCOUNT_STORE_API_HOST: http://fsd-account-store.dev.pre-award.local:8080
```
gidsg, line 31: This is hard-coded to dev, so would break on other envs. I think this should work instead:
```suggestion
  ACCOUNT_STORE_API_HOST: http://fsd-account-store.${COPILOT_ENVIRONMENT_NAME}.pre-award.local:8080
```

sc15zs: Changed as requested.

find/scripts/extract_download_logs.py
```diff
@@ -72,33 +70,38 @@ def extract_ids_from_message(message: List[List[dict]]) -> List[str]:
...
+        response_data = response.json()
+
+        email_addresses = {user_id: response_data.get(user_id, {}).get("email_address") for user_id in user_ids}
+
+    except requests.exceptions.RequestException as e:
+        print(f"An error occurred: {e}")
+        for user_id in user_ids:
+            email_addresses[user_id] = None
+
+    return email_addresses
```
gidsg, line 99: Would we want to hard fail in this scenario rather than returning an empty value? If so maybe we can remove this try catch?

sc15zs: Removed the except block.

gidsg (approved):
LGTM 👍

PR #705, opened 2024-09-16
Add end-to-end tests to CI pipeline

samuelhwilliams:
### Change description
Automatically run e2e (browser) tests as past of the deployment pipeline, in the `test` environment.

We add this to the `aws_copilot_deployment` job so that the e2e tests run as part of the concurrency group configured in `deployment_cd_test_prod`, preventing e2e tests from running at the same time, or while other new code is deploying to the same environment
- which would potentially invalidate the e2e test result.


albertkol:
I see we are running the e2e tests after deployment, is it not too late at that point? 
I thought we'd want to run them before we have the code merged into `main` to flag that new code broke core functionality of the apps.

samuelhwilliams:
It's not currently possible to run the tests against a PR because that PR doesn't get deployed anywhere until merging.

These tests will stop broken code reaching production, which is still an improvement.

Running them as PR checks would be nice but we would need something like https://github.com/communitiesuk/funding-service-design-post-award-data-store/pull/590 to exist first.

albertkol:
Oh.. right.. there wouldn't be a `dev` or `test` deployed with the code.

Would there be any value for a `pytest --e2e` to run as on push branch, as we run our unit tests? Or is it too spammy?

samuelhwilliams:
What would they run against?

albertkol (approved):
There is no way to test this before merging? :boom:

samuelhwilliams:
> There is no way to test this before merging? 💥

Not without a load of faff, I think... So fingers crossed? 🤞 

(technically yes there is .. but faff ... so just YOLOing may be faster/easier)

PR #708, opened 2024-09-17
Include default context args and close gracefully, and bump playwright version

samuelhwilliams:
### Change description
This closes the browser contexts correctly at the end of each test run,
and uses playwright's `new_context` fixture which will pass default
parameters correctly, therefore allowing flags like `--video=on` to work
properly.

### DIFF
--- a/tests/e2e_tests/conftest.py
+++ b/tests/e2e_tests/conftest.py
@@ -97,11 +97,16 @@ def user_auth(
     context: BrowserContext,
     e2e_test_secrets: EndToEndTestSecrets,
 ) -> Account:
+    """This fixture sets up the browser with an auth cookie so that the test user is 'logged in' correctly.
+
+    It bypasses the standard authentication process of doing this (using Authenticator), and instead (ab)uses our
+    JWT authentication model by self-signing the blob of data that authenticator provides.
+
+    We should be careful to keep this blob of JWT data in sync with what Authenticator would actually set."""
     email_address = generate_email_address(
         test_name=request.node.originalname,
         email_domain="communities.gov.uk",
     )
-
     roles_marker = request.node.get_closest_marker("user_roles")
     user_roles = roles_marker.args[0] if roles_marker else []
 
--- a/tests/e2e_tests/conftest.py
+++ b/tests/e2e_tests/conftest.py
@@ -4,8 +4,9 @@ import warnings
 
 import jwt
 import pytest
-from playwright.sync_api import Browser, BrowserContext, ConsoleMessage, Page
+from playwright.sync_api import BrowserContext, ConsoleMessage, Page
 from pytest import FixtureRequest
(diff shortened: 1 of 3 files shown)


tests/e2e_tests/conftest.py
```diff
@@ -67,10 +68,11 @@ def authenticator_fund_config(request: FixtureRequest) -> TestFundConfig:
 
 
 @pytest.fixture
-def context(request: FixtureRequest, browser: Browser, e2e_test_secrets: EndToEndTestSecrets):
+def context(new_context: CreateContextCallback, request: FixtureRequest, e2e_test_secrets: EndToEndTestSecrets):
```
albertkol, line 71: question: is it just a syntax change here? Injecting directly `CreateContextCallback` so we don't have to do `browser.new_context`, right?

tests/e2e_tests/conftest.py
```diff
@@ -67,10 +68,11 @@ def authenticator_fund_config(request: FixtureRequest) -> TestFundConfig:
...
     e2e_env = request.config.getoption("e2e_env")
     http_credentials = e2e_test_secrets.HTTP_BASIC_AUTH if e2e_env in {"dev", "test"} else None
-    return browser.new_context(http_credentials=http_credentials)
+    context = new_context(http_credentials=http_credentials)
+    yield context
```
albertkol, line 75: question: any difference between `yield` and `return` is this context? or was `return` not closing the browser context correctly?

albertkol (approved):
:+1:

tests/e2e_tests/conftest.py
samuelhwilliams: Not quite - it's more impactful than that.

The `new_context` fixture provided by `pytest-playwright` also makes sure to inject a bunch of default arguments to the `browser.new_context` call that were previously missing when I created it directly. The fixture also properly closes down any created contexts automatically; when creating them manually I should have been calling `context.close()` after the `yield` (this is why browser tabs were staying open during test runs in `--headed` mode before).

tests/e2e_tests/conftest.py
samuelhwilliams: No different in this context. `yield` is useful in fixtures if you want to do things during test teardown (ie after returning the fixture).

This could've stayed as a `return`. I made it a `yield` because at some point while making this I added some cleanup to the fixture after the `yield`.

When I removed that cleanup I forgot to turn this back into a `return` (and also it doesn't matter)

tests/e2e_tests/conftest.py
albertkol: Oooh.. neat!

tests/e2e_tests/conftest.py
samuelhwilliams: (ive made it a return again)

PR #709, opened 2024-09-17
Remove mypy file exceptions part 1

albertkol:
FPASF-570: Part 1 (checklist)

Removes mypy exceptions from following files:
- data_store/controllers/failed_submission
- data_store/controllers/ingest_dependencies
- data_store/db/entities

### How to test
Is `mypy` happy?
Is `pytest` happy?
Does `flask db check` say that new migrations are needed?

### DIFF
--- a/data_store/controllers/failed_submission.py
+++ b/data_store/controllers/failed_submission.py
@@ -12,11 +12,11 @@ def get_failed_submission(failure_uuid: str) -> str:
     :return: the failed submission
     """
     try:
-        failure_uuid = UUID(failure_uuid, version=4)
+        uuid = UUID(failure_uuid, version=4)
     except ValueError as error:
         raise ValueError("failure_uuid is not a valid UUID.") from error
 
-    file_key = get_failed_file_key(failure_uuid)
+    file_key = get_failed_file_key(uuid)
 
     presigned_url = create_presigned_url(
         bucket_name=Config.AWS_S3_BUCKET_FAILED_FILES, file_key=file_key, filename=file_key
--- a/data_store/controllers/failed_submission.py
+++ b/data_store/controllers/failed_submission.py
@@ -4,21 +4,20 @@ from config import Config
 from data_store.aws import create_presigned_url, get_failed_file_key
 
 
-def get_failed_submission(failure_uuid: str) -> str:
+def get_failed_submission(failure_uuid: UUID) -> str:
     """Returns a presigned ULR to S3 storage to download a submission that
     matches the provided failure id.
 
     :param failure_uuid: UUID that matches the submission.
     :return: the failed submission
     """
-    try:
-        uuid = UUID(failure_uuid, version=4)
-    except ValueError as error:
-        raise ValueError("failure_uuid is not a valid UUID.") from error
 
-    file_key = get_failed_file_key(uuid)
+    file_key = get_failed_file_key(failure_uuid)
 
     presigned_url = create_presigned_url(
-        bucket_name=Config.AWS_S3_BUCKET_FAILED_FILES, file_key=file_key, filename=file_key
+        bucket_name=Config.AWS_S3_BUCKET_FAILED_FILES,
+        file_key=file_key,
+        filename=file_key,
     )
+
     return presigned_url
--- a/data_store/db/entities.py
+++ b/data_store/db/entities.py
@@ -8,7 +8,7 @@ from sqlalchemy.ext.hybrid import hybrid_property
 from sqlalchemy.orm import Mapped
 from sqlalchemy.sql.operators import and_, or_
 from sqlalchemy.orm import mapped_column, relationship
-
+from sqlalchemy.sql import expression
 from data_store.db import db
 from data_store.db.types import GUID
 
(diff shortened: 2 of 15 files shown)


data_store/controllers/failed_submission.py
```diff
@@ -12,11 +12,11 @@ def get_failed_submission(failure_uuid: str) -> str:
     :return: the failed submission
     """
     try:
-        failure_uuid = UUID(failure_uuid, version=4)
+        uuid = UUID(failure_uuid, version=4)
```
samuelhwilliams, line 15: nit: don't super love overloading the `uuid` module name. Can we make this function just take a uuid and change it upstream so that we don't need to do the casting/error handling here - or is that more faff?

data_store/db/entities.py
```diff
@@ -49,18 +49,26 @@ class Funding(BaseModel):
     project: Mapped["Project"] = sqla.orm.relationship(back_populates="funding_records")
 
     __table_args__ = (
+        # migration auto-generate does not pick up these constrainsts
+        # TODO: Create migration to apply these constraints
```
samuelhwilliams, line 53: Good spot - let's fix this ASAP, please make a ticket in the tech backlog 👀

data_store/db/entities.py
```diff
@@ -49,18 +49,26 @@ class Funding(BaseModel):
...
+                ),
+                sqla.Boolean,
             ),
             name="ck_risk_register_programme_junction_id_or_project_id",
         ),
+        # I suspect ck_funding_start_or_end_date is not applied
         # check that both start and end dates are not null at the same time
         sqla.CheckConstraint(
-            or_(start_date.isnot(None), end_date.isnot(None)),
+            "start_date IS NOT NULL OR end_date IS NOT NULL",
```
samuelhwilliams, line 68: question: why do we need to change to text here?

data_store/table_extraction/config/pf_r1_config.py
```diff
@@ -12,7 +12,7 @@
 )
 from data_store.validation.pathfinders.schema_validation.consts import PFEnums
 
-PF_TABLE_CONFIG = {
+PF_TABLE_CONFIG: dict[int | str, dict[str, dict]] = {
```
samuelhwilliams, line 15: Definitely a starting attempt 😓

data_store/transformation/pathfinders/pf_transform_r1.py
```diff
@@ -24,7 +24,7 @@ def transform(df_dict: dict[str, pd.DataFrame], reporting_round: int) -> dict[st
         row["Local Authority"]: row["Reference"][:6] for _, row in project_details_df.iterrows()
     }
     project_name_to_id_mapping = {row["Full name"]: row["Reference"] for _, row in project_details_df.iterrows()}
-    transformed = {}
+    transformed: dict[int | str, pd.DataFrame] = {}
```
samuelhwilliams, line 27: question: Do we _actually_ have int keys here ever?

samuelhwilliams (commented):
It feels like maybe `dict[int | str, pd.DataFrame]` should be a custom defined type somewhere, since we use it in so many places?

data_store/db/entities.py
albertkol: `CheckConstraint` is not accepting of `or_`'s return type. That's why I have to do the `expression.cast` above, which I think it's an ugly, hard to understand, work around to satisfy the typing. So, here, rewriting it to a string seemed both easier to read and fixed the typing, as `str` is accepted.

When I saw `CheckConstraint` not approve `or_`'s return type, it made me wonder if the code even works. That's when I noticed that the database does not list these constraints at all. So I might as well have removed them.

data_store/db/entities.py
albertkol: What makes me nervous if when we migrate these... we might already have "non-compliant" data.

data_store/table_extraction/config/pf_r1_config.py
albertkol: It seems like every time some function expects a `dict[int | str, dict[str, dict]]` type and the dict keys are manually set to always be `str`, we have to type-hint it to include `int` as well, otherwise it will infer that its type is only `dict[str, dict[str, dict]]`. Sad.

data_store/transformation/pathfinders/pf_transform_r1.py
albertkol: Explained above the behaviour I observed.

data_store/db/entities.py
samuelhwilliams: yes, that's why I say we should do (try) it sooner rather than later. Especially before the next rounds 🤞

data_store/validation/initial_validation/checks.py
```diff
@@ -144,7 +144,7 @@ def run(self, workbook: dict[int | str, pd.DataFrame], **kwargs) -> tuple[bool,
         return result, "" if result else error_message
 
 
-class SheetCheck:
+class SheetCheck(Check):
```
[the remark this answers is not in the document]
samuelhwilliams: much lazy

samuelhwilliams (approved):
optional: consider that patch I sent you and whether it's an improvement or a problem (or maybe just wait until the future when we find out any potential repercussions)

data_store/controllers/failed_submission.py
albertkol: I moved the try - catch outside this function. The function accepts directly a UUID now.

Was this was you suggested?

samuelhwilliams: Basically yes :)

PR #711, opened 2024-09-18
Remove mypy exceptions part 2

albertkol:
Remove mypy file exceptions:
- `data_store.messaging.messaging`
- `data_store/db/queries exception`
- `data_store.messaging.tf_messaging`

### How to test
Are tests passing?
Is mypy happy?

### DIFF
--- a/data_store/messaging/messaging.py
+++ b/data_store/messaging/messaging.py
@@ -45,21 +45,30 @@ def remove_errors_already_caught_by_null_failure(error_messages: list[Message]) 
         msgs.BLANK_UNIT_OF_MEASUREMENT,
     ]
 
-    cells_covered_by_null_failures = [
-        (message.sheet, cell_index)
-        for message in error_messages
-        if message.description in null_descriptions
-        for cell_index in message.cell_indexes
-    ]
+    cells_covered_by_null_failures = []
+    for message in error_messages:
+        if message.description not in null_descriptions:
+            continue
 
-    filtered_errors = [
-        message
-        for message in error_messages
-        if not any(
-            ((message.sheet, cell_index) in cells_covered_by_null_failures) for cell_index in message.cell_indexes
-        )
-        or message.description in null_descriptions
-    ]
+        if message.cell_indexes is None:
+            continue
+
+        for cell_index in message.cell_indexes:
+            cells_covered_by_null_failures.append((message.sheet, cell_index))
+
+    filtered_errors = []
+    for message in error_messages:
+        if message.cell_indexes is None:
+            continue
+
+        is_covered_by_null_failure = False
+        for cell_index in message.cell_indexes:
+            if (message.sheet, cell_index) in cells_covered_by_null_failures:
+                is_covered_by_null_failure = True
+                break
+
+        if not is_covered_by_null_failure or message.description in null_descriptions:
+            filtered_errors.append(message)
 
     return filtered_errors
 
--- a/data_store/messaging/tf_messaging.py
+++ b/data_store/messaging/tf_messaging.py
@@ -300,7 +300,7 @@ class TFMessenger(MessengerBase):
             [validation_failure.column] if isinstance(validation_failure.column, str) else validation_failure.column
         )
 
-        cell_index = tuple(
+        cell_index: tuple[str, ...] | None = tuple(
             self._construct_cell_index(
                 table=validation_failure.table,
                 column=column,
@@ -318,7 +318,7 @@ class TFMessenger(MessengerBase):
             cell_index = (
                 (self._get_cell_indexes_for_outcomes(validation_failure.failed_row),)
                 if validation_failure.failed_row is not None
-                else ("",)
+                else None
             )
         if validation_failure.expected_type == datetime:
             message = self.msgs.WRONG_TYPE_DATE.format(wrong_type=actual_type)
@@ -388,7 +388,7 @@ class TFMessenger(MessengerBase):
             [validation_failure.column] if isinstance(validation_failure.column, str) else validation_failure.column
         )
 
-        cell_index = tuple(
+        cell_index: tuple[str, ...] | None = tuple(
             self._construct_cell_index(
                 table=validation_failure.table,
                 column=column,
@@ -412,7 +412,7 @@ class TFMessenger(MessengerBase):
                 cell_index = (
                     (self._get_cell_indexes_for_outcomes(validation_failure.failed_row),)
                     if validation_failure.failed_row is not None
-                    else ("",)
+                    else None
                 )
         elif sheet == "Funding Profiles":
             message = self.msgs.BLANK_ZERO
--- a/data_store/messaging/tf_messaging.py
+++ b/data_store/messaging/tf_messaging.py
@@ -296,7 +296,7 @@ class TFMessenger(MessengerBase):
         actual_type = self.INTERNAL_TYPE_TO_MESSAGE_FORMAT[validation_failure.actual_type]
         cell_index = self._construct_cell_index(
             table=validation_failure.table,
-            column=validation_failure.column,
+            column=str(validation_failure.column),
             row_index=validation_failure.row_index,
         )
 
@@ -305,8 +305,11 @@ class TFMessenger(MessengerBase):
                 "Financial Year 2022/21 - Financial Year 2029/30",
                 ("Outcome Indicators (excluding " "footfall) and Footfall Indicator"),
             )
-            cell_index = self._get_cell_indexes_for_outcomes(validation_failure.failed_row)
-
+            cell_index = (
+                ""
+                if validation_failure.failed_row is None
+                else self._get_cell_indexes_for_outcomes(validation_failure.failed_row)
+            )
         if validation_failure.expected_type == datetime:
             message = self.msgs.WRONG_TYPE_DATE.format(wrong_type=actual_type)
         elif sheet == "PSI":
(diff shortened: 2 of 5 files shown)


data_store/messaging/messaging.py
```diff
@@ -45,21 +45,30 @@ def remove_errors_already_caught_by_null_failure(error_messages: list[Message])
...
-        )
-        or message.description in null_descriptions
-    ]
+    cells_covered_by_null_failures = []
+    for message in error_messages:
+        if message.description not in null_descriptions:
+            continue
+
+        if message.cell_indexes is None:
+            continue
```
samuelhwilliams, line 54: this feels new - why do we need this now? is this defo right?

data_store/messaging/messaging.py
```diff
@@ -45,21 +45,30 @@ def remove_errors_already_caught_by_null_failure(error_messages: list[Message])
...
+    filtered_errors = []
+    for message in error_messages:
+        if message.cell_indexes is None:
+            continue
+
+        is_covered_by_null_failure = False
+        for cell_index in message.cell_indexes:
+            if (message.sheet, cell_index) in cells_covered_by_null_failures:
+                is_covered_by_null_failure = True
+                break
```
samuelhwilliams, line 68:
```suggestion
        is_covered_by_null_failure = any(
            (message.sheet, cell_index) in cells_covered_by_null_failures
            for cell_index in message.cell_indexes
        )
```

maybe

albertkol: I don't feel crazy strong about it. It was an `any()` before. I rewrote it because the line was too long. If the formatter would break it in multiple line's as it shows in your commit suggestion, I would add the suggestion.

data_store/messaging/tf_messaging.py
```diff
@@ -296,7 +296,7 @@ def _wrong_type_failure_message(self, validation_failure: WrongTypeFailure) -> M
         actual_type = self.INTERNAL_TYPE_TO_MESSAGE_FORMAT[validation_failure.actual_type]
         cell_index = self._construct_cell_index(
             table=validation_failure.table,
-            column=validation_failure.column,
+            column=str(validation_failure.column),
```
albertkol, line 299: This one I am not proud of, but it was the simplest way out. I will explain it a bit.

`validation_failure.column` can be `str | list[str]`
`_construct_cell_index` accepts only `column: str`. 

It only happens for one type of `validation_failure` to have a `column` as `list[str]`. When `column` is a list, we loop call `_construct_cell_index` for each column. 

The correct approach is to always loop call `_construct_cell_index` for each column regardless of the type of `validation_failure`. But this was perhaps enough to avoid problems for the rest of the validation failures.

data_store/messaging/messaging.py
samuelhwilliams: very reasonable then :D 👏

data_store/messaging/tf_messaging.py
```diff
@@ -305,8 +305,11 @@ def _wrong_type_failure_message(self, validation_failure: WrongTypeFailure) -> M
                 "Financial Year 2022/21 - Financial Year 2029/30",
                 ("Outcome Indicators (excluding " "footfall) and Footfall Indicator"),
             )
-            cell_index = self._get_cell_indexes_for_outcomes(validation_failure.failed_row)
-
+            cell_index = (
+                ""
+                if validation_failure.failed_row is None
+                else self._get_cell_indexes_for_outcomes(validation_failure.failed_row)
+            )
```
albertkol, line 312: `validation_failure.failed_row` can be `pb.Series | None`
`_get_cell_indexes_for_outcomes` accepts only `failed_row: pb.Series`

I think `validation_failure.failed_row` should never be `None` inside that `if` (when `sheet == Outcome`).
Not sure thought if that's a rule we can rely on.

I added tests to cover this scenario. Message() errors our when cell_index is "". 
I am also considered changing `if sheet == Outcome` to `if sheet == Outcome and validation_failure.failed_row is not None`. Perhaps it's the better approach.

data_store/messaging/tf_messaging.py
```diff
@@ -400,8 +407,8 @@ def _generic_failure(self, validation_failure: GenericFailure) -> Message:
         if not validation_failure.cell_index:
             validation_failure.cell_index = self._construct_cell_index(
                 validation_failure.table,
-                validation_failure.column,
-                validation_failure.row_index,
+                str(validation_failure.column),
+                validation_failure.row_index or 0,
```
albertkol, line 411: Not proud of this one either.

data_store/messaging/tf_messaging.py
```diff
@@ -447,6 +454,9 @@ def _get_cell_indexes_for_outcomes(self, failed_row: pd.Series) -> str:
         financial_year = self._get_uk_financial_year_start(start_date)
         index = failed_row.name
 
+        if not isinstance(index, int):
+            raise TypeError(f"Cell index not int for failed row {failed_row}")
```
albertkol, line 458: Looking at the tests we seem to assume `index` is always an `int`. But it seems to be returning a `Hashable | None`.

data_store/messaging/tf_messaging.py
samuelhwilliams: stop doing stuff you're not proud of :D

data_store/messaging/tf_messaging.py
samuelhwilliams: If it's a `list[str]` then stringifying it will give you something like `['A']`, which will feed into `column_letter = self.TABLE_AND_COLUMN_TO_ORIGINAL_COLUMN_LETTER[table][column]` and then throw a KeyError. IE it would be broken.

If it can ever be a `list[str]` in reality, we should handle this better otherwise we're just getting rid of a mypy error but still leaving the code broken :(

data_store/messaging/tf_messaging.py
samuelhwilliams: Yeah I think pandas treats a few things as just 'Hashables' - but I think in reality maybe one of our code assumptions is that it is _actually_ always an `int` as we use the library. Possibly something we can fix with tighter type hints somewhere ...

data_store/messaging/tf_messaging.py
albertkol: :disappointed: fine. I will loop them. 

But there is this feeling that having both str and list of str as a type is the bigger problem. But that's much harder to fix.

samuelhwilliams: don't disagree necessarily...

data_store/messaging/tf_messaging.py
```diff
@@ -397,18 +429,23 @@ def _unauthorised_submission_failure(self, validation_failure: UnauthorisedSubmi
...
-                validation_failure.column,
-                validation_failure.row_index,
+                table=validation_failure.table,
+                column=validation_failure.column,
+                row_index=validation_failure.row_index or 0,
             )
-        sheet = self.INTERNAL_TABLE_TO_FORM_SHEET[validation_failure.table]
+
+            cell_indexes = (validation_failure.cell_index,)
+        else:
```
samuelhwilliams, line 442: mr "i hate else's"

data_store/messaging/tf_messaging.py
```diff
@@ -343,13 +355,21 @@ def _invalid_enum_value_failure_message(self, validation_failure: InvalidEnumVal
             project_number = get_project_number_by_position(validation_failure.row_index, validation_failure.table)
             section = f"Project Funding Profiles - Project {project_number}"
 
-        cell_index = self._construct_cell_index(
-            table=validation_failure.table,
-            column=validation_failure.column,
-            row_index=validation_failure.row_index,
+        # if column is a str make it a list
+        columns = (
+            [validation_failure.column] if isinstance(validation_failure.column, str) else validation_failure.column
```
samuelhwilliams, line 360: Should we make `validation_failure.column` actually just be `validation_failure.columns` and only `list[str]`? Did you look into that and it's a pain?

data_store/messaging/tf_messaging.py
```diff
@@ -381,13 +409,17 @@ def _non_nullable_constraint_failure_message(self, validation_failure: NonNullab
             if column == "Financial Year 2022/21 - Financial Year 2025/26":
                 section = "Outcome Indicators (excluding footfall) / Footfall Indicator"
                 message = self.msgs.BLANK_ZERO
-                cell_index = self._get_cell_indexes_for_outcomes(validation_failure.failed_row)
+                cell_index = (
+                    (self._get_cell_indexes_for_outcomes(validation_failure.failed_row),)
+                    if validation_failure.failed_row is not None
+                    else ("",)
```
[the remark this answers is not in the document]
albertkol: In practice I think the `else` branch will never happen. All tests had `failed_row`.

If it would end up reaching it.. the code will throw an exception a bit later. Added a test for that case.

I will say that I don't know how `validation_failures` are built and which types has which values.

Is it weird? If there is no `failed_row` you get no `cell_index`. But Message class only likes `tuple[str, ...]`.

data_store/messaging/tf_messaging.py
albertkol: I considered it a few times changing class definitions. But it's quite an intimidating change to make.
We should look into all these classes, but perhaps as a separate exercise when we focus on this specifically.

Personally I am rather tempted to have `column` always as a `str`. Out of all validation failure types only one is breaking the pattern.

samuelhwilliams (approved):

PR #712, opened 2024-09-18
Use ISO date format in test fixtures

samuelhwilliams:
### Change description
We seed our test DB from a bunch of CSVs. These CSVs have dates in the format xx/xx/xxxx, which in some places look like they mean dd/mm/yyyy and in others are actually mm/dd/yyyy.

To prevent confusion, let's use the ISO8601 date format instead: yyyy-mm-dd


wjrm500 (approved):
Agreed current way of doing things is super confusing, makes my head hurt haha. Nice change to see!

PR #713, opened 2024-09-18
Run type checks in github actions

samuelhwilliams:
We are starting to but up against problems using precommit.ci because they "only" allocate ~250mb storage for python dependencies, and we are starting to need to install more than that to allow `mypy` to run with all of the code context it needs.

So this PR moves the `mypy` type check into github actions, but leaves everything else running in precommit.ci since some of the devs like how that works.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,10 @@
 ci:
   autofix_prs: false
 
+  # mypy needs a lot of dependencies installed, and we hit limits in precommit.ci on the size of the dependencies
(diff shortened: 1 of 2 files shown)


.pre-commit-config.yaml
```diff
@@ -1,5 +1,6 @@
 ci:
   autofix_prs: false
+  skip: [mypy]
```
gidsg, line 3: Can we add a comment here explaining we skip because of the size limitation?

gidsg (approved):
LGTM 👍

PR #714, opened 2024-09-18
Use a system hook for pre-commit mypy

samuelhwilliams:
### Change description
Now that we're running mypy in github actions, where python is installed and already has all of our dev dependencies available, we can use a system hook instead of the mypy repo hook.

This will make pre-commit run the `mypy` that is available in the current environment, which will be the one pinned by our requirements-dev.txt. Because it uses that mypy, and the environment's python interpreter, all of our dev dependencies will also inherently be available for mypy's type checking, which should give us the best checks available.

### How to test


albertkol (approved):
.

PR #715, opened 2024-09-18
Update test fixtures for ReportingRound model

samuelhwilliams:
### Change description
In some previous code, we created the ReportingRound model and instances to replace a bunch of disparate data spread across various models.

We didn't update all of the tests to consistently have ReportingRound entities attached via relationships, so this patch fixes it.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -11,6 +11,7 @@ For particularly disruptive or scope changing tests, the function scoped fixture
 since these reset the DB at a function scope, they have the biggest impact on performance.
 """
 
+import uuid
 from datetime import datetime
 from pathlib import Path
 from typing import Any, BinaryIO, Generator
@@ -43,6 +44,7 @@ from data_store.db.entities import (
     ProgrammeProgress,
     Project,
     ProjectFinanceChange,
+    ReportingRound,
     RiskRegister,
     Submission,
 )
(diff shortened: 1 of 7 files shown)


wjrm500 (approved):
Hi @samuelhwilliams looks good and makes sense to have this as a separate PR to minimise the scope of upcoming PRs. Should probably have included this in my original PR, although looking back there were already about a million files in the changeset, so maybe for the best that I didn't haha!

Couple of quick thoughts, no blockers:
- Slight inconsistency in the way we get ID from funds - generating UUID in the application code here: `tests/conftest.py::additional_test_data` versus flushing session to attach ID to entity here: `tests/integration_tests/test_ingest_db_transactions.py::test_next_submission_id_more_digits`
- More generally, is it possible to define all reference data upfront using CSVs so we don't even need to create new funds / organisations or whatever during testing? (Probably a good reason we don't do this, just checking)
- The `ReportingRound` entity defined in the test functions `test_next_submission_numpy_type` and `test_next_submission_id_more_digits` in `tests/integration_tests/test_ingest_db_transactions.py` could be a fixture, as it is identical between those functions

samuelhwilliams:
> * Slight inconsistency in the way we get ID from funds - generating UUID in the application code here: tests/conftest.py::additional_test_data versus flushing session to attach ID to entity here: tests/integration_tests/test_ingest_db_transactions.py::test_next_submission_id_more_digits

fair point, will normalise this as a fixup in next PR 👀 

> * More generally, is it possible to define all reference data upfront using CSVs so we don't even need to create new funds / organisations or whatever during testing? (Probably a good reason we don't do this, just checking)

maybe - not something I'm investigating for now, but might be a good thing for tech debt in the future

> * The ReportingRound entity defined in the test functions test_next_submission_numpy_type and test_next_submission_id_more_digits in tests/integration_tests/test_ingest_db_transactions.py could be a fixture, as it is identical between those functions

probably could be, but I am not overly bothered about making _all_ test setup into fixtures (esp with only two copies)

PR #718, opened 2024-09-19
New unique constraint for programme junction

samuelhwilliams:
https://dluhcdigital.atlassian.net/browse/FPASF-493

### Change description
Because we'll be removing the other unique constraint that currently looks at programme_id + reporting_round (number).

### DIFF
--- a/data_store/db/entities.py
+++ b/data_store/db/entities.py
@@ -450,6 +450,11 @@ class ProgrammeJunction(BaseModel):
             "reporting_round",
             name="uq_programme_junction_unique_submission_per_round",
         ),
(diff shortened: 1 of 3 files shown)


wjrm500 (approved):
Looks good, couple of non-blocking thoughts, both of which I've mentioned over Slack but adding here to keep PRs self-documenting:

- Alembic migration name truncation leading to things like `041_new_programme_round_constrai` is a bit bleurgh, doesn't appear that we can do anything about this sadly
- Prefer a nicer name for the new unique key, something standardised like `uq_programme_junction_programme_id_reporting_round_id` that doesn't try to convey any additional meaning

PR #722, opened 2024-09-20
Remove mypy exceptions part 3

albertkol:
Remove mypy exceptions from files:
- data_store.util 
- scripts.validate
- schema_validation_tests/test_pf_schema_validate
- table_extraction/extract 
- pathfinders/test_pf_transform_r1
- data_store/table_extraction/process

# How to test?
Is `mypy` check passing?
Is `pytest` check passing?

### DIFF
--- a/data_store/table_extraction/extract.py
+++ b/data_store/table_extraction/extract.py
@@ -72,10 +72,10 @@ class TableExtractor:
     def _get_tags(self, id_tag: str, worksheet: pd.DataFrame) -> tuple[list[Cell], list[Cell]]:
         start_tag = self.START_TAG.format(id=id_tag)
         end_tag = self.END_TAG.format(id=id_tag)
-        start_tags = list(zip(*np.where(worksheet == start_tag), strict=False))
-        end_tags = list(zip(*np.where(worksheet == end_tag), strict=False))
-        start_tags = [Cell(row, col) for row, col in start_tags]
-        end_tags = [Cell(row, col) for row, col in end_tags]
+        filtered_start_tags = list(zip(*np.where(worksheet == start_tag), strict=False))
+        filtered_end_tags = list(zip(*np.where(worksheet == end_tag), strict=False))
+        start_tags = [Cell(row, col) for row, col in filtered_start_tags]
+        end_tags = [Cell(row, col) for row, col in filtered_end_tags]
         if not start_tags and not end_tags:
             raise TableExtractionError(f"No {id_tag} tags found.")
         if len(start_tags) != len(end_tags):
--- a/data_store/table_extraction/extract.py
+++ b/data_store/table_extraction/extract.py
@@ -72,8 +72,8 @@ class TableExtractor:
     def _get_tags(self, id_tag: str, worksheet: pd.DataFrame) -> tuple[list[Cell], list[Cell]]:
         start_tag = self.START_TAG.format(id=id_tag)
         end_tag = self.END_TAG.format(id=id_tag)
-        filtered_start_tags = list(zip(*np.where(worksheet == start_tag), strict=False))
-        filtered_end_tags = list(zip(*np.where(worksheet == end_tag), strict=False))
+        filtered_start_tags = zip(*np.where(worksheet == start_tag), strict=False)
+        filtered_end_tags = zip(*np.where(worksheet == end_tag), strict=False)
         start_tags = [Cell(row, col) for row, col in filtered_start_tags]
         end_tags = [Cell(row, col) for row, col in filtered_end_tags]
         if not start_tags and not end_tags:
--- a/data_store/table_extraction/process.py
+++ b/data_store/table_extraction/process.py
@@ -89,7 +89,7 @@ class TableProcessor:
             self._drop_empty_rows(table)
 
         if self.drop_empty_tables and table.df.empty:
-            table.df = pd.DataFrame()
+            table.df = None  # type: ignore[assignment]
 
     def _lift_header(self, table: Table) -> None:
         header: DataFrame = table.df.iloc[: self.num_header_rows]
--- a/data_store/table_extraction/process.py
+++ b/data_store/table_extraction/process.py
@@ -1,5 +1,6 @@
 import numpy as np
 import pandas as pd
+from pandas import DataFrame, Index
 
 from data_store.table_extraction.exceptions import TableProcessingError
 from data_store.table_extraction.table import Table
@@ -75,6 +76,7 @@ class TableProcessor:
         :param table: The table to process.
         :return: None
         """
+
         self._strip_whitespace(table)
         self._lift_header(table)
         self._remove_merged_headers(table)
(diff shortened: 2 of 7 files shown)


data_store/util.py
```diff
@@ -25,13 +25,19 @@ def get_postcode_prefix_set(postcodes: list[str]) -> set[str]:
...
     for postcode in postcodes:
-        postcode = postcode.strip()
-        postcode_area_matches = re.match(POSTCODE_PREFIX_REGEX, postcode)
-        postcode_prefix = postcode_area_matches.group(0)
-        postcodes_prefix_set.add(postcode_prefix.upper())
-    return postcodes_prefix_set
+        postcode_matches = re.match(POSTCODE_PREFIX_REGEX, postcode.strip())
+
+        if postcode_matches is None:
+            continue
```
albertkol, line 35: `match()` can return `None`. `.group(0)` on `None` upsets mypy.

tests/data_store_tests/validation_tests/pathfinders/schema_validation_tests/test_pf_schema_validate.py
```diff
@@ -101,8 +101,11 @@ def dataframe_with_missing_indexes():
 
 
 def build_mock_extracted_table(data: dict[str, list[Any]]) -> Table:
-    data = pd.DataFrame(data=data, index=range(len(list(data.values())[0])))
```
albertkol, line 104: Function has `data` as `dict[str, list[Any]]`. Reassigning it with a different type upsets mypy. I think reassigning in general upsets mypy. Don't upset mypy.

data_store/table_extraction/extract.py
```diff
@@ -72,10 +72,10 @@ def extract(self, worksheet_name: str, id_tag: str) -> list[Table]:
     def _get_tags(self, id_tag: str, worksheet: pd.DataFrame) -> tuple[list[Cell], list[Cell]]:
         start_tag = self.START_TAG.format(id=id_tag)
         end_tag = self.END_TAG.format(id=id_tag)
-        start_tags = list(zip(*np.where(worksheet == start_tag), strict=False))
-        end_tags = list(zip(*np.where(worksheet == end_tag), strict=False))
-        start_tags = [Cell(row, col) for row, col in start_tags]
-        end_tags = [Cell(row, col) for row, col in end_tags]
```
albertkol, line 78: Reassigning `start_tags` and `end_tags`  upsets mypy.

data_store/table_extraction/process.py
```diff
@@ -75,38 +76,45 @@ def process(self, table: Table) -> None:
...
         self._remove_ignored_non_header_rows(table)
         self._replace_dropdown_placeholder(table)
         self._drop_bespoke_rows(table)
+
         if self.drop_empty_rows:
             self._drop_empty_rows(table)
+
         if self.drop_empty_tables and table.df.empty:
-            table.df = None
+            table.df = pd.DataFrame()
```
albertkol, line 92: `Table` class has table.db defined as `pd.DataFrame` type. You cannot make it `None`, unless you change the typing of the class property. `pd.DataFrame()` is how you represent an empty `DataFrame`. 

Tests needed updating from checking: 
`table.df is None` to `table.db.empty`
`table.df is not None` to `not table.db.empty`

data_store/table_extraction/process.py
```diff
@@ -75,38 +76,45 @@ def process(self, table: Table) -> None:
...
 
     def _lift_header(self, table: Table) -> None:
-        header = table.df.iloc[: self.num_header_rows]
-        header = self._concatenate_headers(header, headers_to_ffill=self.merged_header_rows)
-        table.df.columns = header
+        header: DataFrame = table.df.iloc[: self.num_header_rows]
+        table.df.columns = Index(
```
albertkol, line 96: table.df.columns accepts only an Index type. Easily fixed by passing the list to the `Index`.

data_store/table_extraction/process.py
```diff
@@ -75,38 +76,45 @@ def process(self, table: Table) -> None:
...
+                header,
+                headers_to_ffill=self.merged_header_rows,
+            )
+        )
         table.df = table.df.iloc[self.num_header_rows :]
 
     @staticmethod
     def _remove_merged_headers(table: Table) -> None:
-        prev_col = ("", "")
+        prev_col: tuple[str, int] = ("", -1)  # Initialize with valid types
```
albertkol, line 106: This is a but hard to decipher because some seem to like using walrus operators in comprehension lists.
Long story short `prev_col` stores `(col_name: str, idx: int)` but is initialised incorrectly as `(str, str)`.

data_store/table_extraction/extract.py
samuelhwilliams:
```suggestion
        filtered_start_tags = zip(*np.where(worksheet == start_tag), strict=False)
        filtered_end_tags = zip(*np.where(worksheet == end_tag), strict=False)
```

nit: Don't need to turn them into lists if they're only going to be used once to build the lists later, they could stay as generators

data_store/table_extraction/process.py
samuelhwilliams: This change makes the attribute `drop_empty_tables` and the docstring `drop_empty_tables (bool, optional): Whether to drop tables that are entirely empty (default is False).` confusing.

Easier to update the type hint to `db: pd.DataFrame | None`?

albertkol: I don't completely understand this condition:
`if self.drop_empty_tables and table.df.empty` 

I was thinking that whenever `table.df.empty == True`. Doing `table.df = pd.DataFrame()` is redundant. 
Gave it a shot, I removed the whole if and tests still pass..
```
if self.drop_empty_tables and table.df.empty:
    table.df = pd.DataFrame()
```

So it is probably redundant as it's written and used now.

Not sure if making the `table.df` to be `pd.DataFrame | None` would be better. Would have me fixing 19 errors as a result.

albertkol: `drop_empty_tables` is never used. 

I am unsure how empty `df`'s are different from `None` `df`'s..

albertkol: May I delete the property?

samuelhwilliams:
```
if self.drop_empty_tables and table.df.empty:
    table.df = pd.DataFrame()
```

Yes this is redundant, which is why it was doing `table.df = None` before. I guess without deep reading the code I would rather it stay as `table.df = None`.

If we _don't_ do that, then `drop_empty_tables` no longer makes sense as an attribute so we should change something there too.

data_store/table_extraction/process.py
samuelhwilliams:
```suggestion
        prev_col: tuple[str, int]
```

It is, I believe, possible to do this as well if the initialisation is not actually useful

data_store/table_extraction/process.py
albertkol: I removed it. Which does indeed change behaviour in theory. 
In practice, I cannot see the `drop_empty_tables` flag used anywhere.

data_store/table_extraction/process.py
```diff
@@ -75,38 +76,45 @@ def process(self, table: Table) -> None:
...
         self._remove_ignored_non_header_rows(table)
         self._replace_dropdown_placeholder(table)
         self._drop_bespoke_rows(table)
+
         if self.drop_empty_rows:
             self._drop_empty_rows(table)
+
         if self.drop_empty_tables and table.df.empty:
-            table.df = None
+            table.df = None  # type: ignore[assignment]
```
albertkol, line 92: @samuelhwilliams Ignoring it. At least for now. Reverted the tests too.

I don't think the `| None` would be the good approach though. 

I think in theory we either make `pd.DataFrame` always work, or we delete it completely, if unused.

samuelhwilliams: Look at yourself in the mirror Albert. Look who you've become.

samuelhwilliams (approved):

samuelhwilliams (approved):

PR #727, opened 2024-09-20
Stop writing to old reporting round columns

samuelhwilliams:
https://dluhcdigital.atlassian.net/browse/FPASF-494

### Change description
We used to store some reporting round data on Submission and
ProgrammeJunction, but now we have a dedicated ReportingRound table.

We've stopped reading from the old columns, so now let's stop writing to
them. Then we can finally remove them altgother in a future PR.

### How to test


data_store/transformation/pathfinders/pf_transform_r1.py
```diff
@@ -123,15 +123,12 @@ def transform(df_dict: dict[str, pd.DataFrame], reporting_round: int) -> dict[st
 
 def _submission_ref(
```
wjrm500, line 124: Style thing but now we've removed the second arg, this would look neater as a one-line signature. Slight shame that our current auto-formatters don't handle this sort of thing

db/migrations/versions/042_remove_old_reporting_round_d.py
```diff
@@ -0,0 +1,35 @@
+"""remove old reporting round data
```
wjrm500, line 1: Maybe slightly misnamed as the migration doesn't remove data, just makes the columns nullable

wjrm500 (approved):
Looks good, pending merge conflict resolution. Couple of extra notes...

- The docstring for `download_data_base_query` (`funding-service-design-post-award-data-store/data_store/db/queries.py`) mentions...
```
:param min_rp_start: Minimum Reporting Period Start to filter by
:param max_rp_end: Maximum Reporting Period End to filter by
```
...which is slightly confusing now given the move away from these fields - similar story with the docstring for `set_submission_period_condition` (same module) which mentions "filtering on Submission date field entities"
- Do we want to make `reporting_round_id` on `ProgrammeJunction` and `Submission` tables non-nullable? Probably out of scope for this PR, just something I spotted

data_store/transformation/pathfinders/pf_transform_r1.py
samuelhwilliams: This is specifically left to the user to decide. If you add a trailing comma, it will be split over multiple lines. If you don't, then it will collapse it down as far as possible.

I think generally multiline is 'better', because the git diff is simplified when adding/removing lines. But it's so so minor that I don't pay attention to it any more really.

samuelhwilliams:
> Do we want to make reporting_round_id on ProgrammeJunction and Submission tables non-nullable? Probably out of scope for this PR, just something I spotted

Yeah good spot - I do need to do this. I'll do another PR for it just in case it's painful for some reason...

PR #730, opened 2024-09-23
Remove definitions of old reporting round columns

samuelhwilliams:
https://dluhcdigital.atlassian.net/browse/FPASF-494

### Change description
Removes the old reporting round columns from python model defintions. Once this is out, we know that no code is using those columns, and we can remove them safely in a future migration PR.

**Note**: I know that the DB migrations check will fail, because there are outstanding migrations. But this is intentional.

### How to test

### DIFF
--- a/data_store/db/entities.py
+++ b/data_store/db/entities.py
@@ -416,7 +416,6 @@ class ProgrammeJunction(BaseModel):
     )
     programme_id: Mapped[GUID] = sqla.orm.mapped_column(sqla.ForeignKey("programme_dim.id"), nullable=False)
     reporting_round_id: Mapped[GUID] = sqla.orm.mapped_column(sqla.ForeignKey("reporting_round.id"), nullable=False)
-    reporting_round = sqla.Column(sqla.Integer, nullable=True)
 
     # parent relationships
     submission: Mapped["Submission"] = sqla.orm.relationship(back_populates="programme_junction", single_parent=True)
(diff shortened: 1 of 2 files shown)


samuelhwilliams:
(requires https://github.com/communitiesuk/funding-service-design-post-award-data-store/pull/733 first)

data_store/db/entities.py
```diff
@@ -416,7 +416,6 @@ class ProgrammeJunction(BaseModel):
     )
     programme_id: Mapped[GUID] = sqla.orm.mapped_column(sqla.ForeignKey("programme_dim.id"), nullable=False)
     reporting_round_id: Mapped[GUID] = sqla.orm.mapped_column(sqla.ForeignKey("reporting_round.id"), nullable=False)
-    reporting_round = sqla.Column(sqla.Integer, nullable=True)
```
albertkol, line 419: Wouldn't there be a migration with this?

albertkol: Found it!

albertkol: Would've kept the to PR's one.

samuelhwilliams: it's intentional to not do them together, because doing them together is 'theoretically' broken (due to how sqlalchemy builds SQL queries). happy to chat through if useful

samuelhwilliams: actually I'll just explain here.

SQLAlchemy doesn't do a `select *` statement - it will do an explicit `select col1, col2, col` for all columns it knows about based on the python model.

Because we run migrations before deploying the code (as is required), if you remove the python column definitions in the same PR a you have a migration to remove the columns, you can end up with the old code trying to query columns that have been dropped by the migration before the new code has deployed.

PR #732, opened 2024-09-24
Remove secrets & large objects from logging

sc15zs:
Ticket:
https://dluhcdigital.atlassian.net/browse/FPASF-538

Description:
We want to remove any secrets from logging and also ensure that only basic, immutable data types are being logged.

### DIFF
--- a/find/main/routes.py
+++ b/find/main/routes.py
@@ -107,11 +107,15 @@ def download():
             if reporting_period_end:
                 query_params["rp_end"] = reporting_period_end
 
-            query_params_without_email_address = {k: v for k, v in query_params.items() if k != "email_address"}
+            query_params_without_email_address = {
+                k: v
+                for k, v in query_params.items()
+                if k in ["file_format", "organisations", "regions", "funds", "outcome_categories", "rp_start", "rp_end"]
+            }
             try:
                 trigger_async_download(query_params)
                 current_app.logger.info(
-                    "Request for download by {user_id} with {query_params}",
+                    "Request for download by user with ID: {user_id} with query params: {query_params}",
                     extra={
                         "user_id": g.account_id,
                         "query_params": query_params_without_email_address,
--- a/submit/main/routes.py
+++ b/submit/main/routes.py
@@ -82,7 +82,7 @@ def upload(fund_code, round):  # noqa: C901
             # Pre-validation failure
             if Config.ENABLE_VALIDATION_LOGGING:
                 for pre_err in pre_errors:
-                    current_app.logger.info("Pre-validation error: {error}", extra=dict(error=pre_err))
+                    current_app.logger.info("Pre-validation error: {error}", extra=dict(error=str(pre_err)))
             else:
                 current_app.logger.info(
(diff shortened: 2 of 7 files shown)


find/main/routes.py
```diff
@@ -111,10 +112,10 @@ def download():
             try:
                 trigger_async_download(query_params)
                 current_app.logger.info(
-                    "Request for download by {user_id} with {query_params}",
+                    "Request for download by user with ID: {user_id} with query params: {query_params}",
                     extra={
                         "user_id": g.account_id,
-                        "query_params": query_params_without_email_address,
+                        "query_params": json.dumps(query_params_without_email_address),
```
samuelhwilliams, line 118: I think there's a bit of miscommunication on my end here.

The ticket said something like "don't log mutable objects like lists and dicts". The reasoning behind this is because it's easier for those objects to, at some point in the future, have extra data inserted into them that might contain personal data we don't want to log. It's not because lists and dicts are inherently bad objects that can't be logged but strings and ints are OK.

So converting this dict to a string doesn't really meet our objectives - we'll still be logging whatever is inside that dict.

I think the actual principle here is more - don't like things where it isn't immediately obvious what is being logged, and where changes elsewhere might easily change what is being logged here.

This log call is _basically_ fine already because we explicitly craft the logged data in very close proximity to the logging call, so it would be hard for extra things to sneak in. However an even safer approach might be an allowlist rather than a blocklist:

```python
query_params_without_email_address = {k: v for k, v in query_params.items() if k in ['organisations', 'regions', 'funds', 'outcome_categories', 'rp_start', 'rp_end']}
try:
    trigger_async_download(query_params)
    current_app.logger.info(
        "Request for download by user with ID: {user_id} with query params: {query_params}",
        extra={
            "user_id": g.account_id,
            "query_params": json.dumps(query_params_without_email_address),
            "request_type": "download",
        },
    )
```

submit/main/routes.py
```diff
@@ -137,8 +137,8 @@ def upload(fund_code, round):  # noqa: C901
 
             metadata["User"] = g.user.email
             current_app.logger.info(
-                "Upload successful for {fund} round {round}: {metadata}",
-                extra=dict(metadata=metadata, fund=fund_code, round=round),
+                "Upload successful for {fund} round {round}: {metadata_str}",
+                extra=dict(metadata_str=json.dumps(metadata), fund=fund_code, round=round),
```
samuelhwilliams, line 141: We are logging the user's email address here - as you can see we inject it into the `metadata` blob directly above.

This is what I talk about in my last comment - it's not that "dicts can't be logged and strings can be logged" - it's that referencing dict variables makes it easy for sensitive data to sneak in.

Let's remove the email address from here.

find/main/routes.py
sc15zs: Makes sense, thank you! Changed it.

submit/main/routes.py
sc15zs: Changed email address to user ID.

samuelhwilliams (approved):

PR #736, opened 2024-09-24
FPASF-583 - Support new 'Project progress' validation requirement for TF R6

wjrm500:
### Tickets

[Update TF validation for next reporting round](https://dluhcdigital.atlassian.net/browse/FPASF-185)
[Make necessary coding changes](https://dluhcdigital.atlassian.net/browse/FPASF-583)

### Description

Added a new validation rule for the Projects Progress Summary table. It checks that projects marked as "Not yet started" don't have a Start Date in the past. If violated, it triggers a `DATA_MISMATCH_PROJECT_START` error with the content "A Project has not yet started but has a Start Date prior to today declared."

### How to test

- Create a copy of the TF R5 spreadsheet (`funding-service-design-post-award-data-store/tests/integration_tests/mock_tf_returns/TF_Round_5_Success.xlsx`)
- In the copy, change cell G21 on tab "3 - Programme Progress" to "Not yet started"
- In `funding-service-design-post-award-data-store/data_store/controllers/ingest_dependencies.py`, add an import `import data_store.validation.towns_fund.fund_specific_validation.fs_validate_r6 as tf_r6_validate` and for `case ("Towns Fund", 5)`, swap `fund_specific_validation=tf_r4_validate.validate` for `fund_specific_validation=tf_r6_validate.validate`
- Run `data-store` container locally and upload the changed spreadsheet copy at `http://submit-monitoring-data.levellingup.gov.localhost:4001/upload/TF/5`
- Note this row in error table:
Programme progress | Projects progress summary | D21 | A Project has not yet started but has a Start Date prior to today declared

### DIFF
--- a/data_store/messaging/tf_messaging.py
+++ b/data_store/messaging/tf_messaging.py
@@ -28,6 +28,10 @@ class TFMessages(SharedMessages):
     MISSING_OTHER_FUNDING_SOURCES = (
         "You’ve not entered any Other Funding Sources. You must enter at least 1 over all projects."
     )
+    DATA_MISMATCH_PROJECT_START = (
+        "You've entered a project start date of today or earlier, but the project delivery status has been entered as "
+        "'Not yet started'. Add a valid start date or change the status."
+    )
 
 
 class TFMessenger(MessengerBase):
--- a/data_store/messaging/tf_messaging.py
+++ b/data_store/messaging/tf_messaging.py
@@ -29,8 +29,8 @@ class TFMessages(SharedMessages):
         "You’ve not entered any Other Funding Sources. You must enter at least 1 over all projects."
     )
     DATA_MISMATCH_PROJECT_START = (
-        "You've entered a project start date of today or earlier, but the project delivery status has been entered as "
-        "'Not yet started'. Add a valid start date or change the status."
+        "You've entered a project start date that is before the end of the reporting period, but the project delivery "
+        "status has been entered as 'Not yet started'. Add a valid start date or change the status."
     )
 
 
--- a/data_store/validation/towns_fund/fund_specific_validation/fs_validate_r4.py
+++ b/data_store/validation/towns_fund/fund_specific_validation/fs_validate_r4.py
@@ -40,7 +40,9 @@ FUNDING_ALLOCATION = pd.read_csv(Path(__file__).parent / "resources" / "TF-grant
 
 
 def validate(
-    data_dict: dict[str, pd.DataFrame], original_workbook: dict[str, pd.DataFrame] | None = None
+    data_dict: dict[str, pd.DataFrame],
+    original_workbook: dict[str, pd.DataFrame],
+    reporting_round: int,
 ) -> list[GenericFailure]:
     """Top-level Towns Fund Round 4 specific validation.
 
--- a/data_store/validation/towns_fund/fund_specific_validation/fs_validate_r6.py
+++ b/data_store/validation/towns_fund/fund_specific_validation/fs_validate_r6.py
@@ -1,0 +1,49 @@
+from datetime import datetime
+
+import pandas as pd
+
+from data_store.const import StatusEnum
+from data_store.messaging.tf_messaging import TFMessages as msgs
+from data_store.validation.towns_fund.failures.user import GenericFailure
+from data_store.validation.towns_fund.fund_specific_validation.fs_validate_r4 import (
+    validate as validate_r4,
+)
+
+
+def validate(
+    data_dict: dict[str, pd.DataFrame], original_workbook: dict[str, pd.DataFrame] | None = None
+) -> list[GenericFailure]:
+    """Top-level Towns Fund Round 6 specific validation."""
+
+    # Start with validations from Round 4
+    validation_failures = validate_r4(data_dict, original_workbook)
+
+    # Add new validations specific to Round 6
+    project_progress_failures = validate_project_progress(data_dict)
+    validation_failures.extend(project_progress_failures)
+
+    return validation_failures
+
+
+def validate_project_progress(workbook: dict[str, pd.DataFrame]) -> list[GenericFailure]:
+    """Validates the Project Progress table for Round 6 submissions."""
+    project_progress_df = workbook["Project Progress"]
+    not_started_mask = project_progress_df["Project Delivery Status"] == StatusEnum.NOT_YET_STARTED
+    not_started_rows = project_progress_df[not_started_mask]
+    current_date = datetime.now().date()
+    failures = []
+    for idx, row in not_started_rows.iterrows():
+        start_date = row["Start Date"]
+        if pd.isnull(start_date):
+            continue
+        if start_date.date() <= current_date:
+            failures.append(
+                GenericFailure(
+                    table="Project Progress",
+                    section="Projects Progress Summary",
+                    column="Start Date",
+                    message=msgs.DATA_MISMATCH_PROJECT_START,
+                    row_index=idx,  # type: ignore[arg-type]
+                )
+            )
+    return failures
--- a/data_store/validation/towns_fund/fund_specific_validation/fs_validate_r6.py
+++ b/data_store/validation/towns_fund/fund_specific_validation/fs_validate_r6.py
@@ -1,9 +1,8 @@
-from datetime import datetime
-
 import pandas as pd
 
 from data_store.const import StatusEnum
 from data_store.messaging.tf_messaging import TFMessages as msgs
+from data_store.transformation.towns_fund import common
 from data_store.validation.towns_fund.failures.user import GenericFailure
 from data_store.validation.towns_fund.fund_specific_validation.fs_validate_r4 import (
     validate as validate_r4,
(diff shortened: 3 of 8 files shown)


pyproject.toml
```diff
@@ -104,6 +104,10 @@ disable_error_code = ["arg-type", "operator", "union-attr"]
 module = "data_store/validation/towns_fund/fund_specific_validation/fs_validate_r4"
 disable_error_code = ["arg-type", "assignment"]
 
+[[tool.mypy.overrides]]
+module = "data_store/validation/towns_fund/fund_specific_validation/fs_validate_r6"
+disable_error_code = ["arg-type"]
```
albertkol, line 109: Please don't disable the error.

Checking the error. You have one that needs fixing:
```
data_store/validation/towns_fund/fund_specific_validation/fs_validate_r6.py:55: error: Argument "row_index" to "GenericFailure" has incompatible type "Hashable | None"; expected "int | None"  [arg-type]
```

Only change needed is:
`                    row_index=idx,`
to
`                    row_index=tpying.cast(int, idx),` (`import typing` at the top)

row_index is very likely an int. `typing.cast` makes python treat it as a int.

samuelhwilliams: makes mypy consider it an int - doesn't change anything about how python treats it in code

data_store/validation/towns_fund/fund_specific_validation/fs_validate_r6.py
```diff
@@ -0,0 +1,59 @@
...
+)
+
+
+def validate(
+    data_dict: dict[str, pd.DataFrame], original_workbook: dict[str, pd.DataFrame] | None = None
+) -> list[GenericFailure]:
+    """Top-level Towns Fund Round 6 specific validation."""
+
+    # Start with validations from Round 4
+    validation_failures = validate_r4(data_dict, original_workbook)
```
MarcUsher, line 22: This will be tying together the R4 and R6 validation very closely and might make it difficult if we want to 'retire' R4 code at some point. I guess a similar concern as we had for R5, and how much we want to separate things out and make them each their own mini pipeline, especially when they're so similar (it seems silly to add it for one additional check).

This is more a general point in case it's something worth discussing as a team.

data_store/validation/towns_fund/fund_specific_validation/fs_validate_r6.py
```diff
@@ -0,0 +1,59 @@
...
+    validation_failures.extend(project_progress_failures)
+
+    return validation_failures
+
+
+def validate_project_progress(workbook: dict[str, pd.DataFrame]) -> list["GenericFailure"]:
+    """Validates the Project Progress table for Round 6 submissions."""
+
+    # First, get failures from Round 4 validations
+    failures = validate_project_progress_r4(workbook)
```
MarcUsher, line 35: If we're already calling `validate_r4()` in `validate()` above which itself is calling the original `validate_project_progress()`, do we need to call it again here when any `project_progress` validation failures will already be accounted for?

pyproject.toml
```diff
@@ -104,6 +104,10 @@ disable_error_code = ["arg-type", "operator", "union-attr"]
 module = "data_store/validation/towns_fund/fund_specific_validation/fs_validate_r4"
 disable_error_code = ["arg-type", "assignment"]
 
+[[tool.mypy.overrides]]
```
MarcUsher, line 107: Cheeky mypy disabling, @albertkol won't be happy! But maybe he's not got to fixing those particular type errors yet as it looks like the R4 file disables the same mypy check.

MarcUsher (commented):
Looks good! Just a couple of small comments. And like you mentioned to me we can approve this and actually implement the validation checks in the next PR to open the round. Do we want to add an R6 integration test just for safety, as otherwise that'll be 2 TF rounds with no proper integration tests (even though they're so similar to R4)?

pyproject.toml
samuelhwilliams: I think we should have a convo at the next tech chat around where that 'useful type hints' boundary lies. We should definitely be prioritising type hints where they add value/safety, not aiming for 100% coverage for the sake of it (especially around pandas where the type hints will get really gnarly).

albertkol: - `idx` is of type `Hashable | None`
- `GenericFailure` is accepting a `row_index` of type `int | None`
One of the two sides would need to be changed if we want to fix it properly.

`typing.cast` is indeed just treating the symptom. 
If you wish to ignore it, I would still rather suggest you inline ignore it. (`  # type: ignore[arg-type]`)

data_store/messaging/tf_messaging.py
```diff
@@ -28,6 +28,7 @@ class TFMessages(SharedMessages):
     MISSING_OTHER_FUNDING_SOURCES = (
         "You’ve not entered any Other Funding Sources. You must enter at least 1 over all projects."
     )
+    DATA_MISMATCH_PROJECT_START = "A Project has not yet started but has a Start Date prior to today declared."
```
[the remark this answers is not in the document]
wjrm500: Fair point, have raised this with Craig and Kate.

data_store/validation/towns_fund/fund_specific_validation/fs_validate_r6.py
wjrm500: Yeah very fair point, as we did this for R4 and R5 I don't want to rock the boat too much on this specific PR but definitely worth considering whether we want to refactor, and how far we want to isolate code between rounds etc.

pyproject.toml
albertkol: I changed my mind. I thought I could rid the code base of mypy exceptions. But at what cost?!

The file exception is fine, I left a few exceptions myself.

samuelhwilliams: Inline ignoring specific things that interact with pandas feels yeah OK (defo better to inline single exceptions rather than file ignore, unless there are loads)

data_store/messaging/tf_messaging.py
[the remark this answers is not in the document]
wjrm500: Haven't had any feedback on this yet so assuming they are happy with it

wjrm500:
@samuelhwilliams @MarcUsher - changes made and "How to test" instructions added to PR description whenever you're ready to re-review (not super urgent!)

data_store/messaging/tf_messaging.py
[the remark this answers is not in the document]
samuelhwilliams: Kate will not be happy with this so let's flag again

data_store/validation/towns_fund/fund_specific_validation/fs_validate_r6.py
```diff
@@ -0,0 +1,49 @@
...
+    project_progress_df = workbook["Project Progress"]
+    not_started_mask = project_progress_df["Project Delivery Status"] == StatusEnum.NOT_YET_STARTED
+    not_started_rows = project_progress_df[not_started_mask]
+    current_date = datetime.now().date()
+    failures = []
+    for idx, row in not_started_rows.iterrows():
+        start_date = row["Start Date"]
+        if pd.isnull(start_date):
+            continue
+        if start_date.date() <= current_date:
```
samuelhwilliams, line 39: bit of a pedantic one - should the "current date" be the last day of the reporting period? is the data submitted meant to be a snapshot of that moment in time?

I think this may be one to ask the fund team, but my assumption is that they should be reporting the project status as of the last day of the reporting period 🤔

samuelhwilliams (commented):
hm, hard to test this without the ground code required for TF round 6 - should we be adding that first (copy/paste of round 5)?

data_store/validation/towns_fund/fund_specific_validation/fs_validate_r6.py
wjrm500: Fair point and good catch - let's double check this with the TF team and Craig

albertkol: Oi.. go to sleep, m8!

wjrm500:
> hm, hard to test this without the ground code required for TF round 6 - should we be adding that first (copy/paste of round 5)?

Gaahh thought I put something about this in the test instructions in the PR description but apparently not, will add an additional instruction now:

- Create a copy of the TF R5 spreadsheet (funding-service-design-post-award-data-store/tests/integration_tests/mock_tf_returns/TF_Round_5_Success.xlsx)
- In the copy, change cell G21 on tab "3 - Programme Progress" to "Not yet started"
**_- In `funding-service-design-post-award-data-store/data_store/controllers/ingest_dependencies.py`, add an import `import data_store.validation.towns_fund.fund_specific_validation.fs_validate_r6 as tf_r6_validate` and for `case ("Towns Fund", 5)`, swap `fund_specific_validation=tf_r4_validate.validate` for `fund_specific_validation=tf_r6_validate.validate`_**
- Run data-store container locally and upload the changed spreadsheet copy at http://submit-monitoring-data.levellingup.gov.localhost:4001/upload/TF/5
- Note this row in error table:
Programme progress | Projects progress summary | D21 | A Project has not yet started but has a Start Date prior to today declared

data_store/messaging/tf_messaging.py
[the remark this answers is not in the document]
wjrm500: image

Have bumped the message

samuelhwilliams:
Thanks for explaining the workaround for enabling that extra bit of logic. I'll do that for now, but I guess it feels like we should still just add the TF R6 foundations.

wjrm500:
> Thanks for explaining the workaround for enabling that extra bit of logic. I'll do that for now, but I guess it feels in the future we should still add the TF R6 foundations before adding new logic for that round?

Sure, happy with that. My logic was that the TF R6 ingest dependency thing could form part of a later "round enabling PR" but given the need for a workaround here can see that it makes sense to swap the ordering in future.

data_store/validation/towns_fund/fund_specific_validation/fs_validate_r6.py
wjrm500: Have popped a message in the TF hotline

samuelhwilliams (approved):
Tested and works 👍

wjrm500:
Hello again @samuelhwilliams @MarcUsher, I've updated the code as per the new requirement, there is a separate commit so you can see what I changed since your last review. To clarify, this is a brief summary of the previous requirement:

_"If a project is marked with a delivery status of 'Not yet started' and the start date is less than or equal to the date on which the report is ingested then fail validation"_

And this is the new requirement:
_"If a project is marked with a delivery status of 'Not yet started' and the start date is less than or equal to the last day in the reporting period, then fail validation"_

Decided to use existing const to get observation period end date instead of `ReportingRound` entity as I tried this and it opened up a can of worms around database sessions and transactions - reckon we should tackle eliminating that const and adjusting code here and elsewhere in a separate PR.

data_store/validation/towns_fund/fund_specific_validation/fs_validate_r4.py
```diff
@@ -40,7 +40,9 @@
 
 
 def validate(
-    data_dict: dict[str, pd.DataFrame], original_workbook: dict[str, pd.DataFrame] | None = None
+    data_dict: dict[str, pd.DataFrame],
+    original_workbook: dict[str, pd.DataFrame] | None = None,
+    reporting_round: int | None = None,
```
samuelhwilliams, line 45: Can we not make this a required value?

PR #741, opened 2024-09-25
Run e2e tests by default when deploying; allow dev

samuelhwilliams:
### Change description
I find myself fairly often doing deploys of branches to dev and then manually triggering the e2e test suite. I think it's fine to just put these two together by default.

### DIFF
--- a/.github/workflows/aws_copilot_deployment.yml
+++ b/.github/workflows/aws_copilot_deployment.yml
@@ -12,9 +12,9 @@ on:
             type: string
             required: true
         run_e2e_tests:
-            description: "Run end-to-end (browser) tests; only works for the `test` environment"
+            description: "Run end-to-end (browser) tests; only works for the `dev` and `test` environments"
             type: boolean
-            default: false
+            default: true
             required: false
 
 jobs:
@@ -82,7 +82,7 @@ jobs:
 
   e2e_test:
     # Do not run these against the prod environment without addressing the auth/JWT self-signing done by e2e tests.
-    if: ${{ inputs.copilot_environment == 'test' && inputs.run_e2e_tests == true }}
+    if: ${{ inputs.run_e2e_tests == true && (inputs.copilot_environment == 'dev' || inputs.copilot_environment == 'test') }}
     name: Run end-to-end (browser) tests
     needs: [ migrate, deploy ]
     uses: ./.github/workflows/test_e2e.yml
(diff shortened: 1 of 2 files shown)


.github/workflows/aws_copilot_deployment.yml
```diff
@@ -12,9 +12,9 @@ on:
             type: string
             required: true
         run_e2e_tests:
-            description: "Run end-to-end (browser) tests; only works for the `test` environment"
+            description: "Run end-to-end (browser) tests; only works for the `dev` and `test` environments"
             type: boolean
-            default: false
+            default: true
```
albertkol, line 17: Step closer to accidentally run it on prod. :eyes:

.github/workflows/aws_copilot_deployment.yml
```diff
@@ -82,7 +82,7 @@ jobs:
 
   e2e_test:
     # Do not run these against the prod environment without addressing the auth/JWT self-signing done by e2e tests.
-    if: ${{ inputs.copilot_environment == 'test' && inputs.run_e2e_tests == true }}
+    if: ${{ inputs.run_e2e_tests == true && (inputs.copilot_environment == 'dev' || inputs.copilot_environment == 'test') }}
```
albertkol, line 85: Not sure if the syntax is correct. But happy to try it on prod. Looks plausible.

.github/workflows/deploy_combined_service.yml
```diff
@@ -9,6 +9,11 @@ on:
           required: true
           options:
           - dev
+        run_e2e_tests:
+          description: Run end-to-end tests?
+          type: boolean
+          default: true
+          required: true
```
albertkol, line 16: This adds a dropdown to choose if we want to run the e2e tests, right?

albertkol (approved):
:+1:

.github/workflows/deploy_combined_service.yml
samuelhwilliams: Yeah believe so

.github/workflows/aws_copilot_deployment.yml
samuelhwilliams: You could always read the docs and check ;)

albertkol: Good one.. :rofl:

.github/workflows/aws_copilot_deployment.yml
albertkol: Actually.. you should make set `run_e2e_tests` to `false` for the prod job. Just in case.

PR #742, opened 2024-09-25
FPASF-559: Enable check_untyped_defs rule

albertkol:
### Change description
The `check_untyped_defs` setting in mypy instructs mypy to type-check functions that don't have type annotations. By default, mypy skips checking the internal logic of functions that don't explicitly declare types (i.e., untyped function definitions). Enabling check_untyped_defs forces mypy to analyze these untyped functions as if they had implicit Any types.

In summary:
Enabled (True): mypy checks all functions, even those without type annotations.
Disabled (False): mypy skips type checking for functions without type annotations.

### How to test
`mypy` should pass
`pytest` should pass

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -35,7 +35,7 @@ warn_unused_ignores = true
 strict_equality = true
 
 # Strongly recommend enabling this one as soon as you can
-check_untyped_defs = false
+check_untyped_defs = true
 
 # These shouldn't be too much additional work, but may be tricky to
 # get passing if you use a lot of untyped libraries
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -90,3 +91,69 @@ disable_error_code = ["arg-type"]
 [[tool.mypy.overrides]]
 module = "data_store/validation/towns_fund/fund_specific_validation/fs_validate_r4"
 disable_error_code = ["arg-type"]
+
+
+# check_untyped_defs enabled exceptions
+# we should get rid of these one by one
```
samuelhwilliams, line 97: 🎉 yay for the incremental approach

samuelhwilliams (approved):

pyproject.toml
albertkol: Proceeds to fix all errors in the next PR: +1000 -1000 changes

PR #743, opened 2024-09-26
RFP-453 - Precursor - Migration to set output_data.start and outcome_data.start columns to nullable

wjrm500:
### Ticket
[Pathfinders - Proposed Amendments requested June '24](https://dluhcdigital.atlassian.net/browse/FPASF-453)

### Description
Precursor PR to prepare for Pathfinders round two ingestion. New requirements from Pathfinders team for round two include header changes to outputs and outcomes table which indicate that output and outcome start dates can actually be null (column header "Financial year 2023 to 2024" changed to e.g., "Total cumulative outputs to date").

This migration will be deployed, and the entities (SQLAlchemy change) afterwards to avoid race issues.

### DIFF
--- a/db/migrations/versions/045_nullable_output_outcome_star.py
+++ b/db/migrations/versions/045_nullable_output_outcome_star.py
@@ -1,0 +1,52 @@
+"""nullable output data and outcome data start date
+
+Revision ID: 045_nullable_output_outcome_star
+Revises: 044_remove_old_rr_cols
+Create Date: 2024-09-26 11:02:00.000000
+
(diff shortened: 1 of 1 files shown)


samuelhwilliams:
If we do this, there should be python model changes and `.current-alembic-head` also needs to be updated (indicating you've run this locally).

However I now wonder ... does `Total cumulative outputs to date` really say `start_date is None` - or does it say that `start_date` is day 1 of the pathfinders reporting period and we just need to calculate what that is?

PR #744, opened 2024-09-26
Make data-store use post /logout endpoint

albertkol:
### Change description
Bit of scope creeping that crept in the scope:
- Fixing Jinja indentation for submit/base, submit/login, find/base and find/login
- find/login was doing a weird thing so, I brought in line with what submit/login does
- css formatting: there is a standard for the correct css attributes order - had that accidentally on and was applied over the whole custom.css file

### How to test
- Compare submit/login and find/login they will look similar now
- Try the Sign out buttons, they should look the same as before and work the same as befor

After merge we should remove the GET /sign-out endpoint from Authenticator.

### DIFF
--- a/find/templates/find/base.html
+++ b/find/templates/find/base.html
@@ -21,40 +21,75 @@
 {% endblock head %}
 
 {% block header %}
-  {{ govukHeader({
-    'homepageUrl': url_for('find.download'),
-    'serviceUrl': url_for('find.download'),
-    'serviceName': config['FIND_SERVICE_NAME'],
-  }) }}
+  {% if g.user %}
+    {% set signOutButton %}
+      <form method="post" action="{{ config['AUTHENTICATOR_HOST'] + '/sso/logout' }}">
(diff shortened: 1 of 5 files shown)


find/templates/find/base.html
```diff
@@ -21,40 +21,75 @@
...
-    'serviceName': config['FIND_SERVICE_NAME'],
-  }) }}
+  {% if g.user %}
+    {% set signOutButton %}
+      <form method="post" action="{{ config['AUTHENTICATOR_HOST'] + '/sso/logout' }}">
+        <input type="hidden" name="csrf_token" value="{{ csrf_token() }}">
+        <button type="submit" class="button-as-link">
+            <a class="govuk-header__link">Sign out</a>
+        </button>
+      </form>
```
samuelhwilliams, line 31: let's chat.

samuelhwilliams:
I think we think this is a no-goer for now - shall we close or mark as draft?

albertkol:
I am closing. 

Waiting for Gideon to return to have a talk about it. Before restarting work on it.

PR #745, opened 2024-09-26
Pathfinders bugfix - guard against duplicate outputs and outcomes within tables and across tables

wjrm500:
This is a fix for a pre-existing problem with the code that basically meant that if a user entered the same output name more than once, even across both the "Standard outputs" and "Bespoke outputs" tables (as in, the user enters the same output in both tables), AND there is no matching output name in the database table `output_dim`, we get an uncaught database integrity error causing a 500 response, because of a lack of unique validation in the application and a unique constraint on `output_dim.output_name`. The same thing applies for outcomes.

Although it appears that at least one LA entered a duplicate output name for Pathfinders R1 (based on the fact that the test spreadsheet that Craig sent to me a couple days ago, which I believe contained the actual data as it was entered by one of the LAs, contained a duplicate output name), my theory is that we didn't see any error then because the submission containing the duplicated output name was by chance submitted after a submission containing the same output name, but not duplicated across tables, which meant that the output name was inserted into `output_dim` okay. And then when the submission with duplicated output names was submitted afterwards, and `funding-service-design-post-award-data-store/data_store/controllers/load_functions.py::get_outcomes_outputs_to_insert` was run to filter out existing outputs, the duplicated output name was filtered out and the application never attempted to insert it into `output_dim`, so no error was thrown. Again, the same thing applies for outcomes.

The code in this PR applies in-table and cross-table checks to ensure all outputs and outcomes entered are unique. It ensures that we never reach a stage where a database integrity error is possible.

This branch was checked out from `RFP-453_pathfinders-round-2` as it contains R2-specific code. It is not _essential_ to merge this - as I mention above, we got lucky in R1 and we now likely have most (if not all, I haven't checked) of the available output and outcome names already in the `output_dim` and `outcome_dim` tables, which means an error is even less likely this time around. Merging this may also have unintended consequences in terms of preventing users from submitting data in the way they expect (i.e., do we actually _know_ that having the same output name in both standard and bespoke outputs is wrong?). However, I wanted to leave this PR here to give the option if we do decide this problem needs fixing, and wanted to avoid database load changes / migrations.

### DIFF
--- a/data_store/table_extraction/config/pf_r2_config.py
+++ b/data_store/table_extraction/config/pf_r2_config.py
@@ -210,7 +210,7 @@ PF_TABLE_CONFIG: dict[str, dict[str, dict]] = {
         "validate": {
             "columns": {
                 "Intervention theme": string_column(),
-                "Output": string_column(),
+                "Output": string_column(unique=True, report_duplicates="exclude_first"),
                 "Unit of measurement": string_column(),
                 "Total cumulative outputs to date, (Up to and including Mar 2024), Actual": float_column(),
                 "Financial year 2024 to 2025, (Apr to Jun), Actual": float_column(),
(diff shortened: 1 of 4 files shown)


wjrm500:
Manual validation of the 10 spreadsheets sent out to LAs for PF R2 showed no duplicate outputs or outcomes with locked cells, so no urgency on this fix going out and can think of a longer-term solution based on stronger understanding of outputs / outcomes / intervention themes and their relationships.

wjrm500:
Thought I already closed this but apparently not - this has been superseded by [this PR](https://github.com/communitiesuk/funding-service-design-post-award-data-store/pull/766).

PR #746, opened 2024-09-27
Update start date for pathfinders r2

samuelhwilliams:
### Change description
As part of the changes for Pathfinders Round 2, we've updated the headers for some of the tables to include "everything so far up to march 2024". The start date for these was previously unclear but Pathfinders team have recently clarified that it should be considered 2019-01-01, so let's update.


MarcUsher (approved):
Approving for when that e2e test

PR #748, opened 2024-09-27
Update e2e submit(pathfinders) test for round 2

samuelhwilliams:


tests/e2e_tests/helpers.py
```diff
@@ -84,7 +84,7 @@ def lookup_confirmation_emails(
     client = NotificationsAPIClient(e2e_test_secrets.NOTIFY_SUBMIT_API_KEY)
 
     la_email_subject = "Your Pathfinders data return has been submitted"
-    fund_email_subject = "Record of a Pathfinders submission for Pathfinders Bolton Council"
+    fund_email_subject = "Record of a Pathfinders submission for Pathfinders Test Council"
```
MarcUsher, line 87: This just so you can easily know it's the e2e tests running?

MarcUsher (approved):
👍

tests/e2e_tests/helpers.py
samuelhwilliams: It is mostly because the r2 spreadsheets that were added before are populated by Test Council info, so that's how it shows up

PR #749, opened 2024-09-30
Add flask-debugtoolbar for local development

samuelhwilliams:
It isn't currently possible to inject CSP nonces into Flask-Debugtoolbar
so we need to allow its inline scripts and styles by sha256 hash
directly.

Will try to get this fixed upstream, but low risk as these are
restricted to local development only.


samuelhwilliams:
pre-commit.ci run

samuelhwilliams:
yeah - need to do an update on flask-debugtoolbar to add CSP support, but its meh

PR #758, opened 2024-10-01
Fix e2e tests when running locally

samuelhwilliams:
When flask-debugtoolbar is enabled locally, it injects a lot of HTML onto the page to
help explain what's happened when processing the request. This has
subtly broken our e2e tests (but only when running locally with
flask-debugtoolbar enabled).

The easy part of the fix is to tweak our selectors for key elements
during the find and submit tests. When selecting a radio button for file
type, specifically select the input label. When clicking a link for
Pathfinders, specifically look for a link.

We also turn off flask-debugtoolbar redirect interception by default, so that e2e tests
don't have to deal with that.

---

Finally, the e2e tests flagged some more CSP hashes we "need" to add for a couple of obscure flask-debugtoolbar inline styles 🙃


albertkol (approved):
GG WP

PR #760, opened 2024-10-02
Pass ingest params as function args, not a body dict

samuelhwilliams:
Start cleaning up some of the code around ingest after the app consolidation work. The `ingest` function still acts like an HTTP endpoint, returning JSON around status codes etc instead of just raising errors.

We start to tackle that here by removing the arbitrary `body` dict it receives and making these individual typed arguments instead, so that the function signature is clearer.

To come in future patches:
* Replacing the `status_code` return values of the function with just raised errors of some kind
* Removing the `post_ingest` middleware in the `submit` module.


wjrm500 (commented):
Looking good, only thing I spotted is that `parse_auth` is still sitting pretty at the bottom of `ingest.py` and should be removed, and any tests that use it should be removed / modified.

samuelhwilliams:
Ah good spot. I did notice at some point that it was being orphaned and had it in the back of my head to clean up - but didn't do it for some reason. Will take it out now - thanks 👍

wjrm500 (approved):

PR #762, opened 2024-10-03
FPASF-599 - Pre-emptive fix of issue with validating outputs and outcomes during PF R2 ingestion

wjrm500:
### Ticket

[Fix PF R2 outputs / outcomes validation issue](https://dluhcdigital.atlassian.net/browse/FPASF-599)

### Description

In this change we remove the cross-table validation checks on the outputs, outcomes, bespoke outputs and bespoke outcomes tables for the PF R2 ingest pipeline. These checks ensure that (A) all outputs and outcomes are associated with an expected intervention theme and unit of measurement, based on the control tables in the hidden worksheet 'Dropdown Values', and (B) all bespoke outputs and outcomes are expected for the local authority submitting the report, based on the control tables in the hidden worksheets 'Bespoke Outputs' and 'Bespoke Outcomes'. They were in place during R1 ingest as 'tamper checks', given that users could in theory manipulate the spreadsheet to override spreadsheet data validation rules.

However, for R2 ingest these checks are likely to cause problems, owing to the fact that during spreadsheet creation, data was copied and pasted into the outputs, outcomes, bespoke outputs and bespoke outcomes tables from an unverified source, severing the connection between the control tables in the hidden worksheets and the user-facing worksheets for these tables and meaning that these checks can no longer be safely assumed to pass. Simply removing these checks from R2 is seen as okay because (A) the relevant columns in the spreadsheet tables are now locked, whereas they were not previously, and (B) identifying and hard-coding expected values as per the actual spreadsheets that have already gone out would be labour-intensive and messy.

As part of this PR, I've also added some additional errors the `cross_table_validation_fails` test, as it was looking very bare after removing these checks (it only had one error left!).


wjrm500:
Closing as superseded by PR that retains modified versions of the checks.

PR #763, opened 2024-10-03
Add SubmissionDate to download's SubmissionRef section

sc15zs:
Ticket:

Description:
This is the first step of the work in the ticket, to add "SubmissionDate" to the "SubmissionRef" section of the data extract.

Screenshot:
Screenshot 2024-10-04 at 09 18 09

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -290,6 +290,7 @@ def additional_test_data() -> dict[str, Any]:
     submission = Submission(
         submission_id="TEST-SUBMISSION-ID",
         reporting_round=reporting_round,
+        submission_date=datetime(2024, 10, 1),
     )
 
     organisation = Organisation(organisation_name="TEST-ORGANISATION")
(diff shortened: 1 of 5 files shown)


tests/conftest.py
```diff
@@ -290,6 +290,7 @@ def additional_test_data() -> dict[str, Any]:
     submission = Submission(
         submission_id="TEST-SUBMISSION-ID",
         reporting_round=reporting_round,
+        submission_date=datetime.now(),
```
samuelhwilliams, line 293: It's generally not a good idea to use test data that can vary on a test-by-test basis, because it leads to complexity like that you've encountered and tried to fix by rounding to the nearest minute (I think).

If we use a static datetime here, does a lot of that complexity and frailty disappear?

samuelhwilliams: A random (but set in stone) one - yes :)

sc15zs: Changed it to a static datetime.

samuelhwilliams (approved):

samuelhwilliams:
Have chased TF again to understand if this will break them

samuelhwilliams:
@sc15zs - TF team confirmed this is OK to merge in :)

PR #767, opened 2024-10-04
FPASF-599 - PF R2 - SQL to change output names to account for text changes between round 1 and round 2 of Pathfinders reporting

wjrm500:
### Ticket

[Fix PF R2 outputs / outcomes validation issue](https://dluhcdigital.atlassian.net/browse/FPASF-599)

### Description

This PR contains an SQL script that is to be run against the production database, to update the names of the outputs that were renamed between Pathfinders rounds 1 and 2. By running these updates, in conjunction with the code in the other PR for this ticket (see ticket comments), we ensure that the same real-world output maps to a single output in the database, despite the name change.


samuelhwilliams (approved):
Looks sensible to me - what's your plan for running these?

PR #768, opened 2024-10-04
Upgrade utils version; pinning greenlet due to conflict

albertkol:
### Change description
Upgrade fsd utils version and pin greenlet version due to conflict.

### How to test
Set `FSD_UTILS_ENABLE_CONFIG_TABLE` in `.env` docker runner.

`data-store` will no longer show the table.

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -36,7 +36,9 @@ blinker==1.8.2
     #   flask
     #   sentry-sdk
 boto3==1.34.151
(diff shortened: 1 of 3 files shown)


albertkol:
Not too sure if pinning he `greenlet` version was the right approach to fix the dep conflict. If there is a better way, happy to change it.

samuelhwilliams (approved):

PR #769, opened 2024-10-07
Bugfix - ReportingRound admin - observation period start was taking the value from observation period end erroneously

wjrm500:


samuelhwilliams (approved):
good spot 👀

PR #771, opened 2024-10-08
BAU - reifying Pathfinders schemas for extraction, processing and validation

wjrm500:
For Pathfinders, we use a single schema to configure extraction of 'tables' from spreadsheet worksheets, processing of those tables to create clean, labelled DataFrames, and validation of those cleaned DataFrames using Pandera. Currently, the schema is a plain Python object. This change introduces new classes for representing the objects and sub-objects that form the schema, and uses those new classes in the schema.

### DIFF
--- a/data_store/table_extraction/process.py
+++ b/data_store/table_extraction/process.py
@@ -2,6 +2,7 @@ import numpy as np
 import pandas as pd
 from pandas import DataFrame, Index
 
+from data_store.table_extraction.config.common import ProcessConfig
 from data_store.table_extraction.exceptions import TableProcessingError
 from data_store.table_extraction.table import Table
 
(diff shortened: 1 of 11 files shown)


data_store/table_extraction/process.py
```diff
@@ -41,34 +42,19 @@ class TableProcessor:
...
-        self.drop_empty_rows = drop_empty_rows
-        self.drop_empty_tables = drop_empty_tables
-        self.dropdown_placeholder = dropdown_placeholder
+        self.num_header_rows = process_config.num_header_rows
+        self.ignored_non_header_rows = process_config.ignored_non_header_rows
+        self.merged_header_rows = process_config.merged_header_rows
+        self.col_names_to_drop = process_config.col_names_to_drop
+        self.drop_empty_rows = process_config.drop_empty_rows
+        self.drop_empty_tables = process_config.drop_empty_tables
+        self.dropdown_placeholder = process_config.dropdown_placeholder
```
samuelhwilliams, line 57: nit: it maybe feels a bit weird to have a container for config in `ProcesConfig` but then just dupe all of the attributes - should we just store the process_config object and redirect through that?

It's probably fine for now and we could address in the future if needed.

data_store/validation/pathfinders/schema_validation/validate.py
```diff
@@ -61,17 +61,19 @@ class TableValidator:
...
+    def __init__(self, validate_config: ValidateConfig) -> None:
         """Initialises a TableValidator.
 
-        :param schema_config: A dictionary containing schema configuration.
+        :param validate_config: The configuration for the schema to validate against.
             The schema configuration should include information about columns, data types,
             and any additional validation rules.
-        :raises ValueError: If the schema has coercion enabled.
         """
-        if schema_config.get("coerce") or any(column.coerce for column in schema_config.get("columns", {}).values()):
```
samuelhwilliams, line 72: Do we still need the latter conditional checking for `coerce` on the columns? It's not available on the schema_config itself now but is it still avaialble on the columns?

samuelhwilliams (approved):
+1k/-1k - classic Will PR 🙌 

Looks alright to me 👀

PR #772, opened 2024-10-08
Fix r6 last column bug

albertkol:
### Change description
**Problem:** The last column from Sheet 7 was removed. And we used to select the Risk table until column `-1`, second last column, as a result we would no longer select the Column P, Risk Owner column.

Change so we don't go till column `-1`, but rather hard code it to the exact number, `16`.

### How to test
Prod testing. Also tests pass! Not that it matters, since the sheet we test with, is different from the sheet the funds use. ¯\_(ツ)_/¯


data_store/transformation/towns_fund/tf_transform_r3.py
```diff
@@ -716,7 +716,7 @@ def extract_risks(
     :param programme_id: ID of the programme for this ingest
     :return: A new DataFrame containing the extracted programme/risk rows.
     """
-    df_risk_programme = df_risk.iloc[9:14, 2:-1]
+    df_risk_programme = df_risk.iloc[9:14, 2:16]
```
samuelhwilliams, line 719: If this is only changing for round 6 then consider that this would break reingest for round 3, 4, and 5?

albertkol: Surely the table has the same length... right?!

samuelhwilliams: You tell me please 🙏  (ideally with evidence)

albertkol: Checked the spreadsheets and the Risk table all stretch up to column P. Also tests passed for those rounds too.

albertkol: Yeh. Maybe I was too cocky about my fix. 

Updating the tests revealed I missed updating a function. :disappointed:

albertkol:
I saved the xlsx file with LibreOffice, I hope that does not cause issues.

samuelhwilliams (approved):
Thanks for walking me through this 👍  This sounds like a sensible approach

PR #774, opened 2024-10-08
Adding bucket name to the sanitised db restore script

israr-ulhaq:
Removing argument from the script and adding the bucket name directly to the restore script.


gidsg (approved):
LGTM 👍

PR #776, opened 2024-10-09
Ensure TF schema validation on project start and end dates handles non-datetime input

wjrm500:
Bug noticed by Craig when he was playing about with rogue input on test. If we have a non-null, non-datetime input in "Start Date" column of "Project Progress" table in TF spreadsheet, we fail completely at the line `if start_date > completion_date` in `funding-service-design-post-award-data-store/data_store/validation/towns_fund/schema_validation/validate.py::validate_project_dates`, with `'>' not supported between instances of 'WRONG_TYPE' and 'datetime.datetime'`.

### DIFF
--- a/data_store/validation/towns_fund/fund_specific_validation/fs_validate_r6.py
+++ b/data_store/validation/towns_fund/fund_specific_validation/fs_validate_r6.py
@@ -1,3 +1,5 @@
+from datetime import datetime
+
 import pandas as pd
 
 from data_store.const import StatusEnum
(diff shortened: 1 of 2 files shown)


samuelhwilliams:
I feel like we clocked this can potentially happen in a bunch of places at some point before 🤔

wjrm500:
> I feel like we clocked this can potentially happen in a bunch of places at some point before 🤔 Unsure if it's valuable to pick them off one at a time, or if we should do some kind of audit

Always worth an audit but I would say good to get this fix out regardless? I guess it depends on how you scope the issue to say whether there are necessarily fixes that could / should be done elsewhere simultaneously, but this feels like a valuable fix in its own right given that `validate_project_dates` is a shared validation function and the only one that concerns itself with dates

samuelhwilliams (approved):
i won't say "test?" ;)

PR #778, opened 2024-10-10
Update test spreadsheet to match prod

albertkol:
### Change description
Update test spreadsheets for R6. Update the tests too to match different LA.

### How to test
Are tests passing?


samuelhwilliams (approved):
I trust that those files are valid and not full of malware.

albertkol:
Linux got not malware, m8!

PR #779, opened 2024-10-11
Making ingest_date column nullable on Submission model

sc15zs:
Ticket:

Description:
As a first step of dropping the ingest_date column from the Submission, this PR makes the column nullable.


data_store/db/entities.py
```diff
@@ -606,7 +606,7 @@ class Submission(BaseModel):
     reporting_round_id: Mapped[GUID] = sqla.orm.mapped_column(sqla.ForeignKey("reporting_round.id"), nullable=False)
 
     submission_date = sqla.Column(sqla.DateTime(), nullable=True)
```
samuelhwilliams, line 608: Should we make this non-nullable at the same time? I've just checked prod and there are 0 records without a submission date - so this feels like a good time for us to start enforcing that?

There are a few tests that will fail because of this, but I think it should just be adding `submission_date` values for them.

samuelhwilliams:
One comment, otherwise looks all good and I think should be fine to get in :) Nice start.

data_store/db/entities.py
sc15zs: Changed submission_date to be non-nullable.

data_store/db/entities.py
```diff
@@ -605,8 +605,8 @@ class Submission(BaseModel):
     submission_id = sqla.Column(sqla.String(), nullable=False, unique=True)
     reporting_round_id: Mapped[GUID] = sqla.orm.mapped_column(sqla.ForeignKey("reporting_round.id"), nullable=False)
 
-    submission_date = sqla.Column(sqla.DateTime(), nullable=True)
-    ingest_date = sqla.Column(sqla.DateTime(), nullable=False, default=datetime.now())
+    submission_date = sqla.Column(sqla.DateTime(), nullable=False, default=datetime.now())
```
samuelhwilliams, line 608: Be careful - this is copying the same mistake that `ingest_date` was making.

I would suggest it's broadly OK to not set a default time, but if you do, it needs to be `datetime.now` - specifically _not_ called. I think it's safer to not have this default as the mistake is easily made.

samuelhwilliams (changes requested):

samuelhwilliams (dismissed):

tests/integration_tests/test_ingest_component_all_funds.py
```diff
@@ -128,7 +128,17 @@ def test_multiple_rounds_multiple_funds_end_to_end(
...
+    for expected, actual in zip(
+        expected_submission_ref["SubmissionDate"], df_dict["SubmissionRef"]["SubmissionDate"], strict=False
+    ):
+        assert abs((expected - actual).total_seconds()) < 60, f"SubmissionDates don't match: {expected} != {actual}"
+
+    # Check other columns
+    assert_frame_equal(
+        expected_submission_ref.drop(columns=["SubmissionDate"]),
+        df_dict["SubmissionRef"].drop(columns=["SubmissionDate"]),
+    )
```
samuelhwilliams, line 141: Please can you refer to Will's PR which puts a proper fix for this test in place :)

sc15zs: Does it mean to wait until his branch is merged in to main and then rebase my branch on main where this test is fixed, or to install freezegun and then Will to revert 90% of his PR?

samuelhwilliams: Will's PR is a better fix for the broken test than the implementation you've currently provided, IMO.

You can either review Will's PR and get it merged, or you could close his PR, cherry-pick the commit onto your branch and drop this slightly-hacky fix.

I think either is fine.

sc15zs: Done, thank you!

samuelhwilliams (approved):

PR #781, opened 2024-10-14
Updating sanitisation script for new table

israr-ulhaq:
This change will do the following

- Sanitising question and indicator data_blob keys in the funding_question column.
- Sanitising by applying lorem_ipsum to  commentary, important_milestone and leading_factor_of_delay  in the the project_progress column.
- Binding psql version 16 to avoid the mismatch error.
- Adding command to flask-cli to get sanitised db directly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -154,6 +154,15 @@ Seeds the database with sample data.
 flask db-data seed-sample-data
 ```
 
+#### seed-sanitised-db
+
+Seeds the local database with Sanitised data. Make sure docker is up.
+
+
+```python
+flask db-data seed-sanitised-db
+```
+
 #### reset
 
 Reset the database by dropping all data and reseeding the geospatial and fund reference data.
--- a/scripts/restore-sanitised-database.sh
+++ b/scripts/restore-sanitised-database.sh
@@ -6,6 +6,8 @@ LOCAL_DB_PASSWORD="password"
 LOCAL_DB_USER="postgres"
 LOCAL_DB_NAME="data_store"
 DB_IMAGE="postgres:16.2"
+RESTORE_DB=${1}
+
 
 # Function to list and select an AWS Vault profile
 select_aws_vault_profile() {
@@ -41,8 +43,9 @@ echo "Downloading the backup file from S3..."
 aws-vault exec $AWS_VAULT_PROFILE -- aws s3 cp s3://$S3_BUCKET/$SANITISED_DB_PATH /tmp/$SANITISED_DB_PATH || { echo "Failed to download the backup file"; exit 1; }
 echo "${SANITISED_DB_PATH} file is ready to restore"
 
-
-read -p "Do you want to restore the database locally? (y/n): " RESTORE_DB
+if [ -z "$RESTORE_DB" ]; then
+    read -p "Do you want to restore the database locally? (y/n): " RESTORE_DB
+fi
 
 if [[ "$RESTORE_DB" == "y" || "$RESTORE_DB" == "Y" ]]; then
 
--- a/scripts/sanitise-db.sql
+++ b/scripts/sanitise-db.sql
@@ -8,19 +8,24 @@ DECLARE
     query TEXT;
     count_query TEXT;
     pre_update_count INTEGER;
+    table_specific_condition TEXT := '';
 BEGIN
     RAISE NOTICE 'Anonymising table: %, data_blob key: %, with lorem ipsum.', UPPER(table_name), UPPER(data_blob_key);
 
+    IF table_name = 'project_progress' THEN
+        table_specific_condition := ' AND start_date < end_date';  -- Apply this condition only for the 'project_progress' table
+END IF;
+
     count_query := format(
-        'SELECT COUNT(*) FROM %I WHERE %I->>%L IS NOT NULL',
-        table_name, column_name, data_blob_key
+        'SELECT COUNT(*) FROM %I WHERE %I->>%L IS NOT NULL %s',
+        table_name, column_name, data_blob_key, table_specific_condition
     );
     EXECUTE count_query INTO pre_update_count;
     RAISE NOTICE 'Number of rows to be updated: %', pre_update_count;
 
     query := format(
-        'UPDATE %I SET %I = jsonb_set(%I, %L, to_jsonb(anon.lorem_ipsum(characters := length(%I->>%L)))) WHERE %I->>%L IS NOT NULL',
-        table_name, column_name, column_name, format('{%s}', data_blob_key), column_name, data_blob_key, column_name, data_blob_key
+        'UPDATE %I SET %I = jsonb_set(%I, %L, to_jsonb(anon.lorem_ipsum(characters := length(%I->>%L)))) WHERE %I->>%L IS NOT NULL %s',
+        table_name, column_name, column_name, format('{%s}', data_blob_key), column_name, data_blob_key, column_name, data_blob_key, table_specific_condition
     );
     EXECUTE query;
 
@@ -150,12 +155,16 @@ DECLARE
     query TEXT;
     count_query TEXT;
     pre_update_count INTEGER;
+    table_specific_condition TEXT := '';
 BEGIN
     RAISE NOTICE 'Anonymising table: %, data_blob key: %, with random integer data adjusted by -% and +% percentages.', UPPER(table_name), UPPER(data_blob_key), negative_percentage, positive_percentage;
 
+    IF table_name = 'project_progress' THEN
+        table_specific_condition := ' AND start_date < end_date';
+END IF;
     count_query := format(
-        'SELECT COUNT(*) FROM %I WHERE data_blob->>%L IS NOT NULL ',
-        table_name, data_blob_key
+        'SELECT COUNT(*) FROM %I WHERE data_blob->>%L IS NOT NULL %s',
+        table_name, data_blob_key, table_specific_condition
     );
 
     EXECUTE count_query INTO pre_update_count;
@@ -169,9 +178,8 @@ BEGIN
                 (data_blob->>%L)::INTEGER +
                 FLOOR((data_blob->>%L)::INTEGER * ((random() * (%s - %s)) + %s) / 100)
             )
-
-        ) WHERE data_blob->>%L ~ ''^[0-9]+$''',
-        table_name, '{' || data_blob_key || '}', data_blob_key, data_blob_key, positive_percentage, negative_percentage, negative_percentage, data_blob_key
+        ) WHERE data_blob->>%L ~ ''^[0-9]+$'' %s',
+        table_name, '{' || data_blob_key || '}', data_blob_key, data_blob_key, positive_percentage, negative_percentage, negative_percentage, data_blob_key, table_specific_condition
     );
     EXECUTE query;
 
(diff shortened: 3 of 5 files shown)


scripts/restore-sanitised-database.sh
```diff
@@ -6,6 +6,8 @@ LOCAL_DB_PASSWORD="password"
 LOCAL_DB_USER="postgres"
 LOCAL_DB_NAME="data_store"
 DB_IMAGE="postgres:16.2"
+RESTORE_DB=${1}
```
wjrm500, line 9: This parameter is kind of confusing for a couple of reasons:

- The name `RESTORE_DB` does not match the true meaning of the variable, which is more `RESTORE_TO_LOCAL_DB`
- The parameter is not a named parameter meaning that usage e.g., `scripts/restore-sanitised-database.sh y` is not intuitive at all
- Separate point that goes beyond this PR, but if the user selects no they do not want to restore to their local database, the default is to sanitise in the cloud which seems like an odd thing to default to, as this is the more dangerous action. The user should be made to explicitly specify that they want to sanitise the DB on AWS

scripts/sanitise-db.sql
```diff
@@ -339,9 +361,17 @@ DO $$
 BEGIN
     PERFORM anonymize_with_lorem_ipsum('funding', 'data_blob', 'funding_source');   --- LARGE TABLE
     PERFORM anonymize_with_lorem_ipsum('funding_comment', 'data_blob', 'comment');
+    PERFORM anonymize_with_lorem_ipsum('funding_question', 'data_blob', 'question');
+    PERFORM anonymize_with_lorem_ipsum('funding_question', 'data_blob', 'indicator');
     PERFORM anonymize_with_lorem_ipsum('private_investment', 'data_blob', 'additional_comments');
     PERFORM anonymize_with_lorem_ipsum('programme_progress', 'data_blob', 'answer');
 
+    PERFORM anonymize_with_lorem_ipsum('project_progress', 'data_blob', 'commentary');
```
wjrm500, line 369: Could you confirm why these specific fields are being sanitised and not other fields in the `project_progress` data blob including:

- delivery_stage
- adjustment_request_status
- delivery_rag
- project_status

scripts/sanitise-db.sql
```diff
@@ -339,9 +361,17 @@ DO $$
 BEGIN
     PERFORM anonymize_with_lorem_ipsum('funding', 'data_blob', 'funding_source');   --- LARGE TABLE
     PERFORM anonymize_with_lorem_ipsum('funding_comment', 'data_blob', 'comment');
+    PERFORM anonymize_with_lorem_ipsum('funding_question', 'data_blob', 'question');
```
wjrm500, line 364: Could you confirm why these specific fields are being sanitised and not other fields in the `funding_question` data blob including:

- response
- guidance_notes

scripts/sanitise-db.sql
```diff
@@ -260,30 +268,44 @@ CREATE OR REPLACE FUNCTION anonymize_risk_register_table() RETURNS VOID AS $$
...
-                'short_desc', substring(lorem_text from 10 for 16),
-                'mitigations', substring(lorem_text from 1 for 10),
-                'consequences', substring(lorem_text from 1 for 5),
-                'risk_owner_role', substring(lorem_text from 1 for 3)
-            )
-        )
-        WHERE data_blob IS NOT NULL;
+BEGIN
+UPDATE risk_register
+SET data_blob = jsonb_set(
```
wjrm500, line 292: So the idea here is to only update specific fields, leaving unnamed fields unsanitised? Defaulting to unsanitised might not be the best approach in terms of future-proofing the effectiveness of this sanitisation process, as any new fields with PII that are added to the database but are not added here will be visible as-is in the supposedly sanitised output.

If we're okay with this, then a more concise approach could involve using [JSON object concatenation](https://stackoverflow.com/questions/30101603/merging-concatenating-jsonb-columns-in-query) and in-lining the calls to `anon.lorem_ipsum` when building the new JSON object to concatenate the existing one with, e.g.,:

```
new_values := jsonb_build_object(
   'full_desc', substring(anon.lorem_ipsum(words := 50) from 1 for 200),
   'risk_name', substring(anon.lorem_ipsum(words := 10) from 1 for 50),
   'short_desc', substring(anon.lorem_ipsum(words := 20) from 1 for 100),
   'mitigations', substring(anon.lorem_ipsum(words := 30) from 1 for 150),
   'consequences', substring(anon.lorem_ipsum(words := 30) from 1 for 150),
   'risk_owner_role', substring(anon.lorem_ipsum(words := 5) from 1 for 50)
);
```

And then:
```
UPDATE risk_register
SET data_blob = data_blob || new_values
WHERE data_blob IS NOT NULL;
```

scripts/sanitise-db.sql
```diff
@@ -8,19 +8,24 @@ DECLARE
     query TEXT;
     count_query TEXT;
     pre_update_count INTEGER;
+    table_specific_condition TEXT := '';
 BEGIN
     RAISE NOTICE 'Anonymising table: %, data_blob key: %, with lorem ipsum.', UPPER(table_name), UPPER(data_blob_key);
 
+    IF table_name = 'project_progress' THEN
```
wjrm500, line 15: What is the benefit of deliberately not sanitising `project_progress` records where the start date is less than the end date?

scripts/sanitise-post-award-database.sh
```diff
@@ -20,6 +20,20 @@ if [[ $1 == "--help" || $1 == "-h" ]]; then
...
+    echo "psql is not installed. Please install it first."
+    exit 1
+fi
+
+version=$(psql --version | awk '{print $3}' | cut -d. -f1)
+
+echo "psql version you have is: ${version}"
+
+if [ "$version" -ne 16 ]; then
+    echo "You need psql (PostgresSQL) version 16 to run this script."
```
wjrm500, line 33: PostgreSQL is the correct spelling

README.md
```diff
@@ -154,6 +154,15 @@ Seeds the database with sample data.
 flask db-data seed-sample-data
 ```
 
+#### seed-sanitised-db
+
+Seeds the local database with Sanitised data. Make sure docker is up.
```
wjrm500, line 159: Lower-case "Sanitised" and title-case "docker" for correctness

wjrm500 (changes requested):
Hi @israr-ulhaq, interesting changes to review - I enjoyed taking a closer look at the database sanitisation stuff you've been working on, looks like a really challenging problem that you've made a lot of headway on, so kudos!

Regarding this PR, I'm slightly confused what the overall objective is as there is no linked Jira ticket in the description and the name "Updating sanitisation script for new table" doesn't seem to match the code I've just reviewed. For example, what exactly is the new table referred to in the PR name? In this PR I can see a variety of changes, including:

1. Adding anonymisation to additional (but not all) columns in the `funding_question` and `project_progress` tables
2. Deliberately not sanitising certain records in the `project_progress` table
3. An update to the way we handle anonymising the `risk_register` data blob, to generate new lorem ipsum for each field and avoid deleting unnamed fields
4. A new CLI command to abstract away direct usage of the `restore-sanitised-database` script, and a new argument for that script to bypass user input

All are potentially useful (although need clarification on #2), _but_ I'm not sure they should all be hanging out together in the same PR

Interested to get your thoughts on the comments I've left - merge is blocked for now pending your clarification on certain points. Good work anyway :)

israr-ulhaq:
> Hi @israr-ulhaq, interesting changes to review - I enjoyed taking a closer look at the database sanitisation stuff you've been working on, looks like a really challenging problem that you've made a lot of headway on, so kudos!
> 
> Regarding this PR, I'm slightly confused what the overall objective is as there is no linked Jira ticket in the description and the name "Updating sanitisation script for new table" doesn't seem to match the code I've just reviewed. For example, what exactly is the new table referred to in the PR name? In this PR I can see a variety of changes, including:
> 
> 1. Adding anonymisation to additional (but not all) columns in the `funding_question` and `project_progress` tables
> 2. Deliberately not sanitising certain records in the `project_progress` table
> 3. An update to the way we handle anonymising the `risk_register` data blob, to generate new lorem ipsum for each field and avoid deleting unnamed fields
> 4. A new CLI command to abstract away direct usage of the `restore-sanitised-database` script, and a new argument for that script to bypass user input
> 
> All are potentially useful (although need clarification on no. 2), _but_ I'm not sure they should all be hanging out together in the same PR?
> 
> Interested to get your thoughts on the comments I've left - merge is blocked for now pending your clarification on certain points. Good work anyway :)

israr-ulhaq:
> Hi @israr-ulhaq, interesting changes to review - I enjoyed taking a closer look at the database sanitisation stuff you've been working on, looks like a really challenging problem that you've made a lot of headway on, so kudos!
> 
> Regarding this PR, I'm slightly confused what the overall objective is as there is no linked Jira ticket in the description and the name "Updating sanitisation script for new table" doesn't seem to match the code I've just reviewed. For example, what exactly is the new table referred to in the PR name? In this PR I can see a variety of changes, including:
> 
> 1. Adding anonymisation to additional (but not all) columns in the `funding_question` and `project_progress` tables
> 2. Deliberately not sanitising certain records in the `project_progress` table
> 3. An update to the way we handle anonymising the `risk_register` data blob, to generate new lorem ipsum for each field and avoid deleting unnamed fields
> 4. A new CLI command to abstract away direct usage of the `restore-sanitised-database` script, and a new argument for that script to bypass user input
> 
> All are potentially useful (although need clarification on no. 2), _but_ I'm not sure they should all be hanging out together in the same PR?
> 
> Interested to get your thoughts on the comments I've left - merge is blocked for now pending your clarification on certain points. Good work anyway :)

Hi @wjrm500. thanks for thoroughly reviewing the PR.

This PR is actually related to this [ticket](https://mhclgdigital.atlassian.net/browse/FPASF-223) although it's the updated version with some fixes.

This change i'm making because, when @gidsg  tested the restore script from the prod to dev environment, he noticed that some business-sensitive information was still in the `project_progress` table and parts of the `funding_question` table.

To fix this, I decided to anonymise the commentary field, as it had a lot of business-related details. I also sanitised the `risk_rag` and `spend_rag` fields by replacing them with random numbers. I didn’t sanitise every field, though, as that would slow down the script. The fields I didn’t sanitise don’t reveal much on their own without the ones I did anonymise, like commentary.

scripts/restore-sanitised-database.sh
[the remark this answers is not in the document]
israr-ulhaq: The reason this parameter is set in this way is due to the following:

- If someone wants to restore a sanitised database locally, they can use the Flask command db-data seed-sanitised-db. This will only restore the database to the local environment because the `seed_sanitised_db()`  function automatically passes the 'y' argument to the script, which is why the default is set to handle this case.
- 
- If a user wants to restore the sanitised database to an AWS (dev/test) environment, they would typically run the restore script from the terminal using ./scripts/restore-sanitised-database.sh. This way script will not have any arguments and will prompt the user to confirm if they want to restore the database locally. If they respond 'n' (No), the script will proceed to restore the database to the AWS environment based on the profile selected via AWS Vault. Before we were doing only this but flask command has been added so it can be restored locally from cli.


Yes I can change the name **RESTORE_DB** to more clear name like you said **RESTORE_TO_LOCAL_DB**.

scripts/sanitise-db.sql
israr-ulhaq: Same like above :)

scripts/sanitise-db.sql
israr-ulhaq: I mentioned in the other comment, I didn’t sanitise every field, which wasn't revealing a lot but these also can be replace with lorem_ipsum if want to.

scripts/sanitise-db.sql
israr-ulhaq: Yeah that's the pain point and we might think how we will manage the newly coming PII but for now will need to add here if something new comes with the PII. 

In the previous we are using `jsonb_build_object` and it was not sanitising working but yes this looks fine. Thank You.

scripts/sanitise-db.sql
israr-ulhaq: Our code checks the constraint making sure that the start_date must always be before the end_date. When sanitising project_progress,I encounter some errors for entries that violate this constraint, so we only update records where this condition is met.

scripts/restore-sanitised-database.sh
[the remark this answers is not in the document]
wjrm500: Thanks @israr-ulhaq, understood on the reasons but a named argument would be better even if that is only used in the context of the Flask CLI command, just for clarity

Sounds good re the variable name change 👍

scripts/sanitise-db.sql
wjrm500: Cool, understood - let's just sanitise all of the fields in these tables for consistency - sanitisation by default, unless there is a special reason not to

scripts/sanitise-db.sql
wjrm500: Same as above for me too 😄 although quick note - I would have thought the "response" field here could potentially contain some stuff that we'd want to sanitise even if we weren't being super-cautious

scripts/sanitise-db.sql
wjrm500: Ah cool I've looked into this and the reason is that even though we've added a database constraint for start date and end date to the `project_progress` table, existing rows were not validated against that constraint at migration time. When you run the `UPDATE` statement in this sanitisation script against a violating row, the constraint is re-validated and the query fails

From checking the prod database just now, only one row fails that constraint. We ran a re-ingest last week to cover the other cases of this case failing, but this row remains outstanding. @samuelhwilliams can we update this final violating row in prod - whether manually or through reingest - to avoid Israr having to add this condition into the sanitisation script?

samuelhwilliams: have asked about that last bit of data again...

PR #783, opened 2024-10-14
Using freezegun to freeze time in integration test to fix occasional minute-rounding mismatch error

wjrm500:
### Description

Currently we get unpredictable failures in the `test_multiple_rounds_multiple_funds_end_to_end` integration test, which occur when the ingest occurs in the minute before (or multiple minutes before) the expected dataframe for the extract is created. This is because for each ingest, we use `datetime.now` to set the submission date which appears in the extract, and then only after all ingests and the download have taken place do we then use `datetime.now` again to determine the expected values in the extract. The test passing is currently reliant on the first and last calls to `datetime.now` happening before the minute mark ticks across.

You can observe the failure happening 100% of the time if you add a sleep call of 60 seconds just after all ingest calls in the test function.

This change introduces a new test dependency freezegun to freeze time in the test, meaning that we can exert global control over `datetime.now` without worrying about patching. This breaks the type check in Towns Fund schema validation, which currently relies on the line `exp_type == datetime`, which doesn't work when `datetime` has been replaced by freezegun's `FakeDatetime` - so we need to replace the equality operator with `issubclass`, which is more robust anyway.

### DIFF
--- a/data_store/validation/towns_fund/schema_validation/validate.py
+++ b/data_store/validation/towns_fund/schema_validation/validate.py
@@ -157,7 +157,7 @@ def validate_types(
 
             # do not raise an exception for pandas Timestamp, datetime or number values
             if (isinstance(got_value, numbers.Number) and exp_type in [int, float]) or (
-                isinstance(got_value, (datetime, pd.Timestamp)) and exp_type == datetime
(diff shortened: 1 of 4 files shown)


sc15zs (approved):
Installed freezegun and test passes.

PR #784, opened 2024-10-15
Drop ingest_date column

sc15zs:
Ticket:

Description:
We have already made the Submission model's ingest_date column nullable and stopped writing to it. 
This PR is for dropping the column. 

### DIFF
--- a/data_store/db/entities.py
+++ b/data_store/db/entities.py
@@ -606,7 +606,6 @@ class Submission(BaseModel):
     reporting_round_id: Mapped[GUID] = sqla.orm.mapped_column(sqla.ForeignKey("reporting_round.id"), nullable=False)
 
     submission_date = sqla.Column(sqla.DateTime(), nullable=False)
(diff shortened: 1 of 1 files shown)


samuelhwilliams (changes requested):
Just for good practice, can we please split this out into two PRs?

Migrations that remove columns are not backwards compatible, which means that between the migration running and the code deployment finishing, there is a risk of errors.

You can confirm this locally by:

* Upgrade the DB to the migration in this PR
* Checkout main
* Run the e2e tests, which will fail (check docker logs for the stacktrace :) )

The two PRs needed are:

1) The python code changes - `entities.py` only
2) The alembic migration - `.current-alembic-head` and `046_drop_ingest_date.py`

samuelhwilliams:
If you run the e2e tests and investigate the errors thrown, you may be able to work out why that's not _exactly_ true (sadly) :)

samuelhwilliams (approved):

PR #798, opened 2024-10-24
Upgrade pre-commit

samuelhwilliams:
### Change description
Bumps our local pre-commit version to be >=4.0. This matches what runs in pre-commit.ci, so is best to keep roughly in sync.

In particular this bump has been prompted by CI failures on the assessment-store, where one of the installed hooks is incompatible with v4 of pre-commit because the `python_venv` "language" has been removed. We don't see this issue locally on that repo because we're still running an older version of pre-commmit.

### DIFF
--- a/requirements-dev.in
+++ b/requirements-dev.in
@@ -39,5 +39,5 @@ types-WTForms
 pip-tools
 
 
-pre-commit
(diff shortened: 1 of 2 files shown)


MarcUsher (approved):
🫡

PR #804, opened 2024-10-29
Small script fixes

samuelhwilliams:
* Stop the script if any command fails
* Make sure user is inside the `scripts` directory, as this is expected elsewhere.
* Fix a broken `if` statement (missing an exit and an end).
* Fix the `read` command to not exit with status 1.
* Move postgresql_anonymizer cloning before the DB cluster setup step so that if it fails we don't waste a load of time.


israr-ulhaq (approved):
LGTM! 👍

PR #805, opened 2024-10-30
Update vscode debugging port for combined docker runner

samuelhwilliams:
### Change description
For the combined docker runner, the debugging port for data-store has been updated: https://github.com/communitiesuk/funding-service-design-docker-runner/pull/71

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -6,11 +6,11 @@
     "configurations": [
         {
             "name": "Docker runner Data store",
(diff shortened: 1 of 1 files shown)


MarcUsher (approved):
Yar! One other small change we could make here is my VS Code suggests changing the `type` here from `python` to `debugpy` as the `python` thing will be deprecated soon.

samuelhwilliams:
It is done 🙇

PR #811, opened 2024-11-07
Create OrganisationType model

sc15zs:
Ticket:

Description:
As part of the work to be done in the ticket, in this PR a new OrganisationType model is created, 
as a Foreign key to the Organisation model.


gidsg:
How will be populate this table? Is the plan to do that in a separate migration?

MarcUsher (commented):
Seeing as this new table will essentially just have one field (`type_name`) and then be mapped to organisations, does it need to be a separate table at all? Would it be simpler to add a new (enum?) column on `Organisation` instead?

sc15zs:
> Seeing as this new table will essentially just have one field (`type_name`) and then be mapped to organisations, I'm just wondering if it needs to be a separate table at all? Would it be simpler to add a new (enum?) column on `Organisation` instead? Asking in case there's some further implementation context I'm missing.

sc15zs:
> Seeing as this new table will essentially just have one field (`type_name`) and then be mapped to organisations, I'm just wondering if it needs to be a separate table at all? Would it be simpler to add a new (enum?) column on `Organisation` instead? Asking in case there's some further implementation context I'm missing.

Gideon, Jordan and I thought that having it as a Foreign key would be a better solution rather than a hardcoded enumeration. If we ever want to change the type_name, we change it once and it’s then consistent across all related records. 
If we want to add another property any time in the future, it’s also easy and couldn’t be done if it was an enumeration. 
I _think_ a foreign key lookup is more efficient for queries than selecting by enums.

sc15zs:
@gidsg 
> How will be populate this table? Is the plan to do that in a separate migration?

To be honest, I didn’t know how you usually did it and thought to have a separate python script which could be run in dev, test and prod. 
Or do you prefer doing it in this or a separate migration on upgrading, like: op.execute("INSERT INTO organisation_type_dim (type_name) VALUES ('Local Authority’)”)? 

Now the latter seems to be much quicker...

sc15zs:
@gidsg 
I've added the instance creation to the migration file.

db/migrations/versions/050_create_organisationtype.py
```diff
@@ -0,0 +1,57 @@
...
+            batch_op.f("fk_organisation_dim_organisation_type_id_organisation_type_dim"),
+            "organisation_type_dim",
+            ["organisation_type_id"],
+            ["id"],
+            ondelete="SET NULL",
+        )
+
+    # Insert "Local Authority" into organisation_type_dim
+    new_id = str(uuid.uuid4())
+    op.execute(f"INSERT INTO organisation_type_dim (id, type_name) VALUES ('{new_id}', 'Local Authority')")
```
samuelhwilliams, line 44: There has been some discussion before around doing data migrations as part of alembic migrations. Alembic is for structural changes and not really designed for inserting arbitrary "reference" data (which is what we use it for), which is why we have things like the db seeding scripts/functions.

This should probably live there.

That said I agree with Marc that a whole separate table for this feels like overkill compared to an enum...

jd-mhclg:
Really keen to move away from enum's so we can edit the data/update data and look ups without having to do migrations. I think the table is valid even on such a small scale as it is likely to grow once more organisation types identified or if it's required it becomes more granular. 

I do agree on the Alembic migrations bit @samuelhwilliams. Be good to run this was also just some initial work to do before looking further into the delta integration infrastructure/architecture. Which we may want to run past the platform team after our discussion tomorrow @gidsg

jd-mhclg:
Just had a huddle with gids and agreed, probably don't really need a table for now anyway since the only type is local authority. 

The enum vs table question though still interest me, since i think it may depend on how we see this data being updated in the future and by who :)

samuelhwilliams:
Thanks @jd-mhclg - happy that we're going for the simpler thing first.

I do agree that for any/most reference data, using tables and a more mature "data entry" interface (eg flask-admin) feels good rather than needing code changes. But maybe not on this specific case yet.

scripts/setup.py
```diff
@@ -0,0 +1,21 @@
...
+
+    try:
+        session.query(Organisation).update({"organisation_type": local_authority})
+        session.commit()
+        print("Organisation types updated successfully")
+    except Exception as e:
+        session.rollback()
+        print(f"Something went wrong: {e}")
+    finally:
+        session.close()
```
samuelhwilliams, line 21: Hey hey 👋 Now that we're just adding a new column, I think we include a bit of nuance on the "no data migrations in alembic" rule 👀 

In this case I think doing an UPDATE statement to migrate _existing_ records is OK - it's just _new_ data that we shouldn't be adding.

So I think we could make the column non-nullable immediately, and then the alembic migration can:

1) Add the column as nullable=True
2) Update the rows to set the organisation_type to LA
3) Update the column to set it as nullable=False

---

This should work in a sane way, and not break anything with eg our test setup because when the migrations run there, the DB will be empty.

samuelhwilliams (approved):

PR #813, opened 2024-11-12
Alter organisation_type

sc15zs:
Ticket:

Description:
This PR changes the organisation_type on  the Organisation model to nun-nullable, and adds a default value.


db/migrations/versions/051_alter_organisation_type.py
```diff
@@ -0,0 +1,42 @@
...
+revision = "051_alter_organisation_type"
+down_revision = "050_add_organisation_type"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    org_type_enum = sa.Enum("LOCAL_AUTHORITY", name="organisationtypeenum")
+    org_type_enum.create(op.get_bind(), checkfirst=True)
+
```
samuelhwilliams, line 22: We can drop this, as the enum was added in migration 50

samuelhwilliams: (maybe just the 'create' line, on second thoughts)

PR #815, opened 2024-11-12
Fix non aschii filename error

albertkol:
### Change description
Fixes the recent Sentry error caused by uploading files with non-aschii names. 

The change make it so we no longer use the user submitted file name on AWS upload.
The user confirmation email will still show the user submitted file name. 

### How to test
Use a file with non-aschii chars and try to submit it.

### DIFF
--- a/data_store/controllers/ingest.py
+++ b/data_store/controllers/ingest.py
@@ -503,7 +503,7 @@ def save_submission_file_s3(excel_file: FileStorage, submission_id: str):
         object_name=f"{fund_type}/{str(uuid)}",
         metadata={
             "submission_id": submission_id,
-            "filename": excel_file.filename,
+            "filename": f"{fund_type}-{str(uuid)}.xlsx",
             "programme_name": programme_name,
         },
     )
--- a/tests/integration_tests/test_aws.py
+++ b/tests/integration_tests/test_aws.py
@@ -92,7 +92,7 @@ def test_save_submission_file_s3(seeded_test_client, test_buckets):
 
     assert response["Body"].read() == filebytes
     assert metadata["submission_id"] == submission_id
-    assert metadata["filename"] == filename
+    assert metadata["filename"] == f"HS-{uuid}.xlsx"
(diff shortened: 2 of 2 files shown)


data_store/controllers/ingest.py
```diff
@@ -503,7 +503,7 @@ def save_submission_file_s3(excel_file: FileStorage, submission_id: str):
         object_name=f"{fund_type}/{str(uuid)}",
         metadata={
             "submission_id": submission_id,
-            "filename": excel_file.filename,
+            "filename": f"{fund_type}-{str(uuid)}.xlsx",
```
samuelhwilliams, line 506: I think there are a few other options that might be better for fund teams:

* Sanitise the string by removing anything non-ascii
* Insert more useful data - submission datetime, organisation name, submission ID (rather than UUID), etc?

albertkol: Fund teams are not affected by this change. This is only what AWS stores as a filename.

Fund team get the pretty human readable filename that causes other Sentry errors currently. :laughing: 
Example: `File name: 2024-11-12-td-worcester-city-council-apr2024-sep2024.xlsx`

samuelhwilliams: Hm you're right. The only time the filename metadata is used is in admin actions. Do we even need to store the original filename? 🤔

albertkol: We can have a quick chat tomorrow. 

I considered 1st to strip "illegal" chars. But then if this file name affects nothing else, then this is a quick, simple fix.

The user submitted file name still gets echoed in two place, but it's never store, only used during the request:
- HTML template green success box 
- User confirmation email 

This changes only the filename AWS stores.

The file name the Fund Team confirmation email gets is the pretty human readable version. (Until we fix the problem with old fund team confirmation email file name error.)

samuelhwilliams (approved):

data_store/controllers/ingest.py
gidsg: I think as it's also for audit purposes can we keep the original filename but make it us-ascii safe?

PR #818, opened 2024-11-14
Make TF r6 ingest before and beyond columns

albertkol:
### Change description
Make TF round 6 ingest:
- Before 20/21 from 4a - Funding Profiles sheet, SECTION C: Project Funding Profiles
- Beyond 25/26 from 4a - Funding Profiles sheet, SECTION C: Project Funding Profiles
From the _**Towns Fund Funding Source**_ tables.

These were ignored before. As former rounds were not uisng those columns. From Round 6 and onwards, they will be required.

### How to test
Are tests passing?


gidsg (approved):
👍

PR #841, opened 2024-11-17
Extend from shared renovate config

samuelhwilliams:
### Change description
Pulls renovate config from a shared central repository ([communitiesuk/renovate-config](https://github.com/communitiesuk/renovate-config/blob/main/default.json)). The main difference this adds, for now, is that it will group updates for ruff in Python and pre-commit together, because otherwise it's annoying.


MarcUsher (approved):
Let's goooooo

PR #849, opened 2024-11-18
Update Spreadsheet Retrieval using Metadata

israr-ulhaq:
https://mhclgdigital.atlassian.net/browse/FLS-788

When a Local Authority (LA) submits their return, we send the fund team a link to download the file. If the LA re-submits the file, the old submission is removed, and the fund team gets an error because the submission_id has changed.

To fix this, we now store the filename in the `download_filename` metadata when the file is uploaded. This way, we can use the metadata to retrieve the file instead of relying on the database. For older files, if the new metadata is missing, we use a custom fallback name to ensure the file can still be accessed.

### DIFF
--- a/data_store/controllers/retrieve_submission_file.py
+++ b/data_store/controllers/retrieve_submission_file.py
@@ -71,10 +71,13 @@ def get_custom_file_name(submission_id: str) -> str:
         .one_or_none()
     )
 
-    date = submission_info.submission_date.strftime("%Y-%m-%d")
-    start_date = submission_info.observation_period_start.strftime("%b%Y")
-    end_date = submission_info.observation_period_end.strftime("%b%Y")
+    if submission_info:
+        date = submission_info.submission_date.strftime("%Y-%m-%d")
+        start_date = submission_info.observation_period_start.strftime("%b%Y")
+        end_date = submission_info.observation_period_end.strftime("%b%Y")
 
-    file_name = f"{date}-{submission_info.fund_code}-{submission_info.organisation_name}-{start_date}-{end_date}"
+        file_name = f"{date}-{submission_info.fund_code}-{submission_info.organisation_name}-{start_date}-{end_date}"
+    else:
+        file_name = f"data-return-{submission_id}"
 
     return file_name.replace(" ", "-").lower()
--- a/data_store/controllers/retrieve_submission_file.py
+++ b/data_store/controllers/retrieve_submission_file.py
@@ -1,3 +1,5 @@
+from typing import Optional
+
 from config import Config
 from data_store.aws import create_presigned_url, get_file_header
 from data_store.db.entities import Fund, Organisation, Programme, ProgrammeJunction, ReportingRound, Submission
@@ -52,7 +54,12 @@ def retrieve_submission_file(submission_id) -> str:
     return presigned_url
 
 
-def get_custom_file_name(submission_id: str) -> str:
+def get_custom_file_name(
+    submission_id: str,
+    fallback_date: Optional[str] = None,
+    fallback_fund_code: Optional[str] = None,
+    fallback_org_name: Optional[str] = None,
+) -> str:
     submission_info = (
         Programme.query.join(ProgrammeJunction)
         .join(Submission)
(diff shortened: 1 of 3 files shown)


find/main/routes.py
```diff
@@ -195,12 +194,15 @@ def retrieve_spreadsheet(fund_code: str, submission_id: str):
         return abort(404)
 
     form = RetrieveForm()
-
+    file_metadata = file_header["metadata"]
+    file_name = file_metadata.get("download_filename")
+    if not file_name:
+        file_name = f"{fund_code}-{submission_id}"
```
samuelhwilliams, line 200: I think in the fallback case we would still need to include the file extension, eg `.xlsx`.

samuelhwilliams (changes requested):
Just a recap on my side for the problem:

When a submission happens, we assign it a UUID (id) and a human-readable code (submission_id).

If a re-submission happens, it gets a new UUID (id) but retains the same human-readable code (submisison_id).

The record of the old submission (UUID) is dropped out of the DB.

For each submission, we email the fund team with a link to the specific UUID file uploaded.

After a re-submission, the original UUID(id) is no longer in the DB so the lookup fails.

Ideally we would retain all versions, but there's is "work" that needs to happen to facilitate that first. This PR is an attempt to fix the problem in the interim.

With your change, when a user does a submission, we now start freezing the download filename in S3 and using this when the download happens, as opposed to trying to generate the filename on demand (because this will fail if the row in the DB associated with the UUID is no longer present).

All historical submissions will not have the `download_filename` metadata, which means they fall through to the hard-coded download filename. This feels sub-optimal, because we're providing a "worse" filename. In most cases, it's possible that the "nicer" filename could be generated - we would only know by trying to do that first.

I think we should update the `get_custom_file_name` function to try to fetch the data first and generate a nice filename, and only provide a fallback if the lookup fails.

This will lead to nice download filenames in as many cases as possible.

israr-ulhaq:
> Just a recap on my side for the problem:
> 
> When a submission happens, we assign it a UUID (id) and a human-readable code (submission_id).
> 
> If a re-submission happens, it gets a new UUID (id) but retains the same human-readable code (submisison_id).
> 
> The record of the old submission (UUID) is dropped out of the DB.
> 
> For each submission, we email the fund team with a link to the specific UUID file uploaded.
> 
> After a re-submission, the original UUID(id) is no longer in the DB so the lookup fails.
> 
> Ideally we would retain all versions, but there's is "work" that needs to happen to facilitate that first. This PR is an attempt to fix the problem in the interim.
> 
> With your change, when a user does a submission, we now start freezing the download filename in S3 and using this when the download happens, as opposed to trying to generate the filename on demand (because this will fail if the row in the DB associated with the UUID is no longer present).
> 
> All historical submissions will not have the `download_filename` metadata, which means they fall through to the hard-coded download filename. This feels sub-optimal, because we're providing a "worse" filename. In most cases, it's possible that the "nicer" filename could be generated - we would only know by trying to do that first.
> 
> I think we should update the `get_custom_file_name` function to try to fetch the data first and generate a nice filename, and only provide a fallback if the lookup fails.
> 
> This will lead to nice download filenames in as many cases as possible.

Thanks Sam for recapping the issue. You're absolutely right, I've updated the fallback mechanism.

data_store/controllers/retrieve_submission_file.py
```diff
@@ -71,10 +71,13 @@ def get_custom_file_name(submission_id: str) -> str:
...
-    end_date = submission_info.observation_period_end.strftime("%b%Y")
+    if submission_info:
+        date = submission_info.submission_date.strftime("%Y-%m-%d")
+        start_date = submission_info.observation_period_start.strftime("%b%Y")
+        end_date = submission_info.observation_period_end.strftime("%b%Y")
 
-    file_name = f"{date}-{submission_info.fund_code}-{submission_info.organisation_name}-{start_date}-{end_date}"
+        file_name = f"{date}-{submission_info.fund_code}-{submission_info.organisation_name}-{start_date}-{end_date}"
+    else:
+        file_name = f"data-return-{submission_id}"
```
samuelhwilliams, line 81: I think this fallback filename could be made nicer. The `submission_id` variable here is just a UUID, but the S3 metadata for the file should contain the human-readable code.

Unfortunately our naming here is not great - `submission.id` is a UUID, `submission.submission_id` is the friendly version. The latter should be available on the S3 object, and it also has the programme name.

So maybe in the fallback case we can build the filename from the S3 metadata. We can still get: the date of the submission (should be something like the object's CreatedAt value), the friendly submission ID and the LA name (metadata `programme_name`).

I realise this is a bit more work, but feels like the nice thing to do and the best information we can provide.

Thoughts?

israr-ulhaq: Updated the code to get user friendly filename.

samuelhwilliams (approved):

PR #852, opened 2024-11-19
Making ascii safe filenames for submission

israr-ulhaq:
When a user uploaded a file with a non-ASCII safe filename, it caused an error in our system (logged on Sentry).

To address this, I made the following updates:

When a file is uploaded, we first try to make the original filename ASCII-safe.
If for any reason the filename cannot be sanitised e.g., it contains characters that can’t be converted to ASCII, the system automatically falls back to a user-friendly, filename.
This ensures that file uploads will always work smoothly, even when the original filename contains problematic characters. Users will either get an ASCII-safe version of the original filename or a simple, readable fallback name.

### How to test

### DIFF
--- a/data_store/controllers/ingest.py
+++ b/data_store/controllers/ingest.py
@@ -1,6 +1,5 @@
 """Provides a controller for spreadsheet ingestion."""
 
-import os
 import typing
 import uuid
 from datetime import datetime
@@ -526,8 +525,7 @@ def make_ascii_safe_filename(filename: str) -> str:
     """
     Converts a filename to an ASCII-safe format.
     """
-    base, ext = os.path.splitext(filename)
+    ascii_safe_filename = filename.encode("ascii", "replace").decode("ascii")
+    ascii_safe_filename = ascii_safe_filename.replace("?", "--")
 
-    base = "".join((c if ord(c) < 128 else "-") for c in base)
-
-    return f"{base}{ext}"
+    return ascii_safe_filename
--- a/data_store/controllers/ingest.py
+++ b/data_store/controllers/ingest.py
@@ -1,5 +1,6 @@
 """Provides a controller for spreadsheet ingestion."""
 
+import os
 import typing
 import uuid
 from datetime import datetime
@@ -496,14 +497,15 @@ def save_submission_file_s3(excel_file: FileStorage, submission_id: str):
         .with_entities(Submission.id, Fund.fund_code, Programme.programme_name)
         .distinct()
     ).one()
-
     upload_file(
         file=excel_file,
         bucket=Config.AWS_S3_BUCKET_SUCCESSFUL_FILES,
         object_name=f"{fund_type}/{str(uuid)}",
         metadata={
             "submission_id": submission_id,
-            "filename": f"{fund_type}-{str(uuid)}.xlsx",
+            "filename": make_ascii_safe_filename(
+                excel_file.filename or f"{fund_type}-{programme_name}-{submission_id}.xlsx"
+            ),
             "programme_name": programme_name,
         },
     )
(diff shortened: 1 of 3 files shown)


data_store/controllers/ingest.py
```diff
@@ -518,3 +520,14 @@ def save_failed_submission(file: IO) -> uuid.UUID:
...
+
+def make_ascii_safe_filename(filename: str) -> str:
+    """
+    Converts a filename to an ASCII-safe format.
+    """
+    base, ext = os.path.splitext(filename)
+
+    base = "".join((c if ord(c) < 128 else "-") for c in base)
+
+    return f"{base}{ext}"
```
samuelhwilliams, line 533: Could we get a set of explicit unit tests for this that make it clear what goes in and what comes out? It feels like a nice self-contained thing to test.

tests/integration_tests/test_aws.py
```diff
@@ -92,10 +92,26 @@ def test_save_submission_file_s3(seeded_test_client, test_buckets):
...
     assert response["Body"].read() == filebytes
     assert metadata["submission_id"] == submission_id
-    assert metadata["filename"] == f"HS-{uuid}.xlsx"
+    assert metadata["filename"] == filename
     assert metadata["programme_name"] == "Leaky Cauldron regeneration"
 
 
+def test_save_submission_file_s3_ascii_safe(seeded_test_client, test_buckets):
+    filename = "example–file.xlsx"
+    ascii_safe_filename = "example-file.xlsx"
```
samuelhwilliams, line 101: With close inspection it's possible to see these are different, but it's not easy.

Maybe we can make this more obvious?

```suggestion
    filename = "example\u2013file.xlsx"
    ascii_safe_filename = "example-file.xlsx"
```

data_store/controllers/ingest.py
```diff
@@ -496,14 +497,15 @@ def save_submission_file_s3(excel_file: FileStorage, submission_id: str):
...
-
     upload_file(
         file=excel_file,
         bucket=Config.AWS_S3_BUCKET_SUCCESSFUL_FILES,
         object_name=f"{fund_type}/{str(uuid)}",
         metadata={
             "submission_id": submission_id,
-            "filename": f"{fund_type}-{str(uuid)}.xlsx",
+            "filename": make_ascii_safe_filename(
+                excel_file.filename or f"{fund_type}-{programme_name}-{submission_id}.xlsx"
```
gidsg, line 507: Will this not still throw an error as the excel filename will always have some value and there is no try-catch? (sorry if I missed something)

israr-ulhaq: It should always be defined during the file upload process. I’ve included a fallback in case the filename is empty, None, or if something goes wrong during ascii validation. This is also to  a warning from mypy.

data_store/controllers/ingest.py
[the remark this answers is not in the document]
israr-ulhaq: Yes, we can use and it works fine for ascii chars, I'll update the code. It only does not cover special character which i think we don't mind to have as S3 accept them?

israr-ulhaq: Sure!

tests/integration_tests/test_aws.py
israr-ulhaq: Done!

tests/submit_tests/test_utils.py
```diff
@@ -17,3 +18,37 @@ def test_calculate_past_date(submit_test_client):
...
+    filename = "file_2024é.xlsx"
+    expected_output = "file_2024--.xlsx"
+
+    assert make_ascii_safe_filename(filename) == expected_output
+
+
+def test_ascii_safe_filename_with_only_invalid_ascii():
+    filename = "file_2024\u2019.xlsx"
+    expected_output = "file_2024--.xlsx"
+    assert make_ascii_safe_filename(filename) == expected_output
```
samuelhwilliams, line 54: Where the logic of the test is the same, just the values are different, it can be a lot cleaner and simpler to use test parameterisation instead:

```suggestion
@pytest.mark.parametrize(
    'filename, expected_output',
    (
        ("téstfilé_2024.xlsx", "t--stfil--_2024.xlsx"),
        ("testfile_2024é.xlsx", "testfile_2024_.xlsx"),
        ("file_2024é.xlsx", "file_2024--.xlsx"),
        ("file_2024\u2019.xlsx", "file_2024--.xlsx"),
    )
)
def test_ascii_safe_filename(filename, expected_output):
    assert make_ascii_safe_filename(filename) == expected_output
```

israr-ulhaq: Done!

samuelhwilliams (approved):

PR #855, opened 2024-11-22
Editing external_reference_code

sc15zs:
Ticket:

Description:
This PR allows editing of an organisation's external_reference_code in the admin view.

### DIFF
--- a/admin/entities.py
+++ b/admin/entities.py
@@ -1,7 +1,9 @@
 from abc import abstractmethod
 from datetime import datetime
 
-from flask import current_app, g
+from flask import current_app, g, request
+from flask_admin.babel import gettext
+from flask_admin.base import expose
 from flask_admin.contrib import sqla
 from flask_admin.contrib.sqla.typefmt import DEFAULT_FORMATTERS
 from flask_admin.form import DatePickerWidget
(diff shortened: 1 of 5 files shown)


admin/entities.py
```diff
@@ -103,6 +104,49 @@ class SubmissionAdminView(BaseAdminView):
...
+
+        return super(OrganisationAdminView, self).edit_view()
+
+    def on_model_change(self, form, model, is_created):
+        if not is_created:
+            for field in form.data:
+                if field not in self.fields_allowed_to_edit and hasattr(model, field):
+                    setattr(model, field, getattr(self.instance, field))
+
+        return super(OrganisationAdminView, self).on_model_change(form, model, is_created)
```
samuelhwilliams, line 149: This is important behaviour to get correct - could we add some tests around this?

sc15zs: Added a test.

tests/admin_tests/test_actions.py
```diff
@@ -103,3 +103,28 @@ class TestRetrieveSubmissionAdminView:
...
     ...
+
+
+def test_organisationadmin_edit(setup_organisationadmin, admin_test_client):
+    view = setup_organisationadmin
+    OrganisationTypeEnum.NEW_VALUE = "New Value"
+    form_data = {
+        "organisation_name": "New Name",
+        "external_reference_code": "New Code",
+        "organisation_type": OrganisationTypeEnum.NEW_VALUE,
```
samuelhwilliams, line 114:
```suggestion
        "organisation_type": OrganisationTypeEnum.NEW_VALUE.value,
```

I think this needs a `.value` at the end so that it's the actual string, rather than an Enum instance. Only the text representation would come to us over the wire, so we should send that rather than a python object.

tests/conftest.py
```diff
@@ -876,3 +877,20 @@ def towns_fund_round_6_file_failure() -> Generator[BinaryIO, None, None]:
         "rb",
     ) as file:
         yield file
+
+
+@pytest.fixture
+def setup_organisationadmin(test_session):
+    view = OrganisationAdminView(test_session)
+    view.fields_allowed_to_edit = ["external_reference_code"]
```
samuelhwilliams, line 885: Why do we need to instantiate this here? This feels quite unusual 🤔

sc15zs: Yes, it was totally unnecessary.

tests/admin_tests/test_actions.py
sc15zs: Changed it such that we pass the value to the form_data now, but had to create a new Enum class for this to work. With the original version, adding the value attribute it just returned an error.

admin/entities.py
```diff
@@ -103,6 +104,49 @@ class SubmissionAdminView(BaseAdminView):
 
 class OrganisationAdminView(BaseAdminView):
     _model = Organisation
+    instance = None
```
samuelhwilliams, line 107: I am not sure that storing the instance on a class variable is very safe in a web app. When we start handling multiple requests in parallel, the instances could get mixed up between requests.

The safer way of doing this in Flask is to use the `g` object, so maybe we can store this on `g.__organisation_admin_instance` or something - if we really need to have some reference that we can't track and pass between functions easily?

sc15zs: Removed the instance from the class and added some tests.


A note regarding manually changing the form data in the html, as you've told me to do in our huddle last time:

1. So, any extra fields added to the POST request's form, which are in the `form_excluded_columns` list will be removed from the form by Flask-Admin. 
I tried it and the `test_organisationadmin_edit_disallowed` function tests this: it updates the allowed column, but removes all other columns from the form and won't update those. In this case we wouldn't even need the `on_model_change` function (but being there doesn't hurt anything).

2. To make absolutely sure that nothing can be changed with a forged post request, the `on_model_change` function aborts editing of an instance if there is any field in the form which is not in the `fields_allowed_to_edit` list. For this to be triggered the given columns should also be removed from the `form_excluded_columns` list (I don't know how it could be achieved though). This is tested in the `test_organisationadmin_on_model_change` function.

I hope it makes sense. =)

gidsg:
@sc15zs I have run this locally and it looks good. Only thing is is there a way to remove the 'Save and Continue' button, as doesn't appear to do anything useful in this case?

sc15zs:
@gidsg 
> @sc15zs I have run this locally and it looks good. Only question I have is, is there a way to remove the '_Save and Continue_' button, as doesn't appear to do anything useful in this case?

Thanks, Gideon! Removed the middle button, so now there's only the 'Save' and 'Cancel' buttons on the form.

gidsg (approved):

PR #858, opened 2024-11-26
Force str type

albertkol:
### Change description
It turns out in some situations the cell is not a string. So `.lower()` throws a 500.

This is blocking PF submissions.


samuelhwilliams (approved):
no tests, no examples, good to go? 👀

albertkol:
I tested it with the file that raised the error, and managed to at least reach the validation page without raising a 500. 

Ideally it was tested and the problematic cell was found, but on the week of the deadline, this will have to do for now.

PR #859, opened 2024-11-27
Ignore nan problems

albertkol:
Ignore nan problems.

Make report not have problems with nan values.


albertkol:
I don't think this will be needed.

PR #861, opened 2024-11-27
Remove round 6 filter by period

albertkol:
Remove the Round 6 finance period limit. 

### How to test
Are tests passing? I added a HS R6 finance sheet test.


gidsg (approved):
LGTM 👍

PR #868, opened 2024-12-09
Pull in latest utils for any-host healthcheck

samuelhwilliams:
We used to hack around utils not supporting healthchecks on flask apps that were in `host_matching` mode, but as of v5.3.0 utils now support both `host_matching` and non-`host_matching` mode.

So let's pull that in and drop our workaround.


MarcUsher (approved):
👍

PR #885, opened 2025-01-02
Don't call API if there are no users

samuelhwilliams:
https://mhclgdigital.atlassian.net/browse/FSPT-207

### Change description
The account-store endpoint 400s/500s if there are no account IDs to resolve, which breaks this script.

If there are no account IDs, let's shortcircuit that bit.


albertkol (approved):
Where tests, m8?

PR #886, opened 2025-01-02
Fix multiple account IDs for bulk-account endpoint

samuelhwilliams:
We should be sending these as multiple query params, not as a comma-separated values in a single query param.


albertkol (approved):
where test?

PR #891, opened 2025-01-15
Add failed deployment notifications

samuelhwilliams:
### Change description
https://mhclgdigital.atlassian.net/browse/FSPT-220

Adds a slack notification if this pipeline fails.

Should not trigger if the pipeline deploy to prod is rejected by a developer, or the pipeline run is cancelled.

image


albertkol (approved):
Where test?

PR #930, opened 2025-02-04
Fls 928 find download bug

israr-ulhaq:
https://mhclgdigital.atlassian.net/browse/FLS-928

It should fix the [sentry]([https://funding-service-design-team-dl.sentry.io/issues/6223252656/?environment=prod&query=is%3Aunresolved issue.priority%3A[high%2C medium]&referrer=issue-stream&statsPeriod=14d&stream_index=0](https://funding-service-design-team-dl.sentry.io/issues/6223252656/?environment=prod&query=is%3Aunresolved%20issue.priority%3A%5Bhigh%2C%20medium%5D&referrer=issue-stream&statsPeriod=14d&stream_index=0))  error.

### DIFF
--- a/find/main/routes.py
+++ b/find/main/routes.py
@@ -195,7 +195,7 @@ def retrieve_spreadsheet(fund_code: str, submission_id: str):
 
     form = RetrieveForm()
     file_metadata = file_header["metadata"]
-    file_name = quote(file_metadata.get("download_filename"))
+    file_name = quote(str(file_metadata.get("download_filename")))
     programme_name = file_metadata.get("programme_name")
     submission_date = file_header["last_modified"].strftime("%d %B %Y")
     if not file_name:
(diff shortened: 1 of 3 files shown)


samuelhwilliams:
From the sentry error it looks like `file_metadata.get("download_filename"))` is `None` here. Will this just be making the filename the string `None` (that feels wrong)?

gidsg:
Looks like the `quote` function is unsafe on a `None` so I think we could put that behind an if statement e.g.?

e.g.
```
file_name = file_metadata.get("download_filename")
if file_name:
    quote_file_name = quote(file_name)
```
etc...

albertkol (approved):
It looks fine. Not sure we can test this before `prod`. But post-award is quiet at the moment, so it should be safe-ish to go ahead with it.

PR #936, opened 2025-03-03
FLS-1038: Give permissions to Pre-award and Form-Runner-Adapter logs

albertkol:
https://mhclgdigital.atlassian.net/browse/FLS-1038

Allow data-store access to pre-award and form-adapter logs.


gidsg:
Is there any ticket number for this? If so can we add it to the PR title, or prefix with BAU?

albertkol:
Agreed @gidsg. 

We will start by exploring a bit. I have about a day and a half of freedom to work on this. Would love to do it properly. There are many questions:
- what logs to catch
- what reports to build form it
- how to share the report (potentially feeding everything directly to PowerBI)

PR #955, opened 2025-03-06
Fix type hints after mypy update

samuelhwilliams:
A recent bump to the version of mypy we use started to flag this type hint error:

find/scripts/extract_download_logs.py:168: error: Argument 1 to "cloudwatch_logs_to_rows" has incompatible type
"list[list[ResultFieldTypeDef]]"; expected "list[list[dict[Any, Any]]]" [arg-type]

Unfortunately we didn't have our type hints CI check down as 'required to pass', so it got auto merged. That's fixed now, and this will make our type checks pass again.


albertkol (approved):
Ayo!

PR #983, opened 2025-03-21
BAU: bump healthcheck wait periods

gidsg:
### Change description
We were overriding these values to be quite a lot lower than the Copilot defaults. See: https://aws.github.io/copilot-cli/docs/include/http-healthcheck/ (Presumably to fail faster? But struggling to find description of this in the git history).

I believe this may be causing the healthcheck failures and slower deployments we are seeing https://github.com/communitiesuk/funding-service-design-post-award-data-store/actions/runs/13989962599/job/39171743177

```
      - (service pre-award-prod-post-award-Service-J2Cso6oerD2j) (port 4001) i                                      
        s unhealthy in (target-group arn:aws:elasticloadbalancing:eu-west-2:23                                      
        3700139423:targetgroup/pre-aw-Targe-R1P2PBRWBAHN/8ec46384e4296c46) due                                      
         to (reason Request timed out).     
```
🤞This restores the values closer to the copilot default so hopefully should speed up deployments.

### How to test
Production deployment shoudn't have healthcheck failures and take longer

### DIFF
--- a/copilot/post-award/manifest.yml
+++ b/copilot/post-award/manifest.yml
@@ -14,10 +14,10 @@ http:
   # You can specify a custom health check path. The default is "/".
   healthcheck:
     path: '/healthcheck'
-    healthy_threshold: 2
+    healthy_threshold: 5
     interval: 6s
     timeout: 5s
-    grace_period: 10s
+    grace_period: 30s
 # Your service is reachable at "http://data-store.${COPILOT_SERVICE_DISCOVERY_ENDPOINT}:4001" but is not public.
 
 # Configuration for your containers and service.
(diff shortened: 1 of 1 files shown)


samuelhwilliams (commented):
I definitely remember us intentionally lowering the thresholds to get faster deploys, so I'd much rather see a far smaller tweak than this - it will make deploys take a long longer to confirm as successful.

I think I have two comments:
1) Can we try a grace_period of 15s.
2) Can we investigate a little why the app takes so long to startup. Doesn't feel like it should need 10 seconds to get going in an ideal world?

PR #985, opened 2025-04-02
Migrate from levellingup to communities domain

wjrm500:
As per [this PR](https://github.com/communitiesuk/funding-service-design-docker-runner/pull/137) in Docker Runner, we will now be running apps locally at `communities.gov.localhost` not `levellingup.gov.localhost`. We should update Post-Award Data Store accordingly.


samuelhwilliams (approved):
Approved but please consider comment above 🙇

PR #986, opened 2025-04-02
FLS-1132 - Towns Fund Round 7

wjrm500:
### Ticket

[Towns Fund Round 7](https://mhclgdigital.atlassian.net/browse/FLS-1132)

### Changes

- **Initial fixes to support upload if initial validation is disabled**
The only difference between this round's spreadsheet and the last (other than "Reporting Period" date range change) is that in each of the tables in the "4a - Funding Profiles" tab, th
e name of the penultimate column has been changed from "Beyond 25/26" to "Financial Year 2026/27 (£s)". We need to make sure we map this new column name to a start and end date in the "convert_financial_halves" utility function, while continuing to support "Beyond 25/26", because one of the columns in the outputs tables maps its name to this.

- **Add initial validation schema and update ingest dependencies**
Initial validation schema almost the same as Round 6, with dates updated. Other than the new initial validation schema, the ingest dependencies for Round 7 reuse existing pipeline components, reflecting the lack of changes.

- **Add integration test with anonymised spreadsheet**
This spreadsheet was emailed over by a member of the Towns Fund team on the morning of Weds 2nd April 2025, and manually anonymised by myself. Textual data was redacted, drop-down values were randomly selected, and numeric data was entered at random.

- **Prepare new round to open conditional on new env var ENABLE_TF_R7**
This won't actually open the round, it will just mean that on the day we want the round to open, we can go into AWS and switch ENABLE_TF_R7 from False to True, with no need for deployment, in line with previous round openings. This change will however CLOSE the previous round.

### Notes

- Only one new integration test - the success case - has been added here. Not adding a failure case was a conscious decision - it seems a bit redundant given the lack of changes between Rounds 6 and 7, and would just slow down running the test suite further.

- On merge, we should add a new secret `ENABLE_TF_R7` to each AWS environment, and remove the `ENABLE_TF_R6` secret, which is made redundant here. `ENABLE_TF_R7` can be set to True in pre-prod environments for testing, but should be False in production of course.

- We will also need to add new ReportingRound records to the database for both TD and FHSF, in all envs. I think we can do this with Flask Admin


israr-ulhaq (approved):
Look great to me!

PR #1025, opened 2025-05-14
Pathfinders R3 actual / forecast column name fix

wjrm500:
For Pathfinders R3, when Bolton Council tried to submit their monitoring return, we got an [error in Sentry](https://funding-service-design-team-dl.sentry.io/issues/6605579921/?alert_rule_id=14494323&alert_timestamp=1747238842450&alert_type=email&environment=prod&notification_uuid=9448c6b8-8161-4eed-84c6-365c98e9a335&project=4505390859747328&referrer=alert_email): the columns in the spreadsheet did not match the columns specified in the backend schema. Specifically, in the spreadsheet we use 'Actual' instead of 'Forecast' for the FY periods 24-25 Oct-Dec and 24-25 Jan-Mar, while we have 'Forecast' for these periods in the backend schema. The spreadsheet is correct here, so we should update the backend schema accordingly.


israr-ulhaq (approved):
Looks great to me!

PR #1030, opened 2025-05-21
Change Pathfinders validation to allow negative values

sc15zs:
Ticket:
https://mhclgdigital.atlassian.net.mcas.ms/browse/FLS-1300

Description:
Loosen Pathfinders return validation for Bolton Council to allow submission of negative values in the Finances tab, in G37
and G51 cells.
This change removes the negative value validation for the columns of these two cells. 

How to test:
Ping me to send the file over and submit that to the portal.


israr-ulhaq (approved):
QA'd and it worked as expected!

